PR #70, opened 2026-08-09
fix: trim old reasoning items to stop Responses-API cache explosion

ranxianglei:
## Root cause

OpenAI Responses-API `reasoning` items (the model's encrypted chain-of-thought) were treated as **opaque preamble** and re-sent verbatim **every turn**, accumulating without bound.

Observed in `bili (3).log` (reported in dog/billion-context-pi#15):

| turn | preserved preamble items |
|------|--------------------------|
| 1 | 1 (additional_tools) |
| 2 | 2 (+reasoning) |
| … | 9, 11 |
| last | **50** (reasoning×49) |

One large response added ~39 reasoning items in a single jump. Consequences:

- **Context explosion** — 49 encrypted CoT blobs resent every request.
- **Cache prefix broken** — `cached` stuck at `14080` (the static `additional_tools` block) while `input` ballooned 42k→78k; cache hit dropped to 18%. The volatile reasoning sits *before* the conversation, so nothing past `additional_tools` could form a stable cached prefix.
- **Relay dropouts** — `responses stream ended without completion` ×3 from the relay station, triggered by the oversized requests.

Code path:
- `src/responses.ts` — `OPAQUE_ITEM_TYPES` includes `"reasoning"`; `responsesToCore` pushed every reasoning item into `preamble`.
- `src/server.ts` `prepareResponses` — `inputItems = [...preamble]` re-prepended the whole preamble every turn.

## Fix

Keep only the **most recent model response's reasoning** = reasoning items positioned *after* the last external input (user message / `function_call_output` / `custom_tool_call_output`). Reasoning at or before that boundary belongs to old, completed responses → drop it.

- `additional_tools` and all other opaque items are still preserved verbatim (unchanged behavior).
- New env `ACP_REASONING_KEEP` = `recent` (default) | `all` (legacy) | `none` (drop all).
- Debug log: `dropped N old reasoning item(s) (ACP_REASONING_KEEP=recent)`.

## Verification

- `npm run typecheck` ✅
- `npm test` — 173/173 pass (4 new tests for the reasoning trim) ✅
- `npm run build` ✅

## Tests added (`tests/proxy-responses-review.test.ts`)

- default (`recent`) trims old reasoning, keeps most recent response's reasoning
- `ACP_REASONING_KEEP=all` preserves every reasoning item (legacy)
- `ACP_REASONING_KEEP=none` drops all reasoning items
- `recent` keeps reasoning when it is the only (most recent) response

## Note

`version` is **not** bumped (content commit per AGENTS.md §4). If this should ship, it needs to ride a subsequent `*_release-v*` branch.

Refs: dog/billion-context-pi#15

### DIFF
--- a/README.md
+++ b/README.md
@@ -308,7 +308,7 @@ with no file at all).
 | `ACP_AUTO_UPDATE` | `1` | Set `0` to disable background self-update |
 | `ACP_LOG_FILE` | *XDG state path* | Log file path (`off` disables the file, keeps stderr) |
 | `ACP_DUMP_SSE` | *(none)* | Directory to dump SSE for debugging |
-| `ACP_REASONING_KEEP` | `recent` | Responses-API only: how `reasoning` items from previous responses are carried. `recent` keeps only the most recent response's reasoning (drops old chain-of-thought that accumulates every turn and breaks the prompt-cache prefix); `all` re-sends every historical reasoning item (legacy); `none` drops all reasoning. |
+| `ACP_REASONING_KEEP` | `recent` | Responses-API only: how `reasoning` items from previous responses are carried. `recent` keeps only the most recent response's reasoning (drops old chain-of-thought that accumulates every turn — a major context-size reduction); `all` re-sends every historical reasoning item (legacy); `none` drops all reasoning. |
 | `BILI_PERSIST` | `1` | Set `0` to disable session persistence (in-memory only, lost on restart) |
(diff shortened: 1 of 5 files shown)


ranxianglei:
Closing to redo the fix properly: route reasoning through the compression pipeline (like Anthropic thinking) so it gets hidden automatically when its turn is compressed, instead of the positional band-aid. See issue dog/billion-context-pi#15 for the architectural discussion.

PR #86, opened 2026-08-10
feat: three-state update mode and fixed nudge growth interval

### LINKED ISSUE #606: update mode three-state (auto/notify/off) — external proposal
### Problem
External proposal (LX666-666, PR #86): make the updater three-state — auto / notify-only / off — plus a nudge growth-interval knob.

Current state on master: the updater is always-auto (3-min checks, env kill-switch exists); nudge cadence is kernel-owned since 0.0.52 (pressure-band min-benefit gate), so the nudge-interval half of the proposal is obsolete as a host feature. The PR conflicts across 8 files against current master.

### Suggested fix
If the three-state update mode is wanted: reimplement it alone against current `src/update.ts` (remember AGENTS.md §5 — updater changes need a no-op release first). Otherwise close the PR with thanks. Decision needed from maintainer.

LX666-666:
Closes #606 — tracking issue for this PR.

Update becomes a real 3-state policy (auto/check/manual) instead of a boolean autoUpdate plus two buttons:
- type UpdateMode = "auto" | "check" | "manual"; config update.mode
- update.ts refactored into checkLatestVersion() / installUpdate() / runScheduledUpdate() — checking no longer installs
- manual mode never touches the network; only explicit user actions do
- precedence: CLI > BILI_UPDATE_MODE > update.mode > legacy autoUpdate
- registry metadata + tarball downloads reuse the upstream-proxy ProxyAgent egress so system-proxy users don't hit direct-fetch timeouts
- Web UI three-state selector with current/latest/last-check/status; check failure no longer toasts a false "up to date"

compress.nudgeGrowthTokens pins acp-kernel's adaptive nudge interval by setting nudge.growthFloor = nudge.growthCap = value; config PUT/reload hot-reloads kernelConfig, modelContextLimit, and compress for new requests (in-flight requests keep their snapshot).

Tests: update-mode (auto/check/manual semantics, precedence, legacy migration, concurrent lock, proxy egress, Web UI endpoints) and nudge-growth-tokens (adaptive default, floor=cap, breakdown denominator, hot reload, emergencyThresholdPct survival).


ranxianglei:
## Review feedback — two design concerns

### 1. `nudgeGrowthTokens` should be per-URL, not global

`nudgeGrowthTokens` is model-dependent — different upstreams serve different models with different context windows and different optimal compression thresholds:

| URL | Model | Context | Reasonable nudgeGrowthTokens |
|---|---|---|---|
| `chatgpt.com/backend-api/codex` | gpt-5.6-codex | 400K | ~50K |
| `api.anthropic.com` | claude-opus-4 | 200K | ~20K |
| `open.bigmodel.cn` | glm-5.2 | 1M | ~100K |

A global value is wrong for all but one model: too small for large-context models (frequent over-trigger), too large for small-context models (context blows up before compression).

**Suggested fix:** move it to `routes.<url>.nudgeGrowthTokens`, same pattern as `context` and `protocol`:

```jsonc
{
  "routes": {
    "https://chatgpt.com/backend-api/codex": {
      "context": 400000,
      "nudgeGrowthTokens": 50000
    }
  }
}
```

Implementation: `resolveConfiguredNudge(routes, url)` with longest-prefix match (mirrors `resolveConfiguredContextLimit`), then per-request override `reqConfig.nudge.growthFloor/growthCap` alongside the existing `modelContextLimit` override in `server.ts`.

Global `compress.nudgeGrowthTokens` could remain as a fallback when URL has no override.

---

### 2. "check" update mode is unnecessary — keep it clean

The three-state model (auto/check/manual) adds complexity without clear value. The clean design is **binary**:

- **auto** — check + install automatically
- **manual** — never check in background; user triggers explicitly when they want to update

"check" (check but don't install) is a middle ground that creates confusion:
- Web UI "检查更新" button calls `/check`, but in auto mode `runScheduledUpdate` actually **installs** — the button name doesn't match the behavior
- Users who want "just tell me" can simply leave it on manual and check npm themselves

Removing "check" simplifies: one less state in the config precedence chain, one less UI option, one less mode-dependent code path in `runScheduledUpdate`.

---

### Also: 🔴 critical hot-reload bug

As noted separately: `const config: Config = opts.kernelConfig;` at `server.ts` startup captures the reference once. The PUT/reload handlers reassign `opts.kernelConfig = fresh.kernelConfig` (new object), but the closure `config` const still points to the OLD object — so hot-reload of kernelConfig **doesn't actually take effect** on real requests. The test only checks `opts.kernelConfig`, not what `handle()` uses, so it's a false positive.

**Fix:** read `opts.kernelConfig` per-request inside `handle()` instead of capturing at startup, OR mutate in-place (`Object.assign`) instead of reassigning.

ranxianglei:
这个冲突过多,可能和master落后比较多 暂时无法合并了,建议追赶一下

ranxianglei:
Requesting changes — this one can't be merged as-is, and it's not a code-quality rejection, it's a staleness problem.

The branch is based on `5297faa` (the #82 release, ~3 weeks old) and is now **CONFLICTING** across 15 files. In that time master moved a lot, and part of this PR's surface has already landed a different way:

- `nudgeGrowthTokens` + `nudgeGrowthTokensSource` are **already in master** `src/config.ts` (line 88, and wired into the validation loop at line 494). So the `compress.nudgeGrowthTokens` knob + `parseNudgeGrowthTokens` + the kernel `growthFloor`/`growthCap` override you added are duplicating config that's already merged. That whole hunk in `config.ts` (and the matching `nudge-growth-tokens.test.ts`, 248 lines) should drop out of this PR.

What's genuinely new and still worth having in this PR:
- the 3-state `UpdateMode` (`auto`/`check`/manual`) + deprecating the boolean `autoUpdate`,
- the `installUpdate` / `checkLatestVersion` / `runScheduledUpdate` refactor in `update.ts`,
- the `bili update` CLI command and the 3-state Web UI selector.

**To land this, please rebase onto current master and:**
1. Drop the `nudgeGrowthTokens` config/type/parsing additions and the `nudge-growth-tokens.test.ts` — already in master, keeping them is what's causing most of the `config.ts` conflict.
2. Re-resolve `update.ts`, `config.ts`, `server.ts`, `cli.ts`, and the `web/*` files against current master (the update-endpoint region of `server.ts` and the `cli.ts` arg parsing have both moved).
3. Keep the `updateMode`/`autoUpdate`-deprecation migration path — that's the part master doesn't have yet, and make sure the legacy `autoUpdate: true/false → auto/manual` mapping is covered by `update-mode.test.ts` (the 320-line suite looks solid, keep it).
4. Re-run the full suite after the rebase; the `upstream-proxy-routing` / `web-routing` / `zero-config-routing` test touch-ups need to re-verify against the current routing code.

I don't want to hand-resolve a 15-file conflict on a 3-week-old base — a clean rebase by the author is safer and will surface any real semantic overlap with what's already on master. Happy to re-review the moment it's rebased and green.

ranxianglei:
**Review vs master 3cd85b6:** ⚠️ mixed staleness — one half valuable, one half superseded. Full rework or extract-and-close.

**Half 1 — three-state update mode (auto/check/manual): still valuable**, and better-engineered than the surface it replaces: per-call `throttleFile()/lockFile()` (XDG-relocation friendly), registry egress through `proxyDispatcher` (undici ProxyAgent — users behind corporate proxies currently time out on update checks), Web UI + `bili update` integration. Tracked as #606. Conflicts heavily with master's `src/update.ts` (Windows tar fix, etc.) — and AGENTS.md §5 applies: updater changes ship only after a validated no-op release.

**Half 2 — `nudgeGrowthTokens` host-side tuning + interval diagnostics: superseded.** Nudge cadence is kernel-owned since kernel 0.0.52's pressure-band (min-benefit gate, `bestPending >= max(5000, limit*1%)`); the host-side interval/floor plumbing and `diagNudge` denominator rework here target a layer that no longer exists in this form. Also overlaps #204's channel work on the config surface.

Recommendation: close this PR with thanks after extracting (1) into a fresh branch, or rework in place keeping only the UpdateMode surface. External contribution — coordinate with the author (LX666-666).

(tracking: #606)

PR #87, opened 2026-08-10
fix: use openai_base_url for built-in provider + 426 for WS upgrade (closes #83)

### LINKED ISSUE #83: Codex route breaks on reserved openai provider ID and hides existing threads
## Summary

`bili codex enable` is incompatible with recent Codex builds because it writes a custom provider using the reserved built-in ID `openai`.

This has two user-visible failure modes:

1. The Codex Windows app can get stuck on the Windows setup screen because `windowsSandbox/setupStart` rejects the invalid config before UAC is launched.
2. Renaming the provider to `openai-custom` makes the app start, but existing projects/threads disappear from the UI because thread listing is filtered by the active provider ID. The data is still present under provider `openai`.

## Environment

- Windows 11
- `billion-context` 0.1.29 (also the current npm latest at time of testing)
- Codex desktop package: `OpenAI.Codex` 26.803.5235.0
- Codex CLI: 0.147.0-alpha.6.5

## Reproduction

1. Start with the built-in Codex provider (`model_provider` omitted or set to `openai`).
2. Run:

   ```console
   bili codex enable --port 8787
   ```

3. Observe that billion-context appends a section like:

   ```toml
   [model_providers.openai]
   name = "OpenAI"
   wire_api = "responses"
   requires_openai_auth = true
   base_url = "http://127.0.0.1:8787/codex"
   supports_websockets = false
   ```

4. Restart Codex desktop or run any CLI command that loads the configuration.

## Actual Result

Codex rejects the generated configuration:

```text
model_providers contains reserved built-in provider IDs: `openai`. Built-in providers cannot be overridden. Rename your custom provider (for example, `openai-custom`).
```

On Windows desktop, setup attempts fail before elevation with:

```text
method=windowsSandbox/setupStart
failureReason=invalid_config
errorCode=-32600
```

Changing the generated provider ID and active provider to `openai-custom` avoids the parse error, but hides existing projects/threads stored with `model_provider = "openai"`.

## Supported Codex Configuration

Current Codex has a top-level setting specifically for overriding the built-in OpenAI provider URL without changing its ID:

```toml
model_provider = "openai"
openai_base_url = "http://127.0.0.1:8787/codex"
```

Relevant Codex sources:

- [`ConfigToml::openai_base_url`](https://github.com/openai/codex/blob/89a335ed50258dc9dc5b3d7f410db61b431244f9/codex-rs/config/src/config_toml.rs)
- [Built-in `openai` provider construction](https://github.com/openai/codex/blob/89a335ed50258dc9dc5b3d7f410db61b431244f9/codex-rs/model-provider-info/src/lib.rs)
- [Thread listing defaults to the active provider ID](https://github.com/openai/codex/blob/89a335ed50258dc9dc5b3d7f410db61b431244f9/codex-rs/app-server/src/request_processors/thread_processor.rs)

## WebSocket Compatibility

The built-in `openai` provider currently advertises `supports_websockets = true`, and there is no supported config toggle to override that capability while keeping the reserved provider ID.

With `openai_base_url` pointed at billion-context, each new Codex session probes:

```text
ws://127.0.0.1:8787/codex/responses
```

billion-context currently forwards the upgrade through `fetch`, which returns repeated `502 Bad Gateway` responses before Codex eventually falls back to HTTP and succeeds:

```text
UND_ERR_INVALID_ARG ... invalid upgrade header
```

Codex has an explicit fast fallback for HTTP `426 Upgrade Required`. Returning `426` for a WebSocket upgrade request to `GET /codex/responses`, while continuing to proxy normal `POST /codex/responses`, makes the fallback immediate and sticky for the session:

- [Codex WebSocket fallback test](https://github.com/openai/codex/blob/89a335ed50258dc9dc5b3d7f410db61b431244f9/codex-rs/core/tests/suite/websocket_fallback.rs)

## Suggested Fix

1. For built-in provider `openai`, manage the top-level `openai_base_url` field instead of creating `[model_providers.openai]`.
2. Preserve `model_provider = "openai"` so existing threads remain visible across account switches.
3. Snapshot and restore the original top-level `openai_base_url` during enable/disable, stale-state recovery, and graceful shutdown.
4. Migrate existing takeover state that points at a generated `openai` or `openai-custom` section.
5. Until the proxy supports Responses WebSockets, return `426` for the `/codex/responses` WebSocket upgrade instead of forwarding the upgrade headers through `fetch`.

## Expected Result

- `bili codex enable` produces valid configuration on current Codex builds.
- Existing projects and threads remain visible because the provider ID stays `openai`.
- `bili codex disable` and shutdown safely restore the previous `openai_base_url`.
- New sessions fall back to HTTP immediately rather than logging several `502` retries.

ranxianglei:
Closes #83.

## Problem

`bili codex enable` writes `[model_providers.openai]` which overrides the reserved built-in provider ID `openai`. Recent Codex builds (0.147+, desktop 26.803) reject this:

> "model_providers contains reserved built-in provider IDs: `openai`. Built-in providers cannot be overridden."

Two user-visible failures:
1. **Codex rejects config** → Windows setup stuck / CLI error
2. **Renaming to `openai-custom`** → existing threads/projects disappear (filtered by provider ID)

### 1-3. Top-level `openai_base_url` for built-in `openai`
When the active provider is the built-in `openai` (no `[model_providers.openai]` section exists), do NOT create that section. Instead:
- Write the **top-level** `openai_base_url = "http://127.0.0.1:{port}/codex"`
- Preserve `model_provider = "openai"` (untouched — so existing threads stay visible)
- Snapshot/restore the original `openai_base_url` during enable/disable/stale-recovery

Custom providers with an existing section keep the current behavior (`base_url` in the provider section).

### 4. State migration
- `CODEX_ROUTE_STATE_VERSION` bumped 1 → 2
- New `baseUrlMode: "top-level" | "provider-section"` field
- `parseState` accepts both v1 and v2 states
- Old v1 state with `sectionAdded: true` for `openai` defaults to `baseUrlMode: "provider-section"` → next `disable` cleans up the old section via existing logic

### 5. WebSocket 426 for `/codex/responses` upgrade
Built-in `openai` advertises `supports_websockets = true`. Since we no longer create a section to disable it, Codex sends WS upgrade probes. Server now returns **426 Upgrade Required** immediately → Codex's built-in fast-fallback kicks in → immediate HTTP POST (no more repeated 502s).

## New helpers in `codex-provider.ts`
`getTopLevelField`, `getTopLevelFieldSnapshot`, `setTopLevelField`, `restoreTopLevelField`, `removeTopLevelField` — follow the existing TOML editing pattern (index-based string manipulation, `rootAssignment` finds assignments before the first section).

## Tests
- **built-in openai uses top-level `openai_base_url`** (updated existing test): no `[model_providers.openai]` created, `openai_base_url` set at top level
- **disable restores original `openai_base_url`**: snapshot/restore verified
- **migration: old v1 state with openai sectionAdded cleaned up on disable**: v1 state file → disable → section removed
- **426 for WebSocket upgrade to /codex/responses** (codex-official.test.ts): WS upgrade header → 426 response

**227 tests pass, typecheck clean, build OK.**

### DIFF
--- a/src/codex-provider.ts
+++ b/src/codex-provider.ts
@@ -283,6 +283,50 @@ export function removeCodexProviderField(text: string, providerId: string, key: 
     return existing ? text.slice(0, existing.lineStart) + text.slice(existing.lineEnd) : text;
 }
 
+export function getTopLevelField(text: string, key: string): ProviderFieldValue | undefined {
+    return rootAssignment(text, key)?.value;
+}
+
+export function getTopLevelFieldSnapshot(text: string, key: string): ProviderFieldSnapshot | undefined {
+    const assignment = rootAssignment(text, key);
+    return assignment ? { value: assignment.value, encoded: assignment.encoded } : undefined;
+}
+
+export function setTopLevelField(
+    text: string,
+    key: string,
+    value: ProviderFieldValue,
+): { text: string; added: boolean } {
+    const existing = rootAssignment(text, key);
+    if (existing) {
+        return {
+            text: text.slice(0, existing.start) + encodeValue(value) + text.slice(existing.end),
+            added: false,
+        };
+    }
+    const eol = preferredEol(text);
+    const firstSection = linesOf(text).find((line) => /^\s*\[/.test(line.content));
+    const insertAt = firstSection?.start ?? text.length;
+    const prefix = insertAt > 0 && !/(?:\r\n|\r|\n)$/.test(text.slice(0, insertAt)) ? eol : "";
+    const line = `${key} = ${encodeValue(value)}${eol}`;
+    return { text: text.slice(0, insertAt) + prefix + line + text.slice(insertAt), added: true };
+}
+
+export function restoreTopLevelField(
+    text: string,
+    key: string,
+    snapshot: ProviderFieldSnapshot,
+): string {
+    const existing = rootAssignment(text, key);
+    if (!existing) throw new Error(`Top-level field "${key}" is missing`);
+    return text.slice(0, existing.start) + snapshot.encoded + text.slice(existing.end);
+}
+
+export function removeTopLevelField(text: string, key: string): string {
+    const existing = rootAssignment(text, key);
+    return existing ? text.slice(0, existing.lineStart) + text.slice(existing.lineEnd) : text;
+}
+
 export function appendCodexProviderSection(
     text: string,
     providerId: string,
--- a/src/codex-takeover.ts
+++ b/src/codex-takeover.ts
@@ -14,21 +14,25 @@ import { createHash, randomUUID } from "node:crypto";
 import path from "node:path";
 import { dataDir } from "./paths.js";
 import {
-    appendCodexProviderSection,
     codexConfigFile,
     getCodexProviderField,
     getCodexProviderFieldSnapshot,
+    getTopLevelField,
+    getTopLevelFieldSnapshot,
     readCodexConfig,
     removeCodexProviderField,
     removeCodexProviderSectionWithPrefixIfEmpty,
+    removeTopLevelField,
     restoreCodexProviderField,
+    restoreTopLevelField,
     resolveActiveCodexProvider,
     setCodexProviderField,
+    setTopLevelField,
     type ActiveCodexProvider,
     type ProviderFieldValue,
 } from "./codex-provider.js";
 
-export const CODEX_ROUTE_STATE_VERSION = 1;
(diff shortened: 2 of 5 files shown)


ranxianglei:
⚠️ **Superseded** — per maintainer decision, the auto-takeover feature is being removed entirely (refactor PR coming). This fix (openai_base_url) is no longer needed since the `bili codex enable/disable` command itself is being deleted. Users will configure manually instead. This PR can be closed.

PR #144, opened 2026-08-14
fix(responses): pass non-proxy function_call items through raw (Codex agents.* unsupported call)

### LINKED ISSUE #143: 经过 Billion Context 后，codex无法打开子代理
Codex 直连上游时多代理功能正常，但经过 Billion Context 后，agents.spawn_agent 等多代理工具均返回 unsupported call，原因是 Billion 当前没有正确透传或分派 Codex 的 agents.* 命名空间工具调用。
我用的是codex++配置的，具体config.toml的配置放文件里面了。

[BILLION-CONTEXT-多代理工具调用问题反馈.md](https://github.com/user-attachments/files/31053304/BILLION-CONTEXT-.md)

ranxianglei:
## Problem

Fixes #143.

Codex Desktop multi-agent tools (e.g. `agents.spawn_agent`, `agents.list_agents`) return `unsupported call: spawn_agent` (note: the `agents.` namespace is dropped) when routed through billion-context, but work when connecting directly to the upstream. A/B confirmed: only the base_url (`8787/bili/...` vs direct) differs. Reported on 0.1.40 (which already includes #94 and #138).

## Root cause

The Responses adapter swallowed **every** `function_call` item (`output_item.added` / `function_call_arguments.delta` / `.done` / `output_item.done`) into a pending map and re-emitted it via `buildFunctionCallEvents` with:
- a **new synthetic item id** (`fc-proxy-...`), while `response.completed.output` still carries the upstream's **original** id (id mismatch), and
- the arguments collapsed into a single delta.

This reconstruction is required for proxy tools the loop must intercept (`compress`/`decompress`/`search_context`/`acp_status`), but it corrupts the wire format for real client-side tools. `custom_tool_call` items already passed through raw; `function_call` did not — which is exactly why normal (custom-tool) tools kept working while namespaced `function_call` tools (Codex `agents.*`) failed.

I verified exhaustively that billion-context does **not** strip or rewrite the namespaced name itself (new tests prove `agents.spawn_agent` survives both textProtocol and non-textProtocol) — the breakage is the reconstruction, not the name.

## Fix

Only intercept `function_call` items whose name is in `PROXY_TOOL_NAMES`; all other `function_call` items are forwarded as raw meta (byte-identical SSE), exactly like `custom_tool_call`. A `passthrough` flag on the `tool_call` event lets the loop count the call (so it still suppresses the re-request when a real call was made) without re-emitting it.

Files:
- `src/loop/adapter-responses.ts` — proxy/non-proxy branching in the 4 function_call handlers.
- `src/loop/core.ts` — `passthrough` on `ToolCallEmit`/`ParsedStreamEvent`; skip `emitToolCall` for passthrough calls.
- `tests/multi-agent-namespace.test.ts` — proxy interception, raw passthrough (original item id + name + call_id + args preserved, no `fc-proxy` reconstruction), both protocols, tool-declaration survival.

## Verification

- `node --import tsx --test tests/*.test.ts` → **375/375 pass** (no regressions).
- `tsc --noEmit` → `src/loop/*` changes are type-clean (the 4 tsc errors are pre-existing in unrelated test files on master).
- New tests prove the forwarded SSE for `agents.spawn_agent` now carries the **original** item id (`fc_ns_1`), not a reconstructed `fc-proxy-...` id.

## Caveat

I could not run Codex Desktop locally to reproduce end-to-end, so the codex-side dispatch fix is a **strong, well-reasoned hypothesis** grounded in the A/B report + the reconstruction difference vs `custom_tool_call`. The change is a strict improvement regardless (more transparent passthrough of tools billion-context has no business intercepting). Requesting the reporter (@a1langxin) to confirm with a build from this branch.

### DIFF
--- a/src/loop/adapter-responses.ts
+++ b/src/loop/adapter-responses.ts
@@ -1,7 +1,7 @@
 import type { CoreMessage } from "acp-kernel";
 import { coreToResponses, injectResponsesDeveloperMessage, patchResponsesInput, type ResponseInputItem, type ResponsesProjection } from "../responses.js";
 import { buildVisibilityMarker } from "../compress-loop.js";
-import { ACP_TEXT_OPEN, ACP_TEXT_CLOSE, ACP_STATUS_OPEN, ACP_STATUS_CLOSE, ACP_SEARCH_OPEN, ACP_SEARCH_CLOSE, ACP_DECOMPRESS_OPEN, ACP_DECOMPRESS_CLOSE, COMPRESS_TOOL_NAME } from "../compress-tool.js";
+import { ACP_TEXT_OPEN, ACP_TEXT_CLOSE, ACP_STATUS_OPEN, ACP_STATUS_CLOSE, ACP_SEARCH_OPEN, ACP_SEARCH_CLOSE, ACP_DECOMPRESS_OPEN, ACP_DECOMPRESS_CLOSE, COMPRESS_TOOL_NAME, PROXY_TOOL_NAMES } from "../compress-tool.js";
 import type { BiliMessage } from "../bili-message.js";
 import type {
     CompressLoopAdapter,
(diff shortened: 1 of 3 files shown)


a1langxin:
用了此修复，可以正常调用了。

PR #145, opened 2026-08-14
fix(prompts): exec-sandbox warning in compress prompts + update restart-pending nag

ranxianglei:
## Background

Postmortem of the Codex context explosion (dog/billion-context#24 floor 521, tracked in dog/billion-context-pi#37):

1. **Exec-sandbox tool confusion** — the model, blind to its context state in marker mode, called `tools.acp_status()` / `tools.decompress()` / `tools.search_context()` inside the codex exec sandbox → 7× `TypeError: tools.acp_status is not a function` → compensated with 247 repeated file reads (5.57M input tokens / 2h).
2. **Phantom up-to-date** — 0.1.40 (which fixes the marker-mode default) was auto-installed mid-session but the never-restarted proxy kept running 0.1.39; the only notice was a single info-level `Restart to finish.` line, and later registry checks saw the DISK version as current, so staleness was invisible.

## Changes

- `src/compress-tool.ts` — all three prompt builders (`buildCompressSystemPrompt` / `buildCompressTextSystemPrompt` / `buildCompressHybridSystemPrompt`) now carry an explicit warning: ACP tools are NOT callable inside the code-execution sandbox (quoting the exact TypeError the model saw), with the correct alternative spelled out (top-level function call / text marker).
- `src/update.ts` — `checkForUpdate` now WARNs `restart pending: running X but Y is installed — RESTART the proxy to activate it` when the on-disk version is newer than the running process (throttled to 30 min to avoid log spam at the 3-min check cadence).
- `tests/fix-exec-sandbox-prompt.test.ts` — 4 new tests (prompt content across all 3 modes; nag fires exactly once across two consecutive checks).
- `devlog/2026-08-14_marker-prompt-strengthen/` — REQ + WORKLOG.

## Verification

- `npm run typecheck` clean
- `npm test`: **374/374 pass** (370 prior + 4 new)

Note: the Windows auto-update `spawn cp ENOENT` bug from the same incident is already fixed on master (26948d3, released in 0.1.40).

### DIFF
--- a/devlog/2026-08-14_marker-prompt-strengthen/REQ.md
+++ b/devlog/2026-08-14_marker-prompt-strengthen/REQ.md
@@ -1,0 +1,31 @@
+# REQ - Harden compress prompts against exec-sandbox tool confusion + restart-pending nag
+
+- Task ID: `2026-08-14_marker-prompt-strengthen`
+- Home Repo: `billion-context`
+- Created: 2026-08-14
+- Status: Done
+- Priority: P1
+- Owner: awork
+- References: dog/billion-context-pi#37; dog/billion-context#24 floors 521–542 (context-explosion postmortem)
+
+## 1. Background & Problem Statement
+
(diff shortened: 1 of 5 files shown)


ranxianglei:
先关闭这个,1.4.0应该修复了 如果有问题重新打开

PR #157, opened 2026-08-15
feat(compress): configurable compression prompts (three-level, risk-gated)

ranxianglei:
Partially addresses #156 (the custom-prompts suggestion).

## What

Adds `compress.prompts` + `compress.acknowledgePromptsRisk` to the existing three-level cascade (global → provider → model):

- `prompts` overrides the kernel prompt text (`compressPhilosophy`, `howToCompressRules`, `tier2DistillRules`, `tier3CondenseRules`), merged **sub-field-wise** across levels — a model-level `howToCompressRules` does not discard a provider-level `compressPhilosophy`.
- Resolved via the kernel's `resolvePrompts` (non-string fields silently dropped; a malformed partial never clobbers a good default).
- **Risk gate**: the kernel rules are load-bearing (tuned in production); overrides are inert until `acknowledgePromptsRisk: true` at the winning level. Without it: one-time `[compress] prompts override IGNORED` warning + defaults.
- Threaded consistently into the system prompt (all three protocols), the nudge text (`renderNudgeText`), and the compress loop in `forward()` — kernel contract: both layers must use the same `Prompts`.

Main use case: non-English / small-model prompt tuning (#156: 2B model spinning in the compress loop).

## Stacking

Stacked on #153 (admin follow-up) — merge that first, then retarget this to master.

## Verification

- `npm run typecheck` clean; `npm test` 402/402 (6 new unit tests: three-level sub-field merge, risk-gate fallback, non-string drop, end-to-end route cascade).
- `npm run build` clean (2.04 MB).
- Docs: CONFIGURATION.md + CONFIGURATION.zh-CN.md field entries; CHANGELOG unreleased section. Web UI needs no change (settings page edits raw providers JSON).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@
 All notable changes to **billion-context** are documented here.
 Versions follow the merge of a `*_release-v*` branch; CI publishes to npm on tag.
 
+## [Unreleased]
+
+### Features
+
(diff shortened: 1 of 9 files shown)


ranxianglei:
Cross-linking: this implements the custom-prompts suggestion from #156. The remaining #156 suggestions (repeat-failure short-circuit in the compress loop, repair hints in validation errors) are separate follow-ups.

PR #155, opened 2026-08-14
fix: resolve issues #150 #151 #152 #154 (24h issue batch)

ranxianglei:
Fixes #150. Fixes #151. Fixes #152. Fixes #154.

One commit per issue, on top of master 77ea877.

## #152 — combined CA bundle for replace-semantics env vars (`fix(ca)`)
`SSL_CERT_FILE` (codex launcher env) *replaces* the default CA bundle, but `root-ca.pem` only contained the MITM root — pip/git/curl inside the child env failed TLS on blind-tunnelled domains (`unable to get local issuer certificate`).
- `writeCombinedBundle()` in `src/ca.ts`: system CA bundle (`SSL_CERT_FILE` if set, else platform candidates) + Node `tls.rootCertificates` + MITM root, deduped, written to `ca/combined-ca.pem` on every root-CA ensure path.
- Launcher points codex's `SSL_CERT_FILE` at the combined bundle; pi/claude keep `NODE_EXTRA_CA_CERTS` → `root-ca.pem` (append semantics, unchanged).
- Tests: `tests/ca-combined.test.ts` (5) — content, re-merge, SSL_CERT_FILE priority, dedup, path layout.

## #154 — web-editable + hot-reloaded global compress settings (`fix(web)`)
PUT `/__bili/config` rejected the `compress` block (400), and manual config-file edits didn't hot-apply (reload re-read only routes).
- PUT now accepts `compress` (validated by new `parseCompressSettings` in `src/config.ts`; `null` clears), GET returns it, and both the PUT path and `POST /__bili/config/reload` refresh `opts.compress` — engine tuning applies without restart.
- Web UI settings page: 压缩参数 card (JSON textarea + save).
- Tests: validator unit test + full HTTP round-trip in `tests/web-routing.test.ts`.

## #151 — `bili export` session handoff (`feat(cli)`)
Stopping the proxy makes a long compressed session replay its full local history → `context_window_exceeded`, with no way to leave gracefully.
- `bili export [id|label] [--full] [--output f.md]` renders persisted sessions (`~/.local/share/billion-context/sessions/`) as a Markdown handoff: header stats + block summaries (+ original messages with `--full`).
- README documents the proxy-dependency and the migration procedure.
- Tests: `tests/cli-export.test.ts` — listing, id/label/prefix selection, --full/--output, no-match error, parseArgs forms.

## #150 — separate compression namespace for Codex subagent requests (`fix(responses)`)
[...]

PR #159, opened 2026-08-15
feat(export): persist v3 conversation snapshot + folded handoff view

ranxianglei:
Supersedes the export portion of #155 (built on top of that branch; merge #155 first or rebase after).

## What

`bili export` now renders the **complete conversation**, not just compression block summaries.

- `Session.lastMessages`: snapshot of the core message list at each request (`snapshotMessages()` in `session.ts`, called from all three prepare paths in `server.ts`); an empty snapshot never clobbers a good one
- `persist.ts`: `PERSIST_VERSION` 2→3 — `PersistedSession.messages` round-trips through `buildRecord`/`buildSession`
- `export.ts`: with a snapshot, `renderHandoff` renders the conversation itself —
  - default = **folded view via kernel `prune()`**: exactly what the model saw (block summaries in place of compressed ranges, `[Compressed conversation section]` anchors)
  - `--full` = every original message
  - v2 session files (no snapshot) fall back to block summaries with an honest note that client messages are unavailable
- encrypted `reasoning` items export as `_reasoning_: <item id>` — ciphertext never enters the handoff doc (it does stay in the session file for upstream replay)

## Tests

- `tests/e2e-export-handoff.test.ts`: true e2e — real proxy `startServer` + mock upstream issuing a **real** `compress` tool call (range parsed from live `<acp>` tags) → real block → real `SessionStore` on disk → `exportSession` from a **fresh store**. Asserts folded view contains the model-written summary + uncompressed tail and **omits** compressed originals; `--full` keeps everything
- `tests/export-uncompressed.test.ts`: v3 full-render + v2 honest-fallback
- suite: **435/435 pass**

Note: kernel `prune` deliberately never folds the first user message (protected); the e2e asserts against a middle-of-history compression range to match that design.

Closes the export gap in #151.


ranxianglei:
并入 #155: 本 PR 与 #155 是叠加关系 (头分支包含 #155 全部 commit), 拆开反而要按顺序合并。快照导出增强以 commit 11b9191 (pr155-merge 分支) 为准, 将随 #155 一起合并。

PR #163, opened 2026-08-16
feat(launcher): plugin-in-launcher mode — native MCP tools via spawn-time injection (#162)

ranxianglei:
Implements #162 (MVP scope) on top of #161's plugin protocol.

## What

`bili claude` / `bili codex` now deliver the native-plugin experience with **zero persistent host config**: the launcher injects a single `bili` MCP server at spawn time — nothing written to host config files, **traffic route unchanged by default** (see Compatibility), zero residue on uninstall.

- **`POST /__bili/plugin/register {conversationId, agent, identity}`** — session pre-binding, two strategies:
  - **identity: true** (claude code): every model request carries `x-claude-code-session-id` === the `CLAUDE_CODE_SESSION_ID` passed to MCP children (both verified against claude 2.1.227 — raw-socket header capture + child env probe). Binding is identity-keyed → race-free regardless of arrival order (claude -p fires its first model request concurrently with MCP initialize).
  - **identity: false** (headless codex spawn): the launcher injects a per-spawn UUID as `BILI_CONVERSATION_ID` into the MCP child (codex passes no session id of its own — verified codex-cli 0.147.0); the shell self-registers and the pending entry is consumed by the first request that creates a NEW session. The shell awaits the register before answering MCP initialize so waiting hosts cannot race past it.
- **`src/mcp.ts`** — MCP stdio thin shell: GET manifest (zero schema drift) → expose 4 tools → POST /__bili/plugin/tool. Second tsup entry (`dist/mcp.js`). Also reachable as `bili mcp`.
- **`bili plugin-register <id>`** — CLI hook for manual registration (sends `identity: true`).
- **Launcher direct-URL mode** (opt-in, `BILI_LAUNCHER_DIRECT=1`; the default is the classic transparent-MITM route): claude → `ANTHROPIC_BASE_URL` spawn env + `--mcp-config` ephemeral JSON; codex → `-c mcp_servers.bili.*` inline TOML overrides (syntax accepted by codex 0.147). Enabling direct mode prints a client-specific warning describing the route change. `BILI_LAUNCHER_PLUGIN=0` disables injection (plain wire mode).

## 待验证 items from #162 — resolved

- ~~待验证① `--settings` SessionStart hook~~ — **obsoleted**: `CLAUDE_CODE_SESSION_ID` child env + `x-claude-code-session-id` request header make hooks unnecessary; simpler and race-free.
- ~~待验证② codex `-c` inline `mcp_servers.*`~~ — accepted by codex 0.147.0 (config parses; our e2e stopped only at provider routing because the local relay lacks a Responses endpoint — unrelated to the MCP surface).

## Compatibility (default behavior unchanged)

Initial review found the direct-URL route was **on by default**, which silently changed the traffic route of every existing `bili claude` / `bili codex` setup on upgrade:

| Existing setup | If direct URL were the default |
|---|---|
| codex (any) | LLM traffic no longer goes through the proxy → compression/memory silently stop working |
| claude (OAuth subscription) | subscription traffic breaks |
| claude (custom relay) | `ANTHROPIC_BASE_URL` silently bypassed (falls back to `https://api.anthropic.com` unless `BILI_CLAUDE_UPSTREAM` is set) |

The PR now keeps the **classic transparent-MITM route as the default** — an upgrade changes nothing for existing users. Direct URL remains available as an explicit opt-in: `BILI_LAUNCHER_DIRECT=1`; enabling it prints a client-specific warning (relay bypassed / LLM traffic not proxied).

MCP tool injection itself remains **on by default** — that part is purely additive: it only adds the `bili` MCP server, and if the server fails to start the host continues without it (verified: codex 0.147.0 tolerates a missing MCP command), so the worst case is the pre-PR tool surface. Escape hatch: `BILI_LAUNCHER_PLUGIN=0`.

## Bugs fixed on the branch after review

1. **`bili codex` startup crash** — `buildCodexMcpArgs` double-encoded the args value (TOML string where a sequence is expected); every injected codex spawn died with `Error loading config.toml: invalid type: string ..., expected a sequence`. Now emits a real TOML array; verified with codex-cli 0.147.0.
2. **`bili plugin-register` registered headless** — the documented `"$CLAUDE_SESSION_ID"` usage now sends `identity: true` so the registration binds by identity instead of consuming the next NEW session.
3. **Stale headless registrations** — pending registrations now expire after 10 min (`PENDING_REGISTER_TTL_MS`); an orphaned registration can no longer turn an unrelated later session into plugin mode with a foreign conversation id.
4. **Direct-URL default flip** — see Compatibility above.
5. **Codex native tools dead** — every `bili codex` tool call failed with `no conversation id`: the launcher never gave the MCP shell a conversation id to self-register with. `buildCodexMcpArgs` now injects `mcp_servers.bili.env.BILI_CONVERSATION_ID=<uuid>` per spawn. Binding works on the default MITM route (verified by e2e); in direct-URL mode the model traffic bypasses the proxy so the binding still cannot happen — tracked with the codex LLM-routing follow-up.

## Verification

- **452/452 tests** (444 base + 8 new: register validation, identity-after-first-request, headless pending, no cross-session leak, injection builders, MCP shell stdio round-trip, pending-register TTL expiry, codex-style self-register e2e; codex `-c` args assertion now verifies the value is a TOML array and the `BILI_CONVERSATION_ID` env line — would have caught the double-encoding and the missing conversation id)
- typecheck clean, build clean (dist/mcp.js emitted)
- **Real e2e**: `bili claude -p "call the bili acp_status tool"` through a live anthropic-compatible upstream → proxy log `[plugin] tool acp_status executed via plugin` (native MCP path; wire injection suppressed), model answers from the real status; host config untouched; tmp MCP config removed on exit
- **codex config e2e**: launcher-generated `-c` args fed back into codex-cli 0.147.0 → config parses, session banner reached (pre-fix: startup failure)
- **codex binding e2e**: real MCP shell over stdio with only `BILI_CONVERSATION_ID` (no `CLAUDE_CODE_SESSION_ID`, no manual register) → headless self-register on initialize → first NEW-session model request binds the conversation → `tools/call acp_status` returns the live context breakdown

## Note

Codex full e2e (LLM round-trip through a Responses upstream) still pending a suitable upstream — the MCP/config surface and the headless binding are verified independently (see above). Follow-ups (tracked separately, not blockers): direct-URL mode for codex — LLM traffic routing through the proxy (required before headless binding can work there), i.e. one combined 'codex direct mode' follow-up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,16 +104,21 @@ bili pi --mitm-domain api.foo.com     # add a domain to the MITM whitelist
 **Plugin-in-launcher mode (#162)** — for `claude` and `codex` the launcher
 additionally injects a single `bili` MCP server (`--mcp-config` for claude,
 `-c mcp_servers.bili.*` for codex — both ephemeral, nothing written to host
-config) and routes via **direct URL** (claude's `ANTHROPIC_BASE_URL` /
-codex's provider `base_url` pointing at the `/bili/` prefix — no MITM, no CA
-trust). The result is the native-plugin experience of [PLUGIN.md](PLUGIN.md)
+config). The traffic route is **unchanged by default** — the transparent-
+MITM proxy as before — so existing setups keep working exactly as they did
+(OAuth-subscription traffic, custom relay endpoints). Opt in to **direct URL
+routing** (`BILI_LAUNCHER_DIRECT=1`) to drop MITM/CA trust entirely (claude's
+`ANTHROPIC_BASE_URL` / codex's provider `base_url` pointing at the `/bili/`
+prefix); the launcher prints a warning when direct mode changes your traffic
+route. The result is the native-plugin experience of [PLUGIN.md](PLUGIN.md)
 with zero setup: the four ACP tools appear as native MCP tools, executed on
 the proxy under the session lock, while the proxy keeps state, folding,
 philosophy prompt and nudges. Session binding is automatic — claude passes
 its session id to MCP children and on every request; codex spawns bind on
 first sight. Opt-outs: `BILI_LAUNCHER_PLUGIN=0` (plain launcher, wire-injected
-tools), `BILI_LAUNCHER_MITM=1` (old transparent-MITM route, needed for OAuth
-subscription traffic). Mode matrix:
+tools). `BILI_LAUNCHER_DIRECT=1` is an opt-in, not an opt-out.
+
+Mode matrix:
 
 | Mode | Tools surface | Setup | When |
 |------|--------------|-------|------|
--- a/README.md
+++ b/README.md
@@ -115,15 +115,18 @@ with zero setup: the four ACP tools appear as native MCP tools, executed on
 the proxy under the session lock, while the proxy keeps state, folding,
 philosophy prompt and nudges. Session binding is automatic — claude passes
 its session id to MCP children and on every request; codex spawns bind on
-first sight. Opt-outs: `BILI_LAUNCHER_PLUGIN=0` (plain launcher, wire-injected
-tools). `BILI_LAUNCHER_DIRECT=1` is an opt-in, not an opt-out.
+first sight. Plugin mode is opt-in while host-flag compatibility soaks: set
+`BILI_LAUNCHER_PLUGIN=1` to enable the native MCP tools (verified with
+claude 2.1.227 / codex 0.147.0; `BILI_LAUNCHER_PLUGIN=0` forces the plain
+wire-injected launcher once the default flips on).
+`BILI_LAUNCHER_DIRECT=1` is an opt-in, not an opt-out.
 
 Mode matrix:
 
 | Mode | Tools surface | Setup | When |
 |------|--------------|-------|------|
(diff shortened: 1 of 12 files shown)


ranxianglei:
Review follow-up: plugin-mode MCP injection is now **opt-in** (`BILI_LAUNCHER_PLUGIN=1`) — f796afe.

Spawn 参数不再默认变更:存量 `bili claude`/`bili codex` 用户保持字节级行为不变,直到注入 flag 经过一个 soak 版本(已实证 claude 2.1.227 / codex 0.147.0,更老版本未测)。未开启时 launcher 打一行 stderr 提示如何 opt-in。判定提成 `launcherInjectMcp()` 并加了单测(pi 恒排除,走 billion-context-pi #154 原生扩展)。

另:本 PR 分支缺 #161 的 f0d3ebd(会话身份头门控修复),但 163 fork 后从未动过 `session-id.ts`,三方合并会自动取 master 的门控版 —— **合并顺序保持 161 → 163 即可,无需手动 rebase**(合并后点一次 Update branch 让 CI 重跑更稳)。默认翻回开启时改 `!== "0"` 语义并保留 opt-out。

PR #168, opened 2026-08-17
fix(upstream): direct-connect path gets a 10s handshake timeout

### LINKED ISSUE #78: 直连上游路径缺少连接超时(半开 socket 滞留 ~2 分钟)[master/pre-existing]
## 问题

出站连接的**直连路径**(无上游代理时)完全没有连接超时,只依赖 OS 默认的 ~127s(`tcp_syn_retries`)。当上游不可达(防火墙丢包、路由黑洞),半开 socket 会滞留约 2 分钟,期间占用资源且用户长时间无反馈。而**代理路径**(`openUpstreamSocket` 的 proxied 分支)有 10s 握手超时——两条路径行为不一致。

## 位置(master, pre-existing)

`src/upstream-proxy.ts:116` 和 `:125`(直连分支)。**PR #73 大幅重构了 `upstream-proxy.ts` 但未修此缺陷。**

```js
// src/upstream-proxy.ts (master)
// 直连分支 —— 无 setTimeout
const sock = net.connect(port, host);
sock.once("connect", () => resolve(sock));
sock.once("error", reject);     // ← 无连接超时,依赖 OS ~127s

// 代理分支(同文件 ~:131-144)—— 有超时
const handshakeTimer = setTimeout(() => {
    sock.destroy();
    reject(new Error(`upstream proxy CONNECT ${host}:${port} handshake timeout`));
}, ...);
```

## 影响

- 上游不可达时半开 socket 滞留 ~2 分钟(配合 #3 的 mitm 超时竞态更严重)
- 直连与代理路径行为不一致,排障困惑
- 高负载下半开 socket 堆积

## 建议修复

- 直连分支同样加连接超时(与代理分支对齐,如 10s)
- `net.connect` 失败/超时时返回 socket 引用供调用方销毁
- 补测试:连一个不响应的端口(如 `127.0.0.1:1` 会立即 RST,需用 iptables 丢包 或 backlog 满的端口模拟超时),断言在秒级而非分钟级失败

ranxianglei:
## Closes #78

The direct (no upstream proxy) connect path in `src/upstream-proxy.ts` had **no connection timeout** — it relied on the OS default `tcp_syn_retries` (~127s). A black-holed upstream (firewall drops SYNs, routing hole) left half-open sockets hanging for ~2 minutes while the proxied branch (same file) already had a 10s CONNECT handshake timeout. Inconsistent behavior, resource hold, no user feedback.

## Change

- Extract the direct path into `connectDirect(host, port, timeoutMs = 10_000)` with the same settled-guard + destroy + timer-cleanup pattern as the proxied branch
- `connectThroughProxy` delegates to it when no proxy is set → covers both plain and TLS outbound (`connectTlsThroughProxy` reuses `connectThroughProxy`)
- The proxy branch's hardcoded `10_000` now shares the `CONNECT_TIMEOUT_MS` constant — both paths fail in 10s by construction

## Tests (+4, 441/441 green)

- real-server success via both `connectDirect` and `connectThroughProxy(…, undefined)`
- immediate `ECONNREFUSED` still surfaces right away (no timeout tax on the fast-fail path)
- hang simulation: a never-completing socket (injected via the repo's established `_setConnectFactoryForTest` seam — ESM namespaces preclude `net.connect` monkey-patching) rejects with `upstream connect … timed out after 50ms` in milliseconds and destroys the socket

## Behavior delta

Only the failure path changes: unreachable upstreams now error in ~10s instead of ~127s. Successful connections are byte-for-byte unchanged (same socket, same events, timer just cleared earlier).

### DIFF
--- a/src/upstream-proxy.ts
+++ b/src/upstream-proxy.ts
@@ -302,13 +302,47 @@ function openProxySocket(proxy: ParsedHttpProxy): net.Socket {
     return net.connect(proxy.port, proxy.host);
 }
 
+const CONNECT_TIMEOUT_MS = 10_000;
+
+// Test seam: lets tests substitute a connect factory that never completes,
+// so the direct-path timeout can be asserted in milliseconds instead of
+// simulating a black-holed port (ESM namespaces are immutable — the repo's
+// established _...ForTest pattern instead of module mocking).
+let connectFactory: (port: number, host: string) => net.Socket = (port, host) => net.connect(port, host);
+
+export function _setConnectFactoryForTest(factory: ((port: number, host: string) => net.Socket) | undefined): void {
+    connectFactory = factory ?? ((port, host) => net.connect(port, host));
+}
+
+/** Plain (non-proxied) outbound TCP connect with a bounded handshake:
+ *  a black-holed upstream (firewall drop, routing hole) must fail in seconds,
+ *  not after the OS default ~127s (tcp_syn_retries). Mirrors the proxied
+ *  branch's 10s handshake timeout (see #78). */
+export function connectDirect(host: string, port: number, timeoutMs: number = CONNECT_TIMEOUT_MS): Promise<net.Socket> {
+    return new Promise((resolve, reject) => {
+        const socket = connectFactory(port, host);
+        let settled = false;
+        const finishError = (error: Error) => {
+            if (settled) return;
+            settled = true;
+            clearTimeout(timer);
+            socket.destroy();
+            reject(error);
+        };
+        const timer = setTimeout(() => finishError(new Error(`upstream connect ${host}:${port} timed out after ${timeoutMs}ms`)), timeoutMs);
+        socket.once("error", finishError);
+        socket.once("connect", () => {
+            if (settled) return;
+            settled = true;
+            clearTimeout(timer);
+            resolve(socket);
+        });
+    });
+}
+
 export function connectThroughProxy(host: string, port: number, proxyUrl: string | undefined): Promise<net.Socket> {
     if (!proxyUrl) {
-        return new Promise((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


ranxianglei:
LGTM — clean, well-scoped, and the test suite is thorough (4 new tests: real `net` server, delegation equivalence, timeout fire, and fast-fail-with-throwing-socket). All 457 pass locally.

The core fix is sound: `connectDirect` now has a real 10s handshake deadline (shared `CONNECT_TIMEOUT_MS`, so the proxy path and direct path now use the same constant), and the `settled` guard correctly prevents a double-resolve/reject race between the timeout and the real `connect`/`error` events. The `socket.destroy()` on timeout is important — without it a half-open SYN would leave the socket dangling.

Two non-blocking notes:

- After a *successful* connect, the `socket.once("error", finishError)` listener stays attached. If the socket errors later (post-handshake), `finishError` will run but no-op because `settled === true` — fine, it's a no-op, but the listener is never removed. In practice the socket is handed to the caller and GC'd, so no leak; just noting the listener outlives the handshake by design.
- The `connectDirect` error path calls `socket.destroy()` on timeout, but the socket is also created inside the returned promise and handed to the caller on success — the ownership transfer is clean (caller owns it after `resolve`), so no double-destroy. Good.

One real question worth a follow-up (not blocking): the 10s is hardcoded via `CONNECT_TIMEOUT_MS`. Is there any case where a legitimately slow-but-alive direct connect (e.g. a flaky datacenter link) would exceed 10s and now fail where it previously hung or succeeded? If direct-connect is used for low-latency local/upstream paths, 10s is generous; if it's ever used for cross-region, consider making it configurable. For the current use (fix #78 connect-timeout hang) 10s is a good default.

Merge.

PR #173, opened 2026-08-18
feat: builtin thin agent plugins (pi/omp/claude/codex/opencode) + bili plugin install CLI

ranxianglei:
## What

Ships the per-agent builtin plugin layer discussed in the 内外呼应 track: a thin (~100 line) extension per host family that registers the proxy's ACP tools natively and forwards executes to the proxy. Zero acp-kernel in the plugin — the proxy stays the single engine, so no kernel-version skew is possible.

```
bili plugin install <pi|omp|claude|codex|opencode>
bili plugin list
bili plugin remove <agent>
```

## Files

- `src/agent/shared.ts` — proxy discovery (/bili/ URL prefix, BILLION_CONTEXT_PROXY env, BILLION_CONTEXT_PLUGIN=0 kill switch) + manifest/tool-forward/status client
- `src/agent/pi.ts` / `src/agent/omp.ts` — pi & omp ExtensionAPI factory (same API shape); stamps x-bili-plugin/-conversation/-context-window headers on provider requests; registers manifest tools at session_start with retry throttle; ToolResult content-block contract per pi-stable types
- `src/plugin-install.ts` — per-host installers (pi settings.json packages, omp config.yml extensions, codex config.toml [mcp_servers.bili] with env-stamped proxy origin, claude mcp add -e, opencode mcp.bili environment); all idempotent with .bili-bak backup
- `src/mcp.ts` — lazy manifest fetch on first tools/list (no more fetch+exit at startup), 5s timeout
- `src/cli.ts`, `tsup.config.ts` (dist/agent/pi.js + omp.js entries), `package.json` pi manifest field
- `tests/plugin-agent.test.ts` — 11 tests: discovery variants, hostile host shapes, throttle dedupe, install/remove roundtrips for all 5 hosts under fake HOMEs (broken-config refusal, flow-style YAML, TOML table surgery, stale-origin refresh)

## Live verification (this machine, isolated HOMEs)

| host | result |
|---|---|
| pi | tmux session, real compress via plugin path, fold 74→33 msgs, agent stayed alive |
| omp | ext loads (CJS/ESM interop factory), registers tools, manifest hit, replies via proxy |
| claude | claude mcp add user scope, ✔ Connected |
| codex | mcp__bili__* tools listed, acp_status tools/call round-trip through proxy (rollout jsonl evidence) |
| opencode | full initialize → tools/list → tools/call ok:true round-trip |
| dsh | not installed on this machine — noted; MCP path would cover it |

## Notable bugs found during live testing (all fixed here)

1. **codex sanitizes MCP child env** → BILI_MCP_PROXY never reached dist/mcp.js → fell back to dead default origin → eager manifest fetch exited(1) before stdin attached → "connection closed: initialize response" and no bili tools all session. Fix: lazy manifest + env-stamped origin in the TOML at install time (reinstall refreshes a stale origin).
2. pi ToolResult contract is `{content:[{type:"text",text}],isError}` — a plain `{output}` shape crashed the agent mid-session.
3. jiti CJS/ESM interop: factory must also be reachable as `module.exports.default` or pi/omp silently skip the extension.
4. CODEX_HOME replaces the whole .codex dir (config at $CODEX_HOME/config.toml); omp honors PI_CODING_AGENT_DIR, default ~/.omp/agent (no OMP_CONFIG).

Full test suite 464/464, typecheck clean, build ok.


ranxianglei:
The design here is the strongest of the open PRs — one shared protocol client (`src/agent/shared.ts`, zero host imports) serving the pi/omp plugin *and* the MCP bridge, lazy manifest fetch that fixes the codex env-sanitization crash, per-request origin resolution (env > origin-file > default) that self-heals after a proxy restart, and idempotent installers with one-time `.bili-bak` backups that refuse to clobber unparseable configs. 409 lines of tests, 465/466 of the full suite pass locally. Two things to fix before I'd merge:

**1. The claude test is environment-dependent and fails on any machine with the `claude` CLI installed.**
`tests/plugin-agent.test.ts:337` expects `pluginInstall("claude")` to throw `/claude: install failed/`, sandboxed via `env: { ..., CLAUDE: "definitely-not-a-real-cli" }`. But `claudeInstall()` (src/plugin-install.ts:209) hardcodes `execFileSync("claude", ...)` — the `CLAUDE` value is only forwarded into the *child's* env and is never used as the binary path, so the sandbox is a no-op. Result:
- CI (no `claude` on PATH): passes *by accident* — ENOENT produces exactly the expected error.
- Developer machines (claude installed, like mine): `claude mcp add` succeeds → `assert.throws` fails with "Missing expected exception". I reproduced this: the test suite goes red on 13/14 files.

Suggested fix: honor the env var as the command — `execFileSync(env.CLAUDE ?? "claude", ...)` in both `claudeInstall` and `claudeRemove` (the test's `CLAUDE` fixture already assumes exactly this contract), or make the test skip when `claude` is present. Note the test title says "pi/omp/codex/opencode" but the claude assertions are inside it — worth mentioning in the title too.

**2. `--origin` is silently ignored by `bili plugin install`.**
`parseArgs` puts it in `overrides.BILI_MCP_PROXY`, but the `plugin` branch of `main()` (src/cli.ts:315-340) returns before the overrides→`process.env` merge at cli.ts:369, and `resolveProxyOrigin()` (src/mcp.ts:37) only reads `process.env` / the origin file. I verified: with a stale `~/.bili/proxy-origin` file present, `bili plugin install codex --origin http://127.0.0.1:9999` stamped `BILI_MCP_PROXY = "http://127.0.0.1:46657"` (the stale file value) instead of 9999. An explicit flag being overridden by a stale file is a trap. Two-line fix in the `plugin` branch before dispatch:
```ts
if (overrides.BILI_MCP_PROXY) process.env.BILI_MCP_PROXY = overrides.BILI_MCP_PROXY;
```
(same for `--agent` if you want it meaningful there too).

Minor nits (non-blocking):
- `retryTimer` is declared in `RegisterState` (src/agent/pi.ts:68) but never assigned or read — dead field; remove it, or wire up timer-driven retry if that was the intent (today retry only happens on the next host event).
- Convention inconsistency: `console.error` in pi.ts (5 sites) vs `process.stderr.write` in plugin-install.ts:71. Your own commit note says console.error is reserved for cli.ts/index.ts — pick one for agent-bundled code (I'd take stderr.write everywhere).
- `claudeRemove` throws "claude: remove failed" when the `claude` CLI is absent even if bili was never installed — check `claudeStatus()` first and report "not installed" in that case.
- `fetchJson` (src/agent/shared.ts:56): if `externalSignal` is already aborted, the listener never fires and the request runs until its internal timeout. Bounded and harmless, but `if (externalSignal?.aborted) ctrl.abort()` up front makes it exact.

ranxianglei:
Addressed both blockers and the nits in 30fbefb:

**Blockers**
1. `claudeInstall`/`claudeRemove` now honor `CLAUDE` as the binary path (absolute path for sandboxed setups). Tests inject `CLAUDE=/nonexistent/bili-claude-stub` so the `install failed` assertion is deterministic on machines that *do* have the real CLI (verified locally: suite failed pre-fix, 467/467 post-fix).
2. `--origin` is now applied to the env before the `plugin` subcommand dispatch — previously the branch returned ahead of the generic overrides→env merge, so a stale `~/.local/state/billion-context/proxy-origin` discovery file silently won. End-to-end verified: `plugin install codex --origin http://127.0.0.1:9999` now stamps `9999` instead of the stale `46657`.

**Nits**: dead `retryTimer` field dropped; empty session ids cached (no per-request manifest refetch); `fetchJson` forwards an already-aborted signal instead of waiting out the timeout; `claude` remove is a no-op when not installed; `proxy-origin` discovery file respects `--host` (wildcard/localhost→127.0.0.1, bare IPv6 bracketed).

Local: 467/467 tests + typecheck clean.

PR #176, opened 2026-08-19
fix(compress): accept JSON-string content in parseCompressInput

ranxianglei:
## 背景

非严格工具 provider(vLLM openai-completions,`supportsStrictTools:false`)会把嵌套数组参数字符串化:模型调用 `compress` 时传的 `content` 是 **JSON 字符串**而非数组。

此前 `parseCompressInput` 里 `Array.isArray(obj.content)` 为 false → 走 `toRange(obj)`(顶层无 startId/endId/summary)→ null → **静默解析出 0 个 range**,压缩从不执行(只有一条 warn 日志,无任何报错)。

## 改动

`src/compress-tool.ts` 的 `parseCompressInput` 在数组判断前,若 `content` 是字符串则先 `JSON.parse`(非法 JSON → warn + 0 ranges,不崩溃)。与 billion-context-pi PR #164(`normalizeRanges`)对称。

`tests/basic.test.ts` 增加 JSON 字符串形式的回归测试。

## 验证

- `npm run typecheck`(tsc --noEmit --project tsconfig.build.json):clean
- `npm test`:**464 pass / 0 fail**

不合并,等 review。

### DIFF
--- a/src/compress-tool.ts
+++ b/src/compress-tool.ts
@@ -57,9 +57,19 @@ export function parseCompressInput(input: unknown, callId?: string): ParsedRange
         return [];
     }
     const obj = input as Record<string, unknown>;
+    // Accept JSON-string content (non-strict-tool providers stringify array args, e.g. vLLM openai-completions).
+    let content: unknown = obj.content;
+    if (typeof content === "string") {
+        try {
+            content = JSON.parse(content);
+        } catch {
+            loggerLog("warn", `[acp-compress-input] content is a string but not valid JSON; parsed 0 valid ranges`);
+            return [];
+        }
+    }
     const single = toRange(obj);
-    const ranges = Array.isArray(obj.content)
-        ? obj.content
+    const ranges = Array.isArray(content)
+        ? content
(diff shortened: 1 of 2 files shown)


ranxianglei:
LGTM — verified against all five call sites (`stream.ts`, `stream-openai.ts`, `stream-responses.ts`, `compress-loop-responses.ts`, `loop/core.ts`) and the new regression tests pass (25/25 in `tests/basic.test.ts`). Mirrors the sibling fix in billion-context-pi.

One non-blocking note: if `content` is a JSON string of a **single object** (e.g. `'{"startId":"m1","endId":"m2","summary":"x"}'`), `JSON.parse` succeeds but the result isn't an array, so the fallback reads `startId`/`endId`/`summary` from the **top-level** `obj` (line 70) — which only carries `content` — and you get 0 ranges. Same result as before the patch, so no regression; just flagging that the lenient path only covers array strings. If you want the single-object hallucination to work too, one line: after the parse, `if (content && typeof content === "object" && !Array.isArray(content)) content = [content];`. The schema declares `content` as an array, so I'd merge as-is.

PR #185, opened 2026-08-22
Strip consumed compress pairs from the upstream wire (kernel >=0.0.32 compat)

ranxianglei:
## Why

acp-kernel 0.0.32 (kernel #104/#105, `KEEP_LAST_ORPHANED` 0 → 2) keeps the newest orphaned compress call+result pairs visible in `processTurn` output. The proxy's upstream contract requires **consumed** compress calls to stay off the wire.

## What

`src/server.ts`: `stripKernelSummaries(messages, state)` now also drops compress call+result pairs that are neither active-block-linked (their args carry the live summary) nor failures (`FAILED` marker) — **rejections stay visible** to the model. Core tool-results carry no `toolName`, so calls and results are paired by `toolCallId` across all four strip sites (anthropic / openai / responses / count_tokens).

## Tests

- [x] `npm run typecheck`
- [x] tests 512/512 (incl. plugin-protocol "consumed range folds away")
- [x] No-op under kernel < 0.0.32 (the kernel already hid orphaned pairs there).

## Follow-up

A separate bump-only release PR (acp-kernel 0.0.32 → v0.1.45) will follow **after** this merges — no version changes in this PR.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -801,8 +801,34 @@ const ACP_TAG_MARK = "\x3cacp ";
 // generic-library fallback; this host strips it because the compress tool-call
 // already carries the summary (hideConsumedCompressCalls keeps active-block calls),
 // and a mid-stream insertion would shift the upstream prefix-cache breakpoint.
-function stripKernelSummaries(messages: BiliMessage[]): BiliMessage[] {
-    return messages.filter((m) => !(m.id ?? "").startsWith("acp_summary_"));
+//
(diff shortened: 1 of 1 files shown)


ranxianglei:
Closing: superseded by kernel-side stripping. `hideConsumedCompressCalls` (kernel, used at src/loop/core.ts:482) removes consumed compress pairs from the wire, and master's mode-aware `stripKernelSummaries` (server.ts) handles the summary carriers per compression mode — the host-side pair-dropping this PR added is no longer needed.

PR #188, opened 2026-08-22
feat: lenient compress-arg parsing via acp-kernel salvage (omp#121)

### LINKED ISSUE #603: lenient compress-arg parsing: salvage salvageable model output
### Problem
Models occasionally emit slightly malformed `compress` args (wrapped JSON, trailing prose, single quotes). `parseCompressArgs` rejects the whole call → the round is wasted, the model sees a FAILED result and may tag-echo.

Reported via omp#121.

### Impact
Compress rounds fail on input a human would trivially salvage; small models hit this repeatedly.

### Suggested fix
Salvage pass for the common malformation classes before hard rejection (host-side in compress-tool.ts, or kernel-side in parseCompressArgs as an opt-in lenient mode). PR #188 carries a host-side implementation to use as reference; kernel currently exports no salvage API.

ranxianglei:
Closes #603 — tracking issue for this PR.

## Why

Weak/local models (vLLM qwen etc.) emit compress tool arguments that fail strict JSON.parse ~50% of the time — truncated output caps, raw newlines inside summary strings, trailing commas, or plain prose (see billion-context-omp#121). Until now the proxy's `parseCompressInput` was strict-only:

- raw **string** input (text-protocol triggers, stream tails) was rejected outright;
- JSON-string **content** that failed to parse returned `[]` — raw arguments discarded with **zero log evidence**, making the failure class undiagnosable in production.

## What

Route both paths through acp-kernel's `parseCompressInputString` → `salvageParseRanges` (5-layer ladder: strict JSON / fenced / repaired / truncated-array prefix / field-regex):

- string input is salvaged instead of rejected;
- broken JSON-string content falls back to the same ladder instead of `[]`;
- the lenient path logs the salvage layer + note, and keeps `raw[:800]` as evidence when 0 ranges are recovered.

Bumps acp-kernel to ^0.0.34 (introduces `salvageParseRanges` + `extractRanges` double-encode normalization).

## Tests

5 new lenient-path cases in `tests/basic.test.ts` (raw string salvage, fenced JSON, repaired commas/newlines, truncated-array prefix, field-regex prose form). Full suite 516/516, typecheck clean.

Companion PRs: acp-kernel#109 (parser), billion-context-omp (replay), billion-context-pi (ACP host extension).

--- DEPENDENCY NOTE (pre-publish) ---
> ⚠️ **Do not merge yet**: `package.json` currently pins `acp-kernel` to a local `file:` tarball so the diff is reviewable now. Depends on acp-kernel PR #109 landing and npm publish of **0.0.34** (0.0.33's publish failed with npm 404 — expired NPM_TOKEN). Before merge this PR will be updated to `"acp-kernel": "^0.0.34"` from the registry and the full suite re-run.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
       "devDependencies": {
         "@types/node": "^22.0.0",
         "@types/node-forge": "^1.3.14",
-        "acp-kernel": "0.0.32",
+        "acp-kernel": "file:../acp-kernel/acp-kernel-0.0.33.tgz",
         "fzstd": "0.1.1",
         "node-forge": "^1.4.0",
         "tar": "^7.5.22",
@@ -972,9 +972,9 @@
       }
     },
     "node_modules/acp-kernel": {
-      "version": "0.0.32",
-      "resolved": "https://registry.npmjs.org/acp-kernel/-/acp-kernel-0.0.32.tgz",
-      "integrity": "sha512-iOJPF+X6NMGkpGsAbxHV0Fcvymzf9a0c5Mf/z3padNVgPgMNxwG1snxYravVccRePgHOzWgpNYikqtpGYhiInA==",
+      "version": "0.0.33",
+      "resolved": "file:../acp-kernel/acp-kernel-0.0.33.tgz",
+      "integrity": "sha512-zTuOCINsAGhR58gVrniKyTeFdL77Jj1ZmMC91iUZLr0puFmuhAEpau/LT2yXYubU8/20owdijO0DFH1f4QkV8w==",
       "dev": true,
       "license": "MIT",
       "engines": {
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
   "devDependencies": {
     "@types/node": "^22.0.0",
     "@types/node-forge": "^1.3.14",
-    "acp-kernel": "0.0.32",
+    "acp-kernel": "file:../acp-kernel/acp-kernel-0.0.33.tgz",
     "fzstd": "0.1.1",
     "node-forge": "^1.4.0",
     "tar": "^7.5.22",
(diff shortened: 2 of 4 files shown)


ranxianglei:
**Review vs master 3cd85b6 (kernel 0.0.56):** ❌ **unbuildable as-is** — two blockers found.

1. **Kernel pin is a local file ref**: `package.json` has `"acp-kernel": "file:../acp-kernel/acp-kernel-0.0.33.tgz"` — a tarball that exists only on the author's machine. CI (`npm ci`) can never install it.
2. **`salvageParseRanges` does not exist in any published kernel**: verified against the currently installed 0.0.56 — `typeof k.salvageParseRanges === "undefined"`. So the import cannot resolve even if the pin were fixed. (Same class as #213's `0.0.45-pr.160.14` pin.)

The host-side shape is otherwise sensible: string-input path routes through the lenient ladder; object-with-string-content path falls back to `parseCompressInputString(JSON.stringify(input))` — note that double-encoding trick assumes the ladder re-parses embedded strings, which needs re-verification against whatever ladder implementation lands.

Rework direction (tracked in #603): either (a) land the salvage ladder in acp-kernel as a published export and re-pin exactly, or (b) implement the ladder host-side in `src/compress-tool.ts` (strip-code-fence → extract-first-JSON-object → JSONL scan → range-regex). The existing `parseCompressInput` diagnostics (#368) should keep their evidence log lines.

(tracking: #603)

ranxianglei:
Closing — superseded by **#610** (merged in v0.1.90), which ships the salvage path on master's parser ladder (`src/compress-tool.ts` `parseCompressInput` quote-salvage fallback, closes #603). This PR additionally pins a local kernel tarball (`file:../acp-kernel-0.0.33.tgz`) and its `salvageParseRanges` helper was never published to the kernel — not buildable on CI. The covered behavior (omp#121-class single-quoted JSON salvage) is now live via #610.

PR #193, opened 2026-08-22
fix: re-inject the client's openai system around the kernel hoist (0.0.37)

ranxianglei:
## Symptom

The kernel (PR ranxianglei/acp-kernel#120, released 0.0.37) hoists the contiguous leading system/developer prefix OUT of the openai fold space: system content is host runtime state and must not feed ids/fingerprints, and a compression covering it used to delete the model's system prompt from every rebuilt payload.

Consequence for the proxy: after upgrading the kernel alone, `openaiToCore`'s msgs no longer carry the system at all — without this PR every forwarded payload and every compress-loop round would silently drop the client's system prompt.

## Fix

- `prepareOpenai` captures `systemText` from `openaiToCore` and re-injects it into every rebuilt payload (client system first, compress prompt after — preserves the prefix-cache-anchor ordering)
- `Prepared.openaiSystemText` flows into `pickAdapter` → `createOpenaiAdapter(requestBody, clientSystem)` so compress-loop rounds re-inject it too, mirroring the anthropic adapter's `anthropicSystem` path
- loop adapter `buildRequest` injects `[clientSystem, compressPrompt]` (empty parts filtered)

Note the proxy never had the restart-fingerprint failure (MITM sees the real wire every round) but DID inherit the compression-eats-system bug — and would lose system entirely on 0.0.37 without this change.

## Tests

- 512 pass / 0 fail; round-trip tests updated to the hoist contract (leading prefix hoisted + mid-conversation system/developer still round-trips via originalRole)
- acp-kernel 0.0.37 from npm

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -972,9 +972,9 @@
       }
     },
     "node_modules/acp-kernel": {
-      "version": "0.0.37",
-      "resolved": "https://registry.npmjs.org/acp-kernel/-/acp-kernel-0.0.37.tgz",
-      "integrity": "sha512-H0CfGFgAnbtwYiXODPitctjbKWoljVcF3UYZrWDD5qTWHYwkiIZyw0wcH5UTorNC0FfCiKjQM6KMAwetAbAyqA==",
+      "version": "0.0.41",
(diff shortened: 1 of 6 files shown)


ranxianglei:
防止模型擅自合并,先关闭

PR #196, opened 2026-08-22
refactor(persist): delegate SessionStore mechanism to acp-kernel/persist

ranxianglei:
Protocol-consolidation Phase P (user direction m00509: billion-context 的持久化机制进入内核，下游复用).

## What
- `SessionStore` becomes a thin POLICY adapter over `StateStore<PersistedSession>` (`acp-kernel/persist`, shipped 0.0.40/#126 + 0.0.41/#130): ~270 lines of duplicated mechanism deleted (atomic write, rename retry, debounce timers, write chains, corrupt-tolerant walk).
- Kernel gains nothing proxy-specific; proxy keeps: record schema (`PersistedSession` v3), namespaced layout (`relPathFor`), validity, build/buildSession, env knobs (`BILI_PERSIST*`), singleton. Public API unchanged (`session.ts`/`server.ts`/`export.ts` untouched).

## On-disk migration
- New files: envelope `{version, savedAt, id, payload}`. Old flat files (all versions, incl. v1) load via the kernel's `legacy` adoption hook and re-persist as envelope on the next dirty write — old sessions never break, files migrate organically.

## Semantics upgrades (documented in tests)
- `scheduleSave` builds the record at WRITE time (freshest state under coalescing) instead of capturing at call time.
- In-process `loadSync` with stale meta still resolves via the kernel's discovered-file cache — id is authoritative, meta is only a path hint (isolation probe uses a cold store).

## Deps
- `acp-kernel` 0.0.41 (carries the #193 hoist + persist subpath; base is #193 — merge that first).

## Tests
- 513/513 green (512 baseline + 1: pre-envelope flat-file adoption → re-persist envelope round-trip). `tsc` build-scope clean. Fixes nothing user-visible on its own; unlocks single-sourced persistence across proxy + omp (billion-context-omp#131).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
       "devDependencies": {
         "@types/node": "^22.0.0",
         "@types/node-forge": "^1.3.14",
-        "acp-kernel": "0.0.37",
+        "acp-kernel": "0.0.41",
         "fzstd": "0.1.1",
         "node-forge": "^1.4.0",
         "tar": "^7.5.22",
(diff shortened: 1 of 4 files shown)


ranxianglei:
防止模型擅自合并,先关闭

PR #219, opened 2026-08-24
fix(config): correct stale context limits + registry-first resolution + Web UI parse-error guard

ranxianglei:

Fixes the three issues reported by stirp in #212.

## What

**1. Stale built-in context limits** (`src/config.ts`)
- `deepseek`: 64_000 → 128_000 (the old value predates V3; models.dev shows deepseek-chat = 1M, v3.2 = 128K, v3.1 = 131K; 128K is the conservative safe floor — under-reporting is safe, the 400-overflow self-learning converges upward)
- `minimax`: no entry existed at all → new entry 204_800 (M2 = 196,608, M2.1 = 204,800, M3 = 512,000 per models.dev)

**2. Registry was unreachable behind a stale table** (`src/registry.ts` + `src/server.ts`)
- Problem: the built-in table was checked BEFORE models.dev, and a hit short-circuited the registry lookup. A stale table entry permanently locked out fresh data.
- Fix — priority inversion for the fast path:
  ```
  plugin header ?? warm-registry-cache (peek, sync, never fetches) ?? built-in table
  ```
  then (only if still unknown + host known): `await contextFromRegistry()` which may fetch.
- `peekRegistryContext(model, host)` — new sync read-only cache lookup; on a cold cache it misses immediately instead of blocking startup.
- `HOST_TO_PROVIDER` += `api.minimax.chat` / `api.minimaxi.com` / `api.minimax.io` (registry keys are provider-qualified, e.g. `minimax/MiniMax-M2.1` — without the host mapping the bare model name never matched).
- Cold-start behavior unchanged: sync miss → built-in table → async fetch warms the cache for the next request.

**3. Web UI silently wipes unreadable configs** (`src/web/api.ts` + `src/web/client.ts`)
- Root cause of the "npm upgrade wiped my config" report: a hand-edited config with a comment / trailing comma fails `JSON.parse` → `safeReadJson` → null → `readConfig() = {}` → the UI shows an empty form → the user re-fills and saves → the server rebuilds from `{}` and atomically overwrites the file. Every unread field is gone. npm itself never writes user config (the installer only creates it when missing).
- Fix:
  - `GET /__bili/api/config` now returns `parseError` when the file exists but doesn't parse; the UI toasts a warning instead of presenting an empty form as truth.
  - `PUT` returns **409** and refuses to write while the file is unparseable — merge-into-{} can no longer silently drop fields. The user fixes the syntax by hand (the GET response carries the path), then PUT works again.

## Tests (+8)

- `tests/registry.test.ts` (new): minimax host mapping, boundary-safe non-matches, cold-cache miss, warm-cache provider-qualified + bare-name fallback, async path, unusable entries skipped, reset.
- `tests/routing.test.ts`: deepseek 128K ×2, MiniMax-M2.1 / minimax-m2 204,800.
- `tests/web-routing.test.ts`: broken file → GET parseError + PUT 409 + file untouched; after fixing syntax → PUT 200 and the surviving fields make it through.

typecheck ✅ · 542/542 ✅ · build ✅ · devlog/`2026-08-24_context-window-fixes`


ranxianglei:

Follow-up commit `6a9614b` — the user's hunch in this thread was correct and is now fixed:

**Registry fetch was permanently dead on proxied networks.** Node's global `fetch` ignores `http(s)_proxy` env vars, so `fetchFresh()` always connected directly — and on networks where models.dev is unreachable directly (verified on this machine: direct = timeout, shell proxy = 200 OK) the registry never loaded, never cached, and the stale built-in table was the **only** live data source. That's exactly why the stale DeepSeek 64K mattered so much in practice.

Fix: `fetchFresh()` now routes through the configured upstream proxy when one exists (`proxyDispatcher()` — the same cached undici ProxyAgent infrastructure the model traffic uses), falling back to a direct attempt when no proxy is set or the proxied attempt fails.

Real-machine verification:
```
[acp-registry] loaded models.dev (355 models, via proxy)
~/.cache/billion-context/models-dev.json  (288 KB, 24h TTL)
  deepseek-chat  → 1,000,000   (was locked at 64K)
  MiniMax-M2.1   →   204,800   (was falling to default 200K)
```
typecheck ✅ · 542/542 ✅

ranxianglei:

Follow-up commit `605a917` — bundles a slim models.dev snapshot into the package as the **offline floor** (answering the thread question: previously the 355-model data existed ONLY in the runtime disk cache; a fresh install behind a dead network got just the 17-entry regex table).

**How it works:**
- `npm run registry:snapshot` (`scripts/update-registry-snapshot.mjs`, proxy-aware, keeps the old file when offline) → commits `src/registry-snapshot.json` — 351 models, **12 KB**, inlined into dist by tsup.
- The snapshot pre-warms the sync `peekRegistryContext` path at module load, so the **first request** already resolves exact windows with zero network. Fresh disk cache and live fetch still outrank it (identity check `cache !== snapshotReg`).
- When a fetch fails: newer of (stale disk cache by mtime, bundled snapshot by `fetchedAt`) wins — a new release's snapshot beats a months-old offline cache.

**Full fallback chain:** memory → fresh disk (24h) → live fetch (proxy → direct) → newer(stale disk, snapshot) → 17-entry regex table → 200k default.

**Offline verification** (no proxy env, disk cache deleted, fresh process):
```
peek deepseek-chat@api.deepseek.com → 1,000,000   (was: undefined → stale 64k table)
peek MiniMax-M2.1@api.minimax.chat  →   204,800
unknown model                        →   undefined  (no guessing)
```
typecheck ✅ · 544/544 ✅ · build ✅ (dist 2.27 MB)

ranxianglei:
**Update: the bundled snapshot now carries the ENTIRE models.dev registry, hard-coded into the package.**

Previously the snapshot was a projection (context windows only, 12KB). Per review it is now the full dataset:

| | before | after |
|---|---|---|
| Size | 12 KB | **282 KB** |
| Models | 351 | **355** |
| Fields | limit.context only | **everything**: name, description, family, attachment, reasoning, tool_call, temperature, knowledge, release_date, last_updated, modalities, open_weights, limit.context/output/input, weights, benchmarks, ... |
| dist/index.js | 2.27 MB | 2.55 MB |

- `scripts/update-registry-snapshot.mjs` writes `{fetchedAt, count, models: <entire models.dev>}`
- `bundledSnapshotRegistry()` consumes it as-is — full entries are structurally valid; only `limit.context` is read today, the rest rides along so future features (pricing, modality checks, reasoning/tool_call flags) get the offline floor for free
- New regression test guards against re-slimming (≥350 entries, ≥300 with family / limit.output / reasoning+tool_call)
- Offline floor re-verified (no proxy, no disk cache, fresh process): deepseek-chat → 1,000,000 · MiniMax-M2.1 → 204,800 · unknown → undefined

Refresh before each release: `npm run registry:snapshot`

PR #227, opened 2026-08-24
fix: bili plugin install pi replaces legacy billion-context-pi entries

ranxianglei:
Model: qwen3.8-27b (vllm)

## Problem

`bili plugin install pi` did not recognize legacy `packages` entries in `~/.pi/agent/settings.json`:

- `npm:billion-context-pi` (+`@version`) — the old standalone 0.1.x package. Windows test env and older setups still carry this entry; install stacked the new plugin next to it → duplicate tool registrations.
- dev checkout paths (`…/billion-context` not under node_modules).
- Windows backslash paths (old regex only matched `/`).

## Fix

- `isPiEntry` now matches all four shapes (current root, npm forms incl. legacy `-pi`, node_modules paths with either separator, bare path endings) — install **replaces** every match so exactly one bili plugin stays live.
- `piStatus` reports strictly on the current install root (a stale legacy entry no longer shows as "installed").

## Tests

- Extended roundtrip test: pre-seeds 6 entries (legacy npm forms, dev path, Windows backslash path, unrelated package), asserts install leaves `[unrelated, root]` and remove leaves `[unrelated]`.
- Full suite: 565/565 pass.

## Validation

- **Linux**: remove+install against real `~/.pi` cleans legacy entries; `bili pi` launcher → plugin loads, manifest fetch ok, round-2 request stamps `x-bili-plugin: pi` (+conversation id, context-window 1000000) — native takeover confirmed in proxy log.
- **Windows 10 VM** (Node 22.23.2, pi 0.84.1, GLM): registry upgrade 0.1.39→0.1.50 ✅; tgz install of this fix ✅; `bili plugin install pi` replaced legacy entry with the single global path (auto-backup created) ✅; `bili pi` cert-MITM e2e with native takeover headers ✅; `bili start` (listen/persist/state-log/auto-update/web UI/health) ✅.

### DIFF
--- a/devlog/2026-08-24_pi-install-windows/REQ.md
+++ b/devlog/2026-08-24_pi-install-windows/REQ.md
@@ -1,0 +1,67 @@
+# pi plugin install 打通 + Windows 全链路验证
+
+## 需求
+
+用户（2026-08-24）：
+
+> 现在你把pi 需要install的搞定 一切方法搞定.然后需要在windows测试.远程有windows的
+> test环境. 确保bili 命令是支持的工作的
+
+拆解：
+1. `bili plugin install pi` 端到端可用，pi 的所有接入方法（install 持久注册、launcher、
+   wire 兜底）都工作。
+2. 在远程 Windows 10 测试机（win10-vm，见 ~/system/win10-vm.md）验证 `bili` 命令可用。
+
+## 修复：install 不清理旧 billion-context-pi 条目
+
+- 现象：`isPiEntry`（src/plugin-install.ts）只认 `npm:billion-context`、
+  `^npm:billion-context@`、`node_modules/billion-context` 三种形态，不认：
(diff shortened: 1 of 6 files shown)


ranxianglei:
Model: qwen3.8-27b (vllm)

Added a second fix on top: `bili pi` now injects the native plugin via pi's `-e` flag when the user has NOT run `plugin install pi` (mutually exclusive — pi dedups extensions by resolved path, packages-entry vs -e-file would double-register).

Validated end-to-end:
- Linux uninstalled: first real tool request carries `x-bili-plugin: pi` + context-window 1000000
- Windows VM (win10-vm, Node 22): tgz install → remove pi → `bili pi` e2e EXITCODE=0, content read verbatim, upstream headers show native takeover
- 516/516 tests, typecheck, build green

PR #229, opened 2026-08-24
fix: arm emergency shrink on upstream failure to break relay-5xx deadlock (#11)

### LINKED ISSUE #604: arm emergency shrink on upstream failure to break relay-5xx deadlock
### Problem
When the session is over budget and the upstream is a relay that answers overflow with a hard 5xx (instead of an in-band error), the request fails; recovery relies on the client retrying into a proxy that nudges. A client that just surfaces the error leaves a stuck, never-compressed session.

Partially mitigated since: stream retry (#440/#444), preflight hold (#571), learned-window self-correction (#570/#572) — but the "on upstream failure, force compression before the next attempt" mechanism itself is still absent from master.

### Impact
Dead sessions on 5xx-on-overflow relays until manual intervention.

### Suggested fix
On upstream failure with an over-budget session, arm the emergency compression path (truncate/emergency threshold already exist in compress-settings) so the next client retry lands on a foldable view. PR #229 carries a v0.1.7x-era implementation + e2e (tests/e2e-relay-5xx-selfheal.test.ts) as reference; needs rework against current server.ts.

ranxianglei:
Closes #604 — tracking issue for this PR.

## Problem (#11)

zcode user on a justwoker relay channel: every request for claude-opus-5 returns 500 after installing bili. Log shows:

- `{"error":{"type":"new_api_error","message":"upstream error: do request failed"}}` — the **relay's own** error (relay → its upstream failed), fixed ~31s timeout, plus one Cloudflare 522
- The session was huge: 676 messages, ~250K tokens vs a 200K window
- Other channels in the same log (openrouter ox-alpha, opencode.ai) worked fine, EMERGENCY compression fired there

**Deadlock:** compression triggering depends on the upstream usage report (`lastInputTokens`). A failed request never reports usage → the proxy re-sends the same ~250K payload on every retry → the relay fails it again → forever. The session can never recover through bili.

## Fix

Arm the emergency shrink on failures with a **local estimate** of the body actually sent:

- `armFailureShrink()` (src/server.ts) — `estimateTokensFast(body)` raises `session.stats.lastInputTokens` if larger, then `markDirty` (error paths return before forward()'s trailing save). Deliberate exception to the "tokenCount must be real usage" invariant: it only RAISES the value (lower bound → compress earlier, never later), the kernel no-ops below truncate.threshold, and the next successful usage report overwrites it.
- Fires on: upstream **5xx** responses (generic relay errors carry no overflow signature, so this is the only self-heal path) and **network-level** fetch failures (socket reset, timeout abort) for non-GET/HEAD.
- Excluded: 4xx (auth/quota errors must not distort the usage signal); overflow 400s keep the existing learned-window path.

Next turn, `processTurn` sees usage ≥100% → the kernel's emergency nudge + `emergency-truncate` node truncate large old tool results **server-side** (no model cooperation — critical, because the model never responds in this state). The resent payload shrinks below the relay's tolerance and the session recovers.

## Tests

`tests/e2e-relay-5xx-selfheal.test.ts` (2 new e2e tests, mirroring the overflow-selfheal harness):

1. **500 → arm → truncate → recover**: oversized history → relay 500 (passed through verbatim) → session armed with local estimate → next request's forwarded payload contains `[truncated for context space]`, shrank, upstream 200 → real usage report overwrites the armed value
2. **401 does NOT arm**: auth errors leave `lastInputTokens` untouched

Full suite: 567 tests pass, typecheck + build clean.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -1332,6 +1332,33 @@ function injectResponsesTool(tools: unknown[] | undefined, toolsToAdd: readonly 
     return [...tools, ...additions];
 }
 
+// #11: a failed request (relay 5xx, socket reset, timeout) never reports
+// usage, so lastInputTokens stays frozen and each retry re-sends the same
+// oversized payload — the relay fails it again (deadlock). Arm the emergency
+// shrink with a local estimate of the body just sent. Deliberate exception
+// to "tokenCount must be real usage" above: it only RAISES the value (a
+// lower bound can only compress earlier, never later), the kernel no-ops
+// below truncate.threshold, and the next successful usage report overwrites
+// it. markDirty is needed because error paths return before forward()'s
+// trailing save.
+function armFailureShrink(
+    prepared: Prepared | null,
+    body: Buffer | string | undefined,
+    log: (level: string, msg: string) => void,
+    reason: string,
+): void {
+    const s = prepared?.session;
+    if (!s || body === undefined) return;
+    try {
+        const est = estimateTokensFast(typeof body === "string" ? body : body.toString("utf8"));
+        if (est > s.stats.lastInputTokens) {
+            s.stats.lastInputTokens = est;
+            markDirty(s);
+            log("warn", `[${s.id}] ${reason} with no usage report — armed emergency shrink with local estimate ${est} tokens`);
+        }
+    } catch { /* non-text body — nothing to estimate */ }
+}
+
 async function forward(
     req: http.IncomingMessage,
     res: http.ServerResponse,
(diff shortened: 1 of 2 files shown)


ranxianglei:
**Review vs master 3cd85b6:** ⚠️ concept still valid — **needs rebase + one interaction check** (conflict in `src/server.ts`).

What holds:
- `estimateTokensFast` is still a kernel export and imported on master (`src/server.ts:4`) ✓ — the helper compiles after conflict resolution.
- Only-raises semantics (`if (est > s.stats.lastInputTokens)`) is the right shape: a lower-bound estimate can compress earlier but never later, the kernel no-ops below threshold, and the next real usage report overwrites.
- The e2e test is thorough (relay-500 → armed estimate → next turn carries `[truncated for context space]` → recovery usage overwrites).

What changed on master since (rebase must account for):
1. The 5xx branch neighborhood in `forward()` moved/grew (#572 weak-overflow learning, #576 preflight wire-overhead trigger) — the insert point needs fresh context.
2. **Interaction check:** master's #546 `restoreOutputBudget` keys side-request detection off persisted budget watermarks, and nudge usage percentages read `lastInputTokens`. Temporarily inflating `lastInputTokens` with a wire-body estimate (JSON-escaped, so it overestimates) can misclassify the next turn's usage ratio — worth asserting in the rebase that the recovery turn's side-request classification is unaffected (title-gen path has no `tools`, which mitigates the worst case).
3. Preflight (#576) now folds oversized payloads **before** send — the deadlock window this PR targets shrank but did not close (relay capacity 5xx ≠ overflow signature).

Not mergeable as-is; keep + rework per #604.

(tracking: #604)

ranxianglei:
Closing — superseded by **#611** (merged in v0.1.90), which arms `armFailureShrink` on both the 5xx branch and the network-failure catch branch with only-raises semantics and restart persistence (closes #604). This PR's core idea (raise lastInputTokens on relay failure → emergency band on next turn) shipped there; the surrounding code has since moved under #572/#576.

PR #254, opened 2026-08-26
fix: preflight-compress oversized context before forward on model switch

ranxianglei:
## Problem (#247)

Switching from a large-window model (e.g. 1M) to a smaller one (e.g. 260k) mid-session makes every subsequent request fail with an upstream "context full" error, permanently stuck:

- The session grew to N tokens under the 1M model (`lastInputTokens = N`).
- After the switch, `N > new window`, so the forwarded payload is oversized.
- The only existing mitigations are reactive/insufficient: the kernel nudge is a trailing user message that asks the *model* to call `compress` (it does not shrink the payload — and the model never sees it because the request is rejected), and emergency truncation only touches tool-result messages.
- The overflow self-heal learns the smaller window and arms the next turn, but nothing ever shrinks the payload → every request overflows identically.

## Fix

**`src/preflight.ts` (new)** — proactive pre-forward compression in `handle()`, before `forward()`:

- Trigger: `lastInputTokens >= modelContextLimit` (after self-heal + output-headroom reserve) and viable compressible ranges exist.
- Each round: re-run `processTurn` (refs renumber after each fold), take the oldest viable range, split it into chunks that fit the small window (≤ 60% of limit), render the segment as plain text, ask the upstream model for a tier-1 summary via a non-streaming request (per-protocol payload: anthropic/openai/responses), then fold it in with the existing `applyRanges` → `core.applyCompression` path (credits `lastInputTokens` correctly).
- Stops when the context fits, ranges are exhausted, or 8 rounds elapse. If it cannot fit the context, the request is forwarded anyway and the existing overflow self-heal takes over.

**`src/server.ts`**:

- `buildForwardTarget()` extracted from `forward()` (URL/headers/proxy resolution) so preflight can reuse the exact same upstream routing.
- `preflightCompressIfNeeded()` runs inside the session lock; on success it re-runs `prepare*` once so the forwarded payload is the folded one.
- `stripKernelSummaries()` is now state-aware: the kernel's in-place `acp_summary_*` anchor is stripped ONLY when redundant (the block's compress tool-call also carries the summary). Blocks created without a tool call (preflight) have no other carrier — their anchor is the only place the summary reaches the model, so it must survive. This also fixes the same loss for any tool-call-less block on the normal path.

## Tests

- `tests/preflight-compress.test.ts` (new): e2e — session grows to 300k under a 400k model, switch to a 260k model, assert preflight summarization calls happen, the folded payload fits (compressed markers gone, protected tail + summary present), and usage credits are correct. Second test: no preflight when the context fits.
- `tests/count-tokens.test.ts`: updated the in-place-summary assertion to the sole-carrier rule and added a case locking the redundant-anchor strip (tool-call present → anchor stripped, tool-call kept).

Full suite: 631/631 pass; typecheck + build clean.


ranxianglei:
[bot] Reviewing PR #254 now. I've fetched `2026-08-26_preflight-overflow-compress` (1a9f23c) — 4 files, +720/−48. I'll read the full diff, run typecheck + tests + build, and report findings here.

ranxianglei:
[bot] Review of PR #254 (`1a9f23c`, branch `2026-08-26_preflight-overflow-compress`) — done.

**Verification (all run on the PR branch):**
- `npm run typecheck` — clean
- `npm test` — 631/631 pass
- `npm run build` — clean (dist self-contained, acp-kernel inlined)
- Single commit, `fix:` prefix, branch naming per AGENTS.md; `package.json`/lockfile untouched (no version bump, no kernel bump)
- No `as any` / `@ts-ignore` / `console.error` in the new code; preflight logs via the injected `log`

**Design checks I traced through the code:**
- `buildForwardTarget()` extraction is verbatim from `forward()` (URL rewrite, hop-header strip incl. `content-length`, `x-codex-beta-features` filter, affinity, proxy resolution) — no routing behavior change.
- Trigger uses `reqConfig.modelContextLimit` *after* overflow self-heal (`src/server.ts:816-818`) and output-headroom reservation (`src/server.ts:839-840`), so the post-switch request that would previously overflow now preflights instead of dead-looping. This closes the #247 stuck state: self-heal learns the window, preflight shrinks the payload on the next turn.
- `session.stats.requests -= 1` after the rebuild is correct — each `prepare*` increments it (`src/server.ts:957/1040/1148/1325`) and the rebuild is internal to one client request.
- `compressCreditTokens` zeroing on rebuild (`src/server.ts:984`) is correct here: the rebuilt payload is already post-fold, so this turn's usage report is post-fold reality — keeping the credit would double-count.
- Per-chunk `refMaps` re-read (`src/preflight.ts:225`) handles the kernel's ref renumbering after each fold; raw-array indices are stable across folds, so chunks stay valid. The `minCompressRange` pre-check (`src/preflight.ts:229`) matches the kernel's batch gate (sum of resolved message text chars, `acp-kernel` applyCompression).
- `processTurn` always returns a `nudge` decision (kernel nudge node sets `effects.nudge` unconditionally), so `prepared.nudge?.compressibleRanges` is a reliable trigger input; it's already `viableRanges`-filtered by `prepare*` and re-filtered in preflight.
- Abort wiring: the `res.on("close")` listener is only registered on the slow path (after the fast-path returns), guarded by `writableEnded`.
- `stripKernelSummaries` state-awareness: strictly safer than before — the anchor is stripped only when the block's compress tool-call is actually in the payload; sole-carrier anchors (preflight blocks, and any tool-call-less block on the normal path) now survive. Count-tokens tests pin both sides of the rule.
- Plugin mode is not excluded from preflight, which is right — the overflow exists there too, and `prepare*` computes `nudge` regardless of `pluginMode`.
- e2e test asserts the right things: summary calls happen, blocks hold the upstream summary, rebuilt payload drops old markers but keeps the protected tail + summary, usage credits land at 1000 post-turn; negative test confirms no preflight when the context fits.

**Non-blocking observations (no action required, noting for the record):**
1. **First-request latency after a big switch**: preflight makes up to 8 rounds of non-streaming summarization calls (each ≤ 8192 output tokens) before forwarding. A 1M→260k switch with ~700k of excess can make the first request take a while, with no progress signal to the client. Bounded, and the alternative is a permanently stuck session, so this is the right trade — just worth knowing.
2. **Fit check is slightly optimistic**: it compares the previous turn's `lastInputTokens` against the limit, which excludes this request's new messages. Consistent with the existing "real numbers, never estimates" stance (`src/server.ts:970-977`), and the new content is small in the switch scenario.
3. **Summarization-call usage is not credited** to session stats — the upstream cost of the meta-calls is invisible in the panel. Minor observability gap.
4. **Single message larger than the small window** (e.g. one 300k-token tool result under a 260k model): the chunk exceeds the budget, the summarization call itself overflows, preflight degrades to "forward anyway" — that specific case stays stuck. Graceful, and consistent with the issue's stated fallback.

Conclusion: correct, well-tested, follows the repo conventions. I'm not merging — that's human-only per AGENTS.md. Ready for you to merge when you are.

ranxianglei:
Model: qwen3.8-27b (vllm)

Independent A/B test of this PR against issue #247 (real omp + window-enforcing mock upstream, isolated `PI_CODING_AGENT_DIR`):

**Control (v0.1.56):** 12 turns on a 100k-window model (payload 56k), switch to a 24k-window model → upstream `400 This model's maximum context length is 24000 tokens...`, session stuck. Bug reproduced.

**This PR (+ companion #264):** switch turn → `context 56035 tokens exceeds model window 23744; preflight compressing` → 5 summarization rounds (each ~7.9k est, well under the window) → `preflight compressed 5 range(s), ~39166 tokens saved (56035 → 16869)` → rebuilt payload est 20140 < 24000 → request passed, conversation continued, exit 0. ✅

**Critical companion finding (#264):** omp sends user items without `type:"message"`; the kernel projection silently drops them, so without #264 the overflow payload (mostly user messages) is invisible to preflight — this PR alone does nothing for omp. Merge both.

**Edge note (accepted, not a blocker):** with very small windows (e.g. 8.7k) where kernel protection (`preserveRecentMessages 5` + `preserveRecentTokens 5000`) is a large fraction of the window, preflight compresses one batch then stops (T1 pending → 0). Real switches target ≥262k windows where protection ≈2%, so this doesn't affect the #247 scenario.

Also verified: merge of this PR with current master (159e82b+ incl. #259/#260) is conflict-free and 646/646 green (earlier check), and this branch's preflight runs cleanly under plugin mode (omp native tools path).

ranxianglei:
Model: GLM-5.3 (zhipuai-lb/glm-5.3)

## Review verdict: ✅ Approve (with 2 nits)

Reviewed on a fresh clone at `/tmp/bc-review-254-264` (branch `pr-254`, and combined with #264 on `combined`).

### Diff reviewed
- `src/preflight.ts` (new, 289 lines): `preflightCompress()` + `summarizeRange()` + `estimateCoreMessages()`
- `src/server.ts`: preflight hook in `handle()` before forward; fit check on real post-fold payload
- `tests/preflight-compress.test.ts` (494 lines): full e2e — big→small model switch, preflight fires, payload fits
- `tests/count-tokens.test.ts`: extended

### Why it's correct
1. **Fit check runs on real payload** — uses `max(lastInputTokens, estimateCoreMessages(turn.messages))`, correctly handling the fresh-session-after-model-switch case (id rotates, lastInputTokens=0 but raw history overflows). This was the core of #247 and it's handled.
2. **Re-runs `processTurn` each round** — compress renumbers surviving refs; stale range refs would corrupt state. The loop correctly re-derives ranges.
3. **Guards against non-viable ranges** — `viableRanges()` filter + `minChars` floor before spending a summarization call, so `applyRanges` atomic failure is avoided.
4. **Summary cost netting** — `currentTokens -= compressed; += defaultCountTokens(summary)`; `lastInputTokens` adjusted so the next turn sees post-fold reality.

### Nits (non-blocking)
- `preflightCompress` line: `deps.session.stats.lastInputTokens += defaultCountTokens(summary)` double-counts when `applyRanges` already credited `compressCreditTokens` — verify the interaction is intended (tests pass either way, suggests it's conservative not wrong).
- `MAX_PREFLIGHT_ROUNDS = 8` is a magic number without a comment tying it to worst-case chunk math (`limit/CHUNK_FRACTION`).

### Verification (this environment)
| Check | Result |
|---|---|
| `npm run typecheck` (pr-254) | ✅ 0 errors |
| `npm run build` | ✅ BUILD_OK |
| Full suite (651 tests) | 649 pass / 2 fail — **both pre-existing on master** (`plugin-agent.test.ts`: manifest-fetch + inert-extension, network/env-dependent) |
| Combined with #264 (`combined` branch: 254+264 merged) | ✅ clean merge, no conflicts; typecheck/build/tests all green (same 2 pre-existing failures) |

Behavioral smoke on combined: mock upstream + real proxy via `startServer()`; big-window request → switch to small window → verify preflight summarization calls fire and the forwarded payload fits. Covered by the PR's own e2e test which passes.

ranxianglei:
[bot] Thanks — that A/B is the definitive validation (real omp, window-enforcing upstream, control reproduces the #247 stuck state, fixed path exits 0).

**Heads-up: the branch moved since my first review** (`1a9f23c` → `a7a0436`). Two follow-up commits landed:
- `126f081` — trigger + fit check now use the real post-fold payload size: `max(lastInputTokens, estimateCoreMessages(...))` at both the trigger (`src/server.ts:1508`) and the per-round fit check (`src/preflight.ts:229`). This closes my earlier observation #2 (fit check was optimistic on fresh sessions where `lastInputTokens = 0` but raw history overflows).
- `a0b80c7` — CJK-aware accounting: `estimateTokensFast` (chars/4, ~4× undercount on CJK) replaced with the kernel's `defaultCountTokens` in `estimateCoreMessages`/`splitChunks`/summary-cost, with the loop floored on the measured `lastInputTokens` baseline (covers system prompt + tool defs).

**Re-verified on the updated head `a7a0436`:** typecheck clean, 651/651 pass, build clean. (The 2 `plugin-agent.test.ts` failures in the GLM review didn't reproduce here — all 651 green in this environment.)

**On the two nits:**
1. **The "double-count" nit is a misread — the interaction is correct.** `applyRanges` only *subtracts*: `lastInputTokens = max(0, lastInputTokens - r.tokensCompressed)` (`src/stream.ts:277`). Preflight then adds back *only the summary's cost* — `lastInputTokens += defaultCountTokens(summary)` (`src/preflight.ts:274`) — because the summary re-enters the payload. Net: `-compressed + summaryCost` = the true post-fold size, mirrored in the loop-local `currentTokens` via the `compressCreditTokens` delta (`src/preflight.ts:270-273`). The `compressCreditTokens` credit itself is a separate usage-recorder counter that the rebuild zeros (`src/server.ts:984`) because the rebuilt payload is already post-fold — so it never leaks into the next turn's nudge decision. Without the `+= summary` line the fit check would *under*count (the conservative direction the reviewer suspected is actually the bug it prevents).
2. **`MAX_PREFLIGHT_ROUNDS = 8`** — agreed it deserves a one-line comment. Suggested wording: "Up to 8 oldest viable ranges per request (one range, all its chunks, per round); when exhausted the request forwards as-is and overflow self-heal takes over." Your A/B needed 5 rounds for a 56k→24k switch, so 8 has headroom for realistic ≥262k-window targets. Non-blocking either way.

**Merge order:** confirmed — #264 is a hard companion for omp (type-less user items are dropped by the kernel projection, so preflight is a no-op for omp without it). For non-omp clients (anthropic/openai wire) #254 stands alone. Both merged, the #247 scenario is fixed across the board.

Still not merging — human-only per AGENTS.md. Ready when you are: #264 and #254.

PR #258, opened 2026-08-26
fix: drop whitespace-only Responses message items at ingress

ranxianglei:
Model: qwen3.8-27b (vllm)

## Problem

omp `--resume` 会话里出现大量 1-token 空消息（模型自己枚举消息列表时发现）。诊断链：

- omp 本地会话文件是**一条**混合 content 块的 assistant 消息（`[thinking, text("\n\n"), toolCall, ...]`），全程 JSON——**不是按回车拆分**
- Responses API `input` 是扁平条目列表，没有"一条消息内混排块"的表达 → omp 序列化时把每个 text 块摊平成独立 message item；模型（SGLang 习惯）工具调用前吐的 `\n\n` 就成了独立空消息（协议必然，非 omp bug）
- pi 不受影响：anthropic wire 块数组可原样放一条消息，6 个真实会话 0 条独立空消息
- **bili 的放大**：给 1-token 空白盖 42 字符 `\x3cacp\x3e` 标签 + 编号 → 10 倍膨胀 + ref 噪音 + **粘性**（标签把空白变"非空"，之后每轮都被原样回放，纯 trim() 判空 dropped=0）

## Fix

- `dropWhitespaceResponsesMessages`：`prepareResponses` 投影前剥 render 标签后判空白，纯空白 message item 删除（倒序 splice，omp 无 type 的 user 条目也覆盖；遇非文本部件（refusal 等）整体保留不删）
- 标签包真内容的消息（如 `\x3cacp tokens="377"\x3em00002\x3c/acp\x3e` + 正文）**永不删**
- 与 #257（omp 原生插件模式）互补不重叠：原生模式解决绑定后的回放，本修解决 wire 模式（含绑定前首次请求）的上下文卫生，对所有 Responses 客户端长期有效

## Verification

- 新增 `tests/responses-empty-messages.test.ts` 4 测试（摊平形态/剥标签判空/string+developer 角色不删/非数组容忍）
- 真实 dump 回放：8 份历史请求 dropped 0/0/2/2/3/3/4/2——每请求剥 2-4 个空壳
- 全量 632/632 + typecheck + build 绿
- devlog: `devlog/2026-08-26_drop-empty-responses-messages/`

PR #259, opened 2026-08-26
regression: #257 omp native plugin mode + #258 drop whitespace-only Responses items

ranxianglei:
Model: qwen3.8-27b (vllm)

Joint regression merge of #257 + #258 (same flow as #250): local merge of both branches, full checks, then a live e2e exercising **both** features in one session.

## Merge

- base master `5982720`; merge #257 (`4ca494e`) + #258 (`7f33f9e`) — zero conflicts
- typecheck ✓ / **636/636** tests / build ✓

## e2e (omp 17.3.8 real binary, isolated PI_CODING_AGENT_DIR + content-routed mock upstream)

Mock returns a mixed turn on main round 1: whitespace `\n\n` text block + `acp_status` function_call. One run hit all four points:

| Point | Evidence |
|---|---|
| #257 native tools in main round | tools = omp 11 builtins + `compress/decompress/search_context/acp_status` (single copy) |
| #257 identity binding | `injectTool=false (plugin mode: wire injection suppressed)` from the first main-round request |
| #257 native tool really executes | `[plugin] tool acp_status executed via plugin (285 chars)` — real kernel adjudication via /__bili/plugin/tool |
| #258 whitespace drop | round-2 replay: `dropped 1 whitespace-only message item(s) before projection (flattened-turn artifact)`; upstream input = [developer, user, function_call, function_call_output], no WS item |

Merging this PR supersedes both #257 and #258 individually.

devlog: `devlog/2026-08-26_regression-257-258/`


ranxianglei:
[bot] On it — verifying the joint regression branch `2026-08-26_regression-257-258` (218c0df) against both PR branches.

⚠️ **Already found one problem before re-running anything:** the local merges pulled in the *original* PR commits, not the branch heads:

- #257 merge (4ca494e) contains only `f5342f3` — it is **missing `2a1bd22`** (`fix: retry omp identity register after failure instead of wedging the session`), the review fix pushed to `2026-08-26_omp-native-plugin-mode` at 09:56Z.
- #258 merge (7f33f9e) contains only `7e7fbb1` — it is **missing `21fa531`** (`docs: remove duplicated Fixes heading in CHANGELOG`), pushed to `2026-08-26_drop-empty-responses-messages` at 10:58Z.

So "merging this PR supersedes both #257 and #258 individually" is **not true as-is** — if #259 lands and #257/#258 close, both review fixes are lost (2a1bd22 is a real bug fix: without it a failed identity register wedges the session in wire mode for its whole life).

I'll merge both branch heads into the regression branch, re-run the full checks (typecheck / tests / build), push, and report back. The e2e evidence should carry over (2a1bd22 only changes behavior on register failure; 21fa531 is docs-only), but I'll re-verify the tree.

ranxianglei:
[bot] Fixed and re-verified. The branch now genuinely supersedes both PRs.

## What I changed (pushed, no force-push)

Two clean merges on top of 218c0df, pulling in each PR's **head** (not just its original commit):

- `a05cba5` — merge #257 head → adds `2a1bd22` (identity-register retry fix: a failed register no longer wedges the session in wire mode)
- `cffc5d5` — merge #258 head → adds `21fa531` (deduped the `### Fixes` heading in CHANGELOG)
- `871c604` — devlog addendum recording the gap + the fix

Verified all four PR commits (`f5342f3`, `2a1bd22`, `7e7fbb1`, `21fa531`) are now ancestors of the branch tip `871c604`. `version` untouched (0.1.55) — correct for a non-release branch.

## Re-run on the fixed tree (not the old one)

- **typecheck ✓**
- **636/636 tests ✓** (632 base + 4 from each PR)
- **build ✓** — `dist/index.js` 2.48 MB self-contained, `dist/agent/omp.js` 11.79 KB

## e2e evidence still holds for the new tree

The original e2e ran on 218c0df (before the two commits). Both added commits are provably inert on the e2e path:
- `2a1bd22` only changes behavior when the identity register **fails** (the retry path). The e2e exercised the happy path (bind succeeds on the first request), whose end state is identical.
- `21fa531` is docs-only.

So the four-point e2e result (native tools single-copy, identity binding from request 1, real kernel adjudication via `/__bili/plugin/tool`, whitespace item dropped before projection) carries over. I did not re-run the live omp e2e since the delta touches none of those paths — say the word if you want it re-run end-to-end on the new tip.

Ready for you to merge when happy — that's yours to do, not mine.

ranxianglei:
Model: qwen3.8-27b (vllm)

**Added 7e95b48: fix plugin-mode passthrough tag-echo strip.**

Wild-caught regression from #257 (omp native mode): the plugin-mode branch piped the upstream Responses stream VERBATIM, bypassing the #206 tag-echo stripper (which only lives in the compress-loop stream path). A fresh omp session re-ingested the model's fake `<acp>` tags verbatim → omp flattened them into session items → replay amplified → tag-echo loop crowded out real prose.

Fix: `pipePluginResponsesWithStrip` in src/plugin.ts — event-level passthrough running the same `createTagEchoFilter` state machine over `output_text.delta` (fast path when clean), tail-flush before done-family events, done payloads stripped, all other events byte-identical. server.ts uses it for Responses plugin mode.

Verified: unit 4/4 (split-delta strip, tail flush carries item_id, byte-identical clean events, done-payload strip), full 640/640, typecheck, build; real omp e2e — proxy logs `[tag-echo] stripped ... (plugin passthrough)`, omp receives prose with tag removed.

Also rebased onto the branch's new review-fix merges (2a1bd22 identity-register retry, 21fa531 CHANGELOG heading fix).

ranxianglei:
[bot] Reviewed `7e95b48` — **LGTM**. Verified:

**Topology** — linear on top of my review-fix merges (`871c604`); all prior commits intact (`f5342f3`, `2a1bd22`, `7e7fbb1`, `21fa531`). Version untouched (0.1.55), no `package.json` changes.

**Diff review** (`pipePluginResponsesWithStrip`, src/plugin.ts:568-685 + server.ts:1704-1708):
- Delta path runs the same `createTagEchoFilter` state machine as the compress-loop adapter (src/loop/adapter-responses.ts:267) — fast path is byte-identical when clean *and* the filter isn't pending (the new `pending()` = `held.length > 0 || swallowUntilClose` is the right gate: a clean delta arriving mid-swallow must still go through `push`).
- Tail flush matches the adapter's `flushFilter` design: synthetic `response.output_text.delta` carrying the last delta's `item_id`/`output_index`, emitted before the first done-family event — so the client's assembled text can't lose content, and the stripped done payload stays consistent with the deltas.
- Done-family payloads (`output_text.done` / `content_part.done` / `output_item.done` / `completed` / `failed` / `incomplete`) stripped via the existing `stripResponsesText` — same shapes the adapter covers.
- Usage sampling preserved (test 1 asserts `lastInputTokens` from the completed event); drain/destroyed handling mirrors `pipeThroughWithUsage`; server wiring keys off `protocol === "responses"` only.
- Tests: 4 new cases cover single-event strip, split-delta swallow + tail flush with `item_id`, byte-identical clean/function_call events, done-payload strip. Good coverage of the failure mode.

**Checks (this tree)**: typecheck ✓ · **640/640** tests ✓ (636 + 4) · build ✓ (dist/index.js 2.48 MB, self-contained).

**Non-blocking notes** (no action required for merge):
1. `pipePluginJson` (non-streaming plugin path) still forwards verbatim, while the compress-loop non-streaming path strips every round (`compressLoopResponsesJson`, src/compress-loop-responses.ts:186). Parity gap on a defensive path — Responses clients stream, so exposure is theoretical. Optional follow-up: run `stripResponsesText` there when `protocol === "responses"`.
2. `rebuildEvent` replaces only the first `data:` line; a multi-line data event (defensively parsed but never emitted by upstreams, which send single-line JSON) would rebuild malformed. Nit.
3. Stream cut without a done-family event drops any held tail — identical exposure to the compress-loop adapter (flushes on done-family only), so not new.
4. Plugin mode with the *openai* protocol is still verbatim (no tag strip) — the observed loop is omp/Responses, so scoping this commit to Responses is right; extend later only if another client shows the loop.

Branch is ready; merge is yours.

ranxianglei:
Model: qwen3.8-27b (vllm)

**Hotfix b5faa15** — follow-up to the strip pipe added in 7e95b48. Real-stream replay (227KB / 469 events from SGLang /v1/responses) caught a framing bug: tag-free done-family events were written WITHOUT their trailing blank line (`write(flushTail(out))` passed `rawEvent` bare while every passthrough path writes `rawEvent + "\n\n"`), fusing the next event's `event:` line into the previous data payload — omp (Bun) then throws `JSON Parse error: Unable to parse JSON string` mid-turn and the conversation dies. Also: flushed tail deltas are now emitted BEFORE the done-family event that triggers them (was after, semantically wrong ordering).

Verified: real-stream replay across 1KB/64B/7B chunk boundaries → 0 malformed data lines, 469/469 events, framing intact; new framing-level regression test locks this (every data line must parse; tag-free done event byte-identical). 641/641 + typecheck + build.

ranxianglei:
[bot] Reviewed `b5faa15` — **LGTM**. Verified:

**Topology** — linear on top of `7e95b48`; only `src/plugin.ts` (4 lines) + the test file changed.

**Root cause confirmed** — my audit of the done-family branch in 7e95b48: `rebuildEvent` ends its output with `\n\n`, but the tag-free branch passed `rawEvent` bare, so `write(flushTail(out))` emitted the done event without its terminator and the next event's `event:` line fused into its data payload — exactly the Bun `JSON Parse error` you saw. And `flushTail` was appending (`before +=`) the tail delta *after* the triggering done event. Both fixed minimally: `rawEvent + "\n\n"` on the raw branch, `flushTail(after)` now prepends the tail delta before the done-family event.

**Framing audit (all 8 write sites in the pipe)** — src/plugin.ts:613, 620, 642, 646, 653, 659 all `rawEvent + "\n\n"`; :636 `flushTail(out)` where both `out` branches now terminate (`rebuildEvent` already did, raw branch fixed) and the prepended tail delta carries its own `\n\n`; :656 `rebuildEvent` (terminates). No other unframed path remains.

**Tests** — the split-delta test now asserts ordering (`flushIdx < completedIdx`, tail delta precedes the done event that triggers it, not just presence), and the new framing regression test locks the invariant: 4 events in → 4 data lines out, every data line parses as JSON, tag-free done event byte-identical, stream ends with `\n\n`. That's the right shape of regression test for this class of bug.

**Checks (this tree)**: typecheck ✓ · **641/641** ✓ (640 + 1) · build ✓ (dist/index.js 2.48 MB).

Fair note: I missed this in my `7e95b48` review — I audited the done-family path and the `rebuildEvent` multi-line nit but didn't trace the raw branch's terminator. The real-stream replay (227KB/469 events across 1KB/64B/7B chunk boundaries) is exactly the coverage the unit matrix couldn't provide.

Branch is ready; merge is yours.

PR #260, opened 2026-08-26
fix: plugin passthrough review follow-ups (non-streaming strip parity, stream-cut tail flush, multi-line data, malformed parts)

ranxianglei:
Model: qwen3.8-27b (vllm)

Addresses the non-blocking review notes left on #257/#258/#259 (all three merged):

1. **`pipePluginJson` verbatim parity gap** (#259 note 1) — the non-streaming plugin path forwarded Responses bodies verbatim while the compress loop's JSON branch (`compressLoopResponsesJson`) strips render tags every round. Now strips via `stripResponsesText` when `protocol === "responses"` and the body contains tags; openai protocol and tag-free bodies stay byte-identical.
2. **`rebuildEvent` multi-line `data:` hardening** (#259 note 2) — subsequent `data:` lines are now dropped when the first is replaced, so a (hypothetical) multi-line data payload collapses into one valid JSON line instead of fusing two payloads.
3. **Held-tail flush on stream cut** (#259 note 3) — if the upstream stream ends without a done-family event, the tag-echo filter's held tail is flushed as a final delta instead of being dropped.
4. **Malformed content parts preserved** (#258 minor) — non-object content parts (e.g. `content: [42]`) now mark the item as mixed → preserved, instead of silently counting toward emptiness.

Not changed (deliberate, #259 note 4): openai-protocol plugin passthrough stays verbatim — the observed tag-echo loop is omp/Responses only; extend only if another client shows it.

Checks: typecheck ✓ · **646/646** ✓ (641 + 5 new) · build ✓


ranxianglei:
[bot] Reviewed `2026-08-26_plugin-passthrough-polish` (single commit `270476d`) against master. All four follow-ups are implemented as described and the scope is exactly what the issue claims.

**1. `pipePluginJson` strip parity** — correct. It reuses the same `stripResponsesText` that `compressLoopResponsesJson` applies every round (`src/compress-loop-responses.ts:178`), gated on `protocol === "responses"` + `containsRenderTagText(text)` (`src/plugin.ts:732-743`). The detection regex matches JSON-escaped tag forms (`<acp ` whitespace lookahead plus the `\u003c` alternative), so tags inside string values are caught. Openai-protocol and tag-free bodies fall through to the verbatim `res.end(text)` — both covered by the new byte-identity test.

**2. `rebuildEvent` multi-line `data:`** — correct. Subsequent `data:` lines are dropped after the first replacement (`src/plugin.ts:684-695`), so the rebuilt event is one valid JSON line. Consistent with the parser, which already joins multi-line data with `\n` before parsing (`src/plugin.ts:611`).

**3. Held-tail flush on stream cut** — correct and safe. `tagFilter.flush()` is idempotent (clears `swallowed`/`held`/`swallowUntilClose`, second call returns `""`), so the post-loop `flushTail("")` (`src/plugin.ts:664-669`) is a no-op when the stream already ended on a done-family event, and the `!res.destroyed && !res.writableEnded` guard skips it when the client disconnected. The flushed delta reuses `lastDeltaMeta` (item_id/output_index), so the client's assembled text stays coherent.

**4. Malformed content parts** — correct. Non-object parts now set `mixed = true; break` (`src/loop/adapter-responses.ts:74-77`), leaving `text === undefined` → item preserved. Previously `content: [42]` joined to `""` and was silently dropped.

**Tests** — 5 new tests, one per behavior plus the openai/tag-free verbatim regression guard; the `makeRes` mock gained `end(b)` capture which the `pipePluginJson` tests need.

**Checks (ran locally on the branch):** typecheck ✓ · 646/646 tests ✓ · build ✓. No `version`/lockfile changes on this content branch — compliant with the one-version-one-branch rule.

No blockers. Ready for merge.

Issue #221, opened 2026-08-24
偶尔模型在acp工具处停止执行


stirp:
环境信息:

<img width="503" height="207" alt="Image" src="https://github.com/user-attachments/assets/287f909e-cfb0-4141-8327-836fdcbd0a7b" />

经过:
本地有一份周报md, 我需要让Agent将其写入wps在线文档, 提示语是:
"周报要写在 https://365.kdocs.cn/l/********?from=docs&reqtype=kdocs&startTime=1787561760723&createDirect=true&newFile=true"

<img width="1277" height="765" alt="Image" src="https://github.com/user-attachments/assets/9bdd6de1-bafa-447c-8dfe-8eca0bc00f71" />

如图所示, 在执行一个工具后, 触发一段thinking, 然后跟了ACP的一段结果,但是内容似乎不是ACP的内容.

之后我提示"现在创建好了 https://365.kdocs.cn/l/**********"

<img width="1242" height="518" alt="Image" src="https://github.com/user-attachments/assets/e54e8f80-3491-49b6-b39f-47ce95208c81" />

Agent再次停止在一段ACP返回之后.

相关日志:

```
8月 24 19:39:04 desktop node[1375636]: 2026-08-24T11:39:04.669Z [info] forward POST → https://aihub.*****.cn/v1/messages?beta=true
8月 24 19:39:11 srp-desktop node[1375636]: 2026-08-24T11:39:11.552Z [info] [bbd62324479f7502] processTurn: 155 msgs, renderTags=text-only, 37 text tagged, 0 tool tagged (should be 0 with text-only)
8月 24 19:39:11 desktop node[1375636]: 2026-08-24T11:39:11.552Z [info] [bbd62324479f7502] nudge idle: usage=3% (5720/204800), growth=-11784/22500 (ref=17504, interval=50000), pendingT1=29238/50000 model=M>
8月 24 19:39:11 desktop node[1375636]: 2026-08-24T11:39:11.554Z [info] forward POST → https://aihub.*****.cn/v1/messages?beta=true
8月 24 19:39:20 desktop node[1375636]: 2026-08-24T11:39:20.119Z [info] [bbd62324479f7502] processTurn: 163 msgs, renderTags=text-only, 38 text tagged, 0 tool tagged (should be 0 with text-only)
[...]

PR #261, opened 2026-08-26
fix: acp-loop auto-retry on no-content upstream truncation (#221)

ranxianglei:
## Context

Follow-up to #221 (agent stops at the acp tool). PR #239 made the upstream-truncation failure *visible* (`[acp-proxy: upstream stream truncated ...]`), and the user's logs confirm that is the failure mode:

```
[error] [acp-loop] upstream stream truncated (no completion event; round 1, 0 text chars + 0 reasoning chars received)
```

But the agent still stopped, because the proxy surfaced the error and terminated the turn.

## Root cause

The relay cuts a long-running request (MiniMax-M3 thinking server-side for minutes) before any token is emitted. The upstream SSE ends with no completion event and **zero content** — nothing reached the client.

## Fix

A truncation that produced no text/reasoning is a clean re-send: the loop now re-requests the same body and continues instead of erroring out. The retry is treated as a later round, so `message_start` is not re-emitted and no prefix is duplicated.

- Bounded by `BILI_TRUNCATION_RETRY_MAX` (default 2 retries; `0` restores the fail-fast error).
- A **partial-content** truncation is deliberately NOT retried (the model would regenerate from scratch and duplicate the already-streamed prefix) and still surfaces the visible error.

## Tests

`tests/upstream-truncation.test.ts` — 3 new cases:
- no-content truncation auto-retries and delivers the full response (single `message_start`, no error marker)
- no-content truncation that keeps truncating surfaces the error after retries are exhausted
- partial-content truncation is NOT auto-retried

Full suite: 644 pass. typecheck + build clean.


ranxianglei:
[bot] Reviewed the PR, verified the core mechanism, and found (and fixed) two edge-case holes in the retry guard. Pushed a follow-up commit to the branch: `03b7f8f`.

**Verified as designed**
- No-content truncation → clean re-send of the same body: `coreMessages` is untouched when no proxy tool ran, so the re-request is byte-identical; the retry flows through the existing re-request machinery, is treated as a later round, and `message_start` (a `firstRoundOnly` meta event) is not re-emitted. ✓
- Partial-content truncation still surfaces the visible error. ✓
- `BILI_TRUNCATION_RETRY_MAX` default 2 / `0` = fail-fast. ✓

**Bug 1 — retry after a flushed real tool call (the important one).** A truncation can carry a *completed* tool call: the Anthropic adapter flushes `tool_use` at `content_block_stop` and the Responses adapter at `output_item.done` — both before the completion event. (OpenAI chat is safe: it settles tool calls only at `finish_reason`/`[DONE]`.) With the original guard (`!sawDone && zeroContent && budget`), a truncated stream that had zero text/reasoning but one closed real tool call would (1) emit that tool call to the client, then (2) auto-retry — streaming the regenerated response *behind* the tool call in the same client response. The client would see one assistant message containing the tool call plus a duplicate regeneration, and its tool-result follow-up would be out of sync with the stream. Fix: `canRetryTruncation` now also requires `realCalls === 0`; that case surfaces the visible error as before. (Proxy-tool calls are unaffected — they take the normal `reRequest` path, and the retry counter/log no longer misattribute that re-request.)

**Bug 2 — final-round truncation swallowed the error.** On round `MAX_LOOP_ROUNDS` (10), a zero-content truncation passed the error check (retry allowed) and fell into the loop-limit branch, ending the turn with a degenerate empty `stop_reason: "length"` completion instead of the visible error. Fix: `canRetryTruncation` now also requires `round < MAX_LOOP_ROUNDS`.

**Tests** — added 2 regression cases to `tests/upstream-truncation.test.ts` (both confirmed to fail on the pre-fix code):
- truncation with a fully closed real `tool_use` block → error surfaces, no retry content behind the tool call
- 9 proxy-tool rounds + zero-content truncation on round 10 → error surfaces, no `length` completion

Full suite: **646 pass** (was 644). typecheck + build clean.

PR is ready for merge whenever you are — I can't merge it myself (AGENTS.md: merges are human-only).

ranxianglei:
## 📦 Built Package Artifact

**Branch:** `2026-08-26_acp-truncation-autoretry` (4d87ab2)

### Option A — Install from npm PR tag (recommended)

```bash
npm install -g billion-context@pr-261
```

Each push to this PR publishes a new version under the `pr-261` npm tag.

### Option B — Download artifact

1. Download the artifact from the [Actions run](https://github.com/ranxianglei/billion-context/actions/runs/32992862445)
2. Extract the tarball and install:

```bash
tar xzf billion-context-pr261.tgz
npm install -g package
```

---
*This comment is automatically updated on each push.*

ranxianglei:
Closing: superseded — master loop/core.ts (the `truncationRetried` one-shot retry, "upstream truncated before any content reached the client; retrying fetch once") implements exactly this no-content-truncation auto-retry, shipped via #440/#444 (v0.1.74) and hardened since.

PR #269, opened 2026-08-26
fix: identity plugin binding survives model switches across sessions

ranxianglei:
Model: qwen3.8-27b (vllm)

## Root cause

`consumePluginRegisterFor` treated an identity registration (`identity: true`, the omp / claude-code path) as a one-shot token: the first request carrying the conversation id consumed and **deleted** it, binding only THAT session into plugin mode.

Switching models or upstreams mid-conversation resolves a **new session** (session key = `protocol|upstream|apiKey|conversation`). The registration was already gone, so the new session silently dropped back to wire injection — native tools gone, the model roleplaying `acp_status` output in the conversation.

Wild-caught on the omp TUI: GLM (chat) → qwen (responses) switch, "原生模式消失了".

## Reproduction (full-fidelity launcher TUI + dual-protocol mocks)

- fresh conversation, chat-protocol model: identity register consumed, plugin mode ✓ (`injectTool=false`)
- switch to responses-protocol model (same conversation, new session): `injectTool=true` — wire ✗

Plugin-side flow verified healthy via an instrumented plugin (module/session_start/before_provider_request logs): manifest fetch, tool registration, identity register POST all fire correctly in one-shot, TUI, and `-r` resume modes. The bug is purely the server-side one-shot consume.

## Fix

Identity registrations are sticky for the conversation: `consumePluginRegisterFor` keeps the entry and refreshes LRU order on hit, bounded by the existing `MAX_PENDING_REGISTERS = 64` cap (least-recently-active conversations evicted first). This matches the documented #162 semantic: "bind the moment any of their requests shows up" — every request, not just the first.

## Test

`launcher identity binding survives model switches (one conversation, multiple sessions)` — dual upstreams, same conversation id, asserts wire suppression on BOTH sessions. Red/green verified: old code fails the new test, fixed code passes.

- launcher-plugin-mode: 8/8
- full suite: 647/647, typecheck, build ✅

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ Versions follow the merge of a `*_release-v*` branch; CI publishes to npm on tag
 
 ## [Unreleased]
 
+### Fixes
+
+- **Identity plugin binding survives model switches (one conversation, many sessions)**: an identity registration (`POST /__bili/plugin/register` with `identity: true` — the omp/claude-code path) was consumed one-shot: the first request that carried the conversation id ate the registration and bound ITS session into plugin mode. Switching models or upstreams mid-conversation resolves a NEW session (session key = protocol|upstream|apiKey|conversation) — the registration was already gone, so the omp TUI flow "chat model → responses model" silently dropped back to wire injection (native tools gone, model roleplaying `acp_status` output). Identity registrations are now sticky for the conversation: every request carrying the id binds, LRU-refreshed and bounded by the same 64-entry cap. Wild-caught as "原生模式消失了" after a GLM-chat → qwen-responses model switch.
+
 ### Features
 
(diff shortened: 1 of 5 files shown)


ranxianglei:
[bot] Reviewed `2026-08-26_plugin-identity-sticky` (7f2e4a0) — verified independently, all green:

- **Fix is correct**: `registeredIds` is a `Map` (insertion-ordered), so delete+re-set in `consumePluginRegisterFor` (`src/plugin.ts:247`) refreshes LRU position, and the existing `MAX_PENDING_REGISTERS = 64` eviction loop (`src/plugin.ts:215-218`) now evicts least-recently-*active* conversations instead of least-recently-*registered*. Headless (`identity: false`) path is untouched — still one-shot with TTL, which is correct since those requests carry no matching id.
- **Red/green confirmed**: swapped in master's `src/plugin.ts` and ran the new test → fails with `AssertionError` (7/8, "upstream B: identity binding survives the model switch"). Restored fixed code → 8/8.
- **Full pre-flight**: typecheck ✅ · `tests/launcher-plugin-mode.test.ts` 8/8 ✅ · full suite 647/647 ✅ · build ✅
- Branch hygiene: single commit, content-only (no `version` bump), CHANGELOG under `[Unreleased]` ✅

Ready to merge — per AGENTS.md the merge is yours to do.

PR #284, opened 2026-08-27
fix: resolve registry window for relay hosts via provider-prefixed key scan

ranxianglei:
Closes the 64k window mismatch on relay hosts: registryLookup now scans provider-prefixed candidates when the host is unknown and the bare key misses (all matches must agree). Includes 5 tests incl. #282 regression.

Agent-prepared (branch pushed by ework-daemon via credhelper); PR opened via git credential API per AGENTS.md flow.

### DIFF
--- a/src/registry.ts
+++ b/src/registry.ts
@@ -261,6 +261,25 @@ function registryLookup(reg: RegistryShape | null, model: string, host?: string)
         const ctx = entry?.limit?.context;
         if (typeof ctx === "number" && ctx > 0) return ctx;
     }
+    // Relay host (not in HOST_TO_PROVIDER): the bare name can miss while the
+    // model exists under a provider-prefixed key (a relay serving
+    // "deepseek-v4-flash" is stored as "deepseek/deepseek-v4-flash"). Scan
+    // */<model> and adopt the value only when every match agrees — a
+    // conflict means different deployments, and guessing is worse than the
+    // static table. Known-provider hosts keep the old behavior: a miss there
+    // means the model is genuinely unlisted for that provider.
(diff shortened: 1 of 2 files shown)


ranxianglei:
[bot] Reviewed PR #284 (`2026-08-27_registry-relay-fallback`, 1 commit ahead of master, master fully contained).

**Diff review** — `src/registry.ts:264-282` + `tests/registry.test.ts` (+5 tests):
- Fallback only fires when `provider === undefined` (unknown relay host or no host), so known-provider hosts keep the old strict behavior — a miss on `api.minimax.chat` still means unlisted, no cross-provider bleed.
- Scan is slash-bounded (`key.endsWith("/" + model)`), so `"v4-flash"` can't match `"deepseek-v4-flash"`.
- Conflict rule verified: two provider-prefixed keys with different windows → `undefined` (falls back to the static table rather than guessing).
- `peekRegistryContext`/`bundledSnapshotLookup`/`contextFromRegistry` all route through `registryLookup`, so the fix covers warm cache, cold bundled snapshot, and the async path uniformly.

**Pre-flight (run on the PR branch):**
- `npm run typecheck` — clean
- `npm test` — 662/662 pass, incl. the 5 new registry tests (relay fallback, conflict no-guess, slash-bounded, known-provider no-scan, and the #282 bundled-snapshot regression: `bundledSnapshotLookup("deepseek-v4-flash", "freeinference.org")` → 1,000,000)
- `npm run build` — success

PR is open and ready. Per AGENTS.md, merging is human-only — please merge when ready: https://github.com/ranxianglei/billion-context/pull/284

PR #249, opened 2026-08-25
Fix Codex Responses session continuity

Haerbin23456:
## Why this change is needed

Codex on ChatGPT exercises the Responses API differently from the API-key clients this proxy was primarily tracking.

In an observed long-running Codex task, one stable Codex task label was split across seven internal billion-context sessions. Compression completed in one shard, but a later request landed in a fresh shard with no token baseline or compression blocks. In a separate failure mode, Codex emitted a native `compaction_trigger` after ACP had already summarized older history. Forwarding the full raw history made the remote compaction request exceed its own input budget, while injecting anything after the trigger made the request invalid because the trigger must be final.

There are three interacting causes:

1. The session hash used the complete Authorization bearer. ChatGPT OAuth access tokens rotate, so the same account and task could receive a new internal session id.
2. Responses subagent isolation used the complete instructions string. Codex instructions contain changing runtime context, so incidental prompt changes could look like a different subagent.
3. A terminal native `compaction_trigger` needs special handling: it must remain final, but the request must still benefit from ACP blocks that already replaced older history.

### Stable account identity, narrowly scoped

For chatgpt.com upstreams, prefer `chatgpt-account-id` over the rotating bearer when deriving the proxy session id. This keeps a task stable across OAuth refreshes while preserving account isolation.

Other upstreams retain the existing credential-based behavior. This avoids changing API-key and custom-endpoint semantics.

### Stable Codex agent identity with compatibility fallback

When Codex thread metadata is present, use its stable agent identity to isolate per-agent compression state. This keeps dynamic instruction tails from fragmenting one agent into multiple proxy sessions, while still separating named agents in the same task.

If the metadata is absent, keep the legacy full-instructions namespace. Existing Responses clients therefore retain their previous behavior instead of being collapsed into a new shared namespace.

### Native compaction preserves prior ACP work

If the final input item is `compaction_trigger`, project the request through the session's existing ACP state with tag/tool/nudge injection disabled. Covered raw history is removed, the surviving Responses items remain byte-preserved, the active compress call carrying the summary remains available, and `compaction_trigger` remains the final item.

The projection is read-only: merely attempting native compaction does not advance ACP survival state, reap blocks, or overwrite the session snapshot. If projection itself fails, the sanitized original input is forwarded unchanged.

[...]

PR #287, opened 2026-08-27
fix: key sessions on the clients native session id, not rotating credentials

ranxianglei:
refs #280
refs #286

## Background

The proxy session id was `hash(protocol | upstream | credential | conversation)`. Two of those dimensions are volatile even within ONE conversation:

- **credential** — for Codex/ChatGPT this is a rotating OAuth bearer;
- **upstream** — changes when a user switches relays.

Either change minted a brand-new session id and orphaned ALL compression state — in-memory blocks, `lastInputTokens`, and the persisted record (keyed by the same id, so it couldn't be restored either). That is the "sessions 缺失" root cause behind #280: after the churn, the session restarts at usage=0% with the full raw history replayed, the nudge goes blind, and the payload overflows the real window.

## Change

The client's **native session id** (codex `session-id` header, claude `x-claude-code-session-id`, opencode `x-opencode-session`, Responses body `session_id`, `prompt_cache_key`) is the only invariant tied to a conversation — the user can switch accounts or relays anytime, and the session id is what stays bound to the conversation (#286, owner-confirmed direction):

- `src/session-id.ts`: `deriveSessionId` takes the resolved identity; when it is **client-provided** (`identity.clientProvided`), the session id is `hash(protocol | conversation)` only.
- `src/server.ts`: passes the resolved identity (per protocol: header signal for anthropic, the openai/responses identity objects) into the derivation.
- The **content-fingerprint fallback** (hermes, dsh, anonymous clients — no client signal) keeps the 4-way hash: the fingerprint is a content hash with a real collision surface and still needs the account/upstream dimensions for isolation. `previous_response_id` (clientDerived, `clientProvided:false`) also stays 4-way.

One-time cost: existing persisted sessions re-key once (old records orphaned) — accepted per #286.

## Tests

- Codex native session id stable across bearer rotation AND relay switch; different conversation/protocol still separate.
- Body `session_id` and `prompt_cache_key` identities credential-independent.
- Fingerprint + `previous_response_id` keep 4-way isolation (no cross-account/cross-provider bleed).

Full suite: 660/660 pass.

## Coordination

Complements #249 (thread-id subagent refinement + compaction fast path) — independent change, no overlap in the identity-derivation code path.

<sub>🤖 ework agent</sub>

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -756,8 +756,20 @@ async function handle(
               : subagentNamespace(
                   responsesIdentity?.value ?? conversationSignalResponses(parsed as ResponsesRequestBody, convHeader),
                   (parsed as ResponsesRequestBody).instructions,
-              );
-        const sessionId = deriveProxySessionId(req.headers, protocol, upstreamOrigin, conversation);
+                );
+        // #286: a client-provided native session id (codex session-id header,
+        // claude x-claude-code-session-id, body session_id, prompt_cache_key…)
+        // is the only invariant tied to a conversation — key the session on
+        // protocol+conversation only, so credential rotation (OAuth bearer)
+        // and relay/upstream switches don't orphan the compression state.
+        const identity: ConversationIdentity | undefined = protocol === "anthropic"
+            ? convHeader
+              ? { value: conversation, source: "header", clientProvided: true }
+              : { value: conversation, source: "content-fingerprint", clientProvided: false }
+            : protocol === "openai"
+              ? openaiIdentity
+              : responsesIdentity;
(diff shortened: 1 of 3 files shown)


ranxianglei:
Superseded by #288 (merged): the verbatim session-id design replaces the hash(protocol|conversation) approach here, and deriveSessionId no longer exists on master. Closing.

ranxianglei:

Acknowledged — closing #287 as superseded by #288. The verbatim native session-id design is the one that landed; the `hash(protocol | conversation)` fallback approach here is discarded. No further action on this branch.

PR #305, opened 2026-08-27
fix: honor anthropic-beta context-1m to size the context window

### LINKED ISSUE #302: fix: preflight compress must fail fast on upstream 429 instead of forwarding the over-window payload as-is
Model: glm-5.3

## Background

Reported in #292 (log evidence there). When the preflight (overflow) compression fails because the upstream is rate-limiting, the proxy currently gives up and **forwards the original over-window payload unchanged**.

## Evidence (from `bili (8).log`)

At `13:26:21` the outer bili computed `input=261206` (double-counted usage, see #300) against a 200K window and fired EMERGENCY preflight compression:

```
13:26:21.161 ... round 1 input=261206 cached=118663 ... EMERGENCY T1 (131%)
13:26:21.803 [warn] upstream 429 ...
13:26:23.811 [warn] upstream 429 ...
... forwarding as-is (50 messages)
```

The preflight summary requests reuse the same already-rate-limited OAuth upstream, so both attempts 429'd, the loop aborted, and the full 261K-token request was sent upstream anyway:

- wasted quota (guaranteed 400 for exceeding context),
- user-visible failure that looks like a proxy bug rather than rate limiting,
- retry storms downstream (client sees 400, retries, same loop).

## Fix required

1. When preflight compression cannot proceed (upstream 429/5xx, or compress budget exhausted), **fail fast**: return a clear error response to the client (502/503 with a diagnostic message naming the cause) instead of forwarding an over-window payload.
2. If (and only if) the payload actually fits the window, falling back to forwarding as-is is fine.
3. Optional hardening: use a separate, cheaper model/endpoint for preflight summaries (config), or exponential backoff with a small cap before failing fast.

## Acceptance

- Test: overflow request + summary upstream returning 429 → client receives a structured error, and the over-window body is NOT forwarded upstream.
- Test: payload within window + preflight 429 → request still forwarded (no false positive).

ranxianglei:
## What

When an Anthropic-format request carries an `anthropic-beta` header with a larger-context beta (e.g. `context-1m-2025-08-07`), the client has negotiated a **1M-token** upstream window. The proxy now sizes its own context window to match (1,000,000), overriding the model-table default (200K for claude).

Fixes #302 (reported in #292).

## Why

The `anthropic-beta` header is forwarded upstream (so the provider correctly serves 1M), but bili's *own* window resolution missed it. The client-facing instance sized its window at the model-table 200K, so a 261K-token conversation read as **131% → EMERGENCY T1**, when it was really only **26%** of the 1M window the client negotiated. The inner instance (which saw the header) correctly sized 1M — the two instances disagreed.

## How

- New `anthropicBetaContextWindow(headers)` (src/server.ts): parses the `anthropic-beta` header for `context-Nm` betas → `N × 1,000,000` (largest wins). Generalizes to future larger-context betas; rejects malformed / non-context tokens.
- The beta is resolved at the **top** of the native-window cascade (above plugin/launcher/registry/table) because it is the most direct per-request evidence of the window the upstream will serve — the table/registry only list the *standard* window. It also clears the `nativeFromFallback` flag (authoritative, not a low-confidence fallback, so the effective-window floor must not apply).
- Resolved **per-request** (re-read every request): the header may appear/disappear between requests of the same session.

Because the window flows into `reqConfig.modelContextLimit`, this single change fixes all four surfaces the issue lists: compression thresholds, the EMERGENCY trigger, usage-percentage logs, and the window reported to clients/plugins (`effectiveContextLimit`).

## Tests

- **Unit** (tests/model-windows.test.ts, +7): `context-1m-2025-08-07` → 1,000,000; no header / no context beta → undefined (model default); mixed beta list; array header value; case/whitespace; `context-Nm` generalization + largest-wins; malformed rejection.
- **e2e** (tests/anthropic-beta-window.test.ts, +3):
  - With the beta → 261K input is 26% of 1M → **no EMERGENCY nudge**.
  - Without the beta → the same 261K input is 131% of 200K → **EMERGENCY nudge fires** (control, proves the scenario can fire).
  - Plugin `effectiveContextLimit` is 1M with the beta and drops to 200K when the header disappears on the next request of the same session (per-request resolution).
- Verified the e2e tests are meaningful: with the fix reverted, tests 1 and 3 fail and the control (test 2) still passes.

## Pre-flight

- `npm run typecheck` — clean
- `npm test` — 701 pass, 0 fail
- `npm run build` — success

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -100,6 +100,31 @@ function launcherContextWindow(model: string): number | undefined {
     return LAUNCHER_MODEL_WINDOWS[model];
 }
 
+/** Parse an `anthropic-beta` header for a larger-context beta (e.g.
+ *  `context-1m-2025-08-07` → 1,000,000). The beta lets the CLIENT negotiate a
+ *  window beyond the model's standard size, so it is the most direct per-request
+ *  evidence of the window the upstream will actually serve — it outranks the
+ *  model table / registry (which list the STANDARD window, e.g. 200K for claude)
+ *  and must be re-read on every request (the header may appear/disappear between
+ *  requests of the same session, #302). `context-Nm` generalizes to future
+ *  larger-context betas (N × 1,000,000). Returns the largest requested window,
+ *  or undefined when no context beta is present. */
+export function anthropicBetaContextWindow(headers: Record<string, string | string[] | undefined>): number | undefined {
+    const raw = headers["anthropic-beta"];
+    if (raw === undefined) return undefined;
+    const list = Array.isArray(raw) ? raw.join(",") : raw;
+    let best: number | undefined;
+    for (const part of list.split(",")) {
+        const m = /^context-(\d+)m\b/.exec(part.trim().toLowerCase());
+        if (!m) continue;
+        const n = Number.parseInt(m[1], 10);
+        if (!Number.isFinite(n) || n <= 0) continue;
+        const w = n * 1_000_000;
+        if (best === undefined || w > best) best = w;
+    }
+    return best;
+}
+
 /** Session ids are client-provided verbatim (#286) — sanitize before using
  *  one in a debug-dump FILENAME so a hostile value cannot escape the dir. */
 function safeSessionId(id: string | undefined): string {
@@ -702,9 +727,13 @@ async function handle(
         const model = (parsed as { model?: string }).model;
         if (model) {
             const embeddedUrl = route?.rewrittenUrl;
-            // Native-window resolution order: (1) a cooperative plugin's
-            // report (the agent's own config — most authoritative, gated on
-            // the x-bili-plugin marker so a plain client cannot rewrite the
+            // Native-window resolution order: (0) the client's `anthropic-beta`
+            // larger-context negotiation (context-1m-… → 1,000,000) — the most
+            // direct per-request evidence of the window the upstream will serve,
+            // so it outranks every static source (the model table / registry
+            // list the STANDARD window, e.g. 200K for claude); (1) a cooperative
+            // plugin's report (the agent's own config — most authoritative, gated
+            // on the x-bili-plugin marker so a plain client cannot rewrite the
             // nudge denominator by name); (1b) the launcher's per-model
             // windows (BILI_LAUNCHER_MODEL_WINDOWS — the client's own
             // models.json/models.yml contextWindow, authoritative for this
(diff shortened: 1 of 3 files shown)


ranxianglei:

[bot] Reviewed PR #305 (branch `2026-08-28_anthropic-beta-1m-window`) against #302. Verdict: **LGTM — ready to merge** (merge is human-only).

**Code review** (`src/server.ts`)
- `anthropicBetaContextWindow(headers)`: the regex `/^context-(\d+)m\b/` is anchored + word-bounded, so it matches `context-1m-…` / `context-2m-…` and correctly rejects `context-m-…`, `context-1x-…`, and `my-context-1m-…`. Largest-wins, array-header join, and case/whitespace handling are all correct.
- Cascade placement is right: the beta sits at the top of the native-window cascade, but operator `compress.modelContextLimit` is still applied on top in `resolveRequestConfig` (`src/compress-settings.ts:170`), so the documented invariant "operator tuning outranks everything" holds.
- `nativeFromFallback` is cleared when the beta is present — correct, since the beta is authoritative (the client's own runtime negotiation), so `FALLBACK_EFFECTIVE_WINDOW_FLOOR` must not raise the window.
- Resolved **per-request** (reads `req.headers` every request), matching the requirement that the header may appear/disappear between requests of one session.
- The window cascade is the single resolution site in `handle` (`src/server.ts:749-769`); all four surfaces (compression thresholds, EMERGENCY trigger, usage logs, `effectiveContextLimit`) flow from `reqConfig.modelContextLimit`. The only other `modelContextLimit` writes are the downstream learned-limit (only lowers) and output-headroom reservation — both pre-existing. So "one change fixes all four surfaces" checks out.

**Pre-flight** (ran on the branch):
- `npm run typecheck` — clean
- `npm test` — 701 pass / 0 fail
- `npm run build` — success

**Test meaningfulness** (verified, not just claimed): I reverted the cascade change (kept the tests) → e2e test 1 (beta → no EMERGENCY) and test 3 (plugin `effectiveContextLimit` = 1M) both **fail**, while the control test 2 (no beta → EMERGENCY fires) still **passes**. Restored the fix → all 3 pass. So the tests genuinely exercise the fix, and the control proves the scenario can fire.

No changes needed. Please merge when ready: https://github.com/ranxianglei/billion-context/pull/305

ranxianglei:
Model: glm-5.3

Reviewed (supervisor pass): `anthropicBetaContextWindow` parses `context-Nm` from the `anthropic-beta` header per request, takes the max, and slots ahead of plugin/launcher/registry/configured/table. `nativeFromFallback` correctly guarded. Tests cover the 200K-table vs 1M-beta case including the 261206-token scenario from #292. Good to merge.

Issue #301, opened 2026-08-27
fix: honor anthropic-beta context-1m to size the context window


ranxianglei:
Model: glm-5.3

## Background

Reported in #292 (log evidence there). When Claude Code sends `anthropic-beta: context-1m-2025-08-07`, the upstream context window is 1M tokens, not 200K.

## Evidence (from `bili (8).log`)

The chain was claude → outer bili (port 8787) → inner bili (port 8787 on the remote Mac, via prefix instance 50157):

- Outer session used window **200000** and logged `EMERGENCY T1 ... 131%` at `round 1 input=261206` — but 261206 is only 26% of the 1M window the client actually negotiated.
- The inner instance correctly sized its window at **1000000** for the same conversation (`claude-fable-5`).

So the two instances disagreed: the one facing the client missed the beta header.

## Fix required

1. When an Anthropic-format request carries `anthropic-beta` containing a `context-1m-…` (or future larger-context) beta, the effective context window for that session must be sized accordingly (1,000,000), overriding the model-table default (200K).
2. This should apply wherever the window is resolved for: compression thresholds, EMERGENCY trigger, usage-percentage logs, and the window reported to clients/plugins.
3. Keep the resolution per-request (beta header may appear/disappear between requests of the same session).

## Acceptance

- Unit test: request with `anthropic-beta: context-1m-2025-08-07` → window 1,000,000; without it → model default.
- e2e or integration test showing EMERGENCY does not fire at ~261K input under the 1M beta.

ranxianglei:
> 🤖 ework agent · vllm-qwen/qwen3.8-27b

[bot] Fixed — PR https://github.com/ranxianglei/billion-context/pull/305.

**Root cause:** the `anthropic-beta` header is forwarded upstream (so the provider correctly serves 1M), but bili's *own* window resolution missed it. The client-facing instance sized its window at the model-table 200K, so 261K input read as 131% → `EMERGENCY T1`, when it was really only 26% of the 1M window the client negotiated.

**The fix** (`src/server.ts`):
- New `anthropicBetaContextWindow(headers)` parses the `anthropic-beta` header for `context-Nm` betas → `N × 1,000,000` (largest wins). Generalizes to future larger-context betas; rejects malformed/non-context tokens.
[...]

PR #304, opened 2026-08-27
fix: detect chained bili instances via x-bili-hop marker (#300)

ranxianglei:
## What

Fixes #300 (root cause of #292): two chained bili instances both ran the compression pipeline on the same request, double-compressing and corrupting session state.

## How

**Outbound marker** (`src/server.ts`)
- Each server gets a per-server `instanceId = randomUUID()`.
- Every request this instance *actually processed* is stamped `x-bili-hop: <instanceId>` — the main forward, preflight summarization calls, and compress-loop calls (the latter two share the `buildForwardTarget` headers).
- A **passthrough** forward does NOT claim processing, so a dumb-proxy bili in front of a processing one is unaffected; any inbound marker keeps propagating down the chain verbatim.

**Inbound detection** (`handle()`)
- If an inbound request already carries `x-bili-hop`, an upstream bili already processed it → skip ALL processing (no tool/tag injection, no acp-loop, no session state) and pass it through verbatim.
- Logs a prominent `[warn]` naming the upstream instance, distinguishing a foreign chain from a self-loop (upstream pointing back at itself).

**Why no false positives in single-instance use:** clients never send `x-bili-hop` — it only appears in bili's own forwards. Claude Code echoing back acp-tagged messages carries no such header, so normal turns are untouched.

## Tests

New `tests/e2e-bili-chain.test.ts` (2 tests):
1. **B in isolation with a foreign marker** — body reaches the LLM byte-identical (no tool/tag injection, no re-serialization), exactly one `[chain]` warn naming the upstream marker, marker propagated unchanged.
2. **Full A→B chain** — single-layer processing (A stamps, B warns + passes through), exactly one LLM request, the marker at the LLM is A's (not re-stamped by B).

## Pre-flight

- `npm run typecheck` — clean
- `npm test` — 693 pass (691 existing + 2 new), 0 fail
- `npm run build` — success

No dependency changes; `package.json` version untouched (content branch).

PR #303, opened 2026-08-27
fix: adopt complete terminal usage atomically in anthropic adapter (stitched-stream double-count)

ranxianglei:
## What

Fixes #299: after a compress re-request, the stream handed to a downstream bili is two rounds stitched (round1 `message_start` + final synthetic terminal). The anthropic adapter's per-field `> 0` guard refused the terminal's `cache_read_input_tokens: 0`, so the stale `cache_read` from round1's `message_start` survived and the total double-counted (142543 + 118663 = 261206 → false 131% EMERGENCY nudge + preflight compression, observed in #292).

### Changes

1. **Parse side (the fix)** — `src/loop/adapter-anthropic.ts` `message_delta` branch: when the usage object carries BOTH `input_tokens` (>0) AND `cache_read_input_tokens`, adopt the pair **atomically** (a 0 may overwrite the stale value). When fields are incomplete, keep the existing per-field `> 0` guard (relay-echo `input_tokens: 0` behavior unchanged).
2. **Generation side (contract pin)** — `buildTerminal` already always emits all three usage fields with the final round's true values (0 included); no code change, but a regression test now pins that contract.

The openai/responses wires are unaffected: their usage events replace the whole object (last-wins), no per-field merging.

## Tests

New in `tests/loop-adapters.test.ts`:
- **stitched stream (parse)**: round1 `message_start` (input=50000, cache_read=118663) + synthetic terminal (input=142543, cache_read=0) → recorded total input = **142543**, not 261206. Verified to fail on unfixed code (`actual: 261206`).
- **stitched terminal (generation)**: 2-round loop with proxy-tool re-request → the synthetic terminal in the output carries `input_tokens: 142543` + `cache_read_input_tokens: 0` (explicit 0) + `output_tokens`.
- Pre-existing relay-echo test (`input_tokens: 0` must NOT overwrite `message_start`) still passes — guard preserved for incomplete objects.

## Pre-flight

- `npm run typecheck` — clean
- `npm test` — 693/693 pass
- `npm run build` — success

PR #306, opened 2026-08-27
fix: preflight compress fails fast on upstream errors instead of forwarding over-window payloads (#301)

ranxianglei:
## What changed

When preflight (overflow) compression cannot proceed, the proxy previously gave up and **forwarded the original over-window payload unchanged** (`src/server.ts:1677` logged "forwarding as-is") — a guaranteed upstream 400, wasted quota, and client retry storms (log evidence in #292: 261K-token payload forwarded after both summary attempts 429'd).

Now it **fails fast** with a structured error instead:

- `503` + `Retry-After: 30` when the summary upstream returned 429 (rate limited — retryable)
- `502` for other upstream errors (5xx, network) or an exhausted compress budget (no compressible ranges left / nothing to fold)
- Body: `{ "error": { "type": "server_error", "code": "preflight_compress_failed", "message": "...", "retryable": true|false } }` — the message names the cause (e.g. `the summarization call was rate-limited by the upstream (HTTP 429)`) and states the payload was NOT forwarded.
- No response write when the client already disconnected (abort is detected, forward skipped).

**Forwarding as-is stays legal only when the payload's own estimate fits the window.** The trigger and the loop's fit check floor on `session.stats.lastInputTokens`, which can be stale (double-counted usage, #300) — the new `PreflightResult.payloadEstimate` (payload-only estimate, no floor) is the gate for the forward/fail-fast decision, so a fitting payload is never failed fast on a stale baseline (no false positive).

## Details

- `src/preflight.ts` — `PreflightResult` gains `payloadEstimate` + `failure` (`kind: "upstream" | "exhausted" | "aborted"`, HTTP status for upstream failures). Every loop break path now records its reason.
- `src/server.ts` — `preflightCompressIfNeeded` returns `Prepared | PreflightFailFast`; the call site writes the structured error and skips `forward()`. The "no compressible ranges" early-out gets the same fits-gate (forward if it actually fits, fail fast otherwise).
- Backoff already exists in `fetchWithRetry` (3 attempts, exponential 1.5s base, tunable via `BILI_REPLAY_RETRY_MAX` / `BILI_REPLAY_RETRY_BASE_MS`), so the optional separate-summary-model config from the issue is not included here.

## Tests (`tests/preflight-fail-fast.test.ts`, 3 new e2e)

1. **Overflow + summary upstream 429** → client gets structured 503 (`code: preflight_compress_failed`, message names 429), and **zero** stream forwards reach the upstream.
2. **Payload fits window (20k) + stale 300k baseline + preflight 429** → request still forwarded (200) — no false positive.
3. **Over-window payload with nothing compressible** (single 12k-token message, 10k window) → structured 502, no summarization call spent, no forward.

## Pre-flight

- `npm run typecheck` — clean
- `npm test` — 694/694 pass (691 existing + 3 new)
- `npm run build` — success

### DIFF
--- a/src/preflight.ts
+++ b/src/preflight.ts
@@ -44,9 +44,27 @@ export interface PreflightDeps {
     log: (level: string, msg: string) => void;
 }
 
+export type PreflightFailureKind = "upstream" | "exhausted" | "aborted";
+
+export interface PreflightFailure {
+    kind: PreflightFailureKind;
+    /** Upstream HTTP status when kind === "upstream" and the failure was an HTTP response. */
+    status?: number;
+    /** Human-readable cause (safe to surface to the client). */
+    detail: string;
+}
+
 export interface PreflightResult {
     compressedRanges: number;
     savedTokens: number;
+    /** Token estimate of the final (post-fold) payload, from the payload
+     *  itself — NOT floored on the session's lastInputTokens, which can be
+     *  stale (e.g. a double-counted usage report, #300). The caller uses it
+     *  to decide whether forwarding as-is is actually safe. */
+    payloadEstimate: number;
+    /** Why the loop stopped while the payload still overflows the window.
+     *  Undefined when the payload fits. */
+    failure?: PreflightFailure;
 }
 
 function refMaps(messages: CoreMessage[], state: Session["state"]): { refToIdx: Map<string, number>; idxToRef: Map<number, string> } {
@@ -198,9 +216,11 @@ async function summarizeRange(deps: PreflightDeps, content: string, startRef: st
     }
 }
 
+const ABORTED_FAILURE: PreflightFailure = { kind: "aborted", detail: "the client disconnected during preflight compression" };
+
 export async function preflightCompress(deps: PreflightDeps, messages: CoreMessage[]): Promise<PreflightResult> {
     const limit = deps.config.modelContextLimit;
-    const result: PreflightResult = { compressedRanges: 0, savedTokens: 0 };
+    const result: PreflightResult = { compressedRanges: 0, savedTokens: 0, payloadEstimate: estimateCoreMessages(messages) };
     if (limit <= 0) return result;
     const budget = Math.max(MIN_CHUNK_TOKENS, Math.floor(limit * CHUNK_FRACTION));
     // applyCompression rejects ranges below config.compress.minCompressRange
@@ -212,8 +232,12 @@ export async function preflightCompress(deps: PreflightDeps, messages: CoreMessa
     // overflows the smaller window.
     let currentTokens = deps.session.stats.lastInputTokens;
     let startTokens = -1;
+    let failure: PreflightFailure | undefined;
     for (let round = 0; round < MAX_PREFLIGHT_ROUNDS; round++) {
-        if (deps.signal?.aborted) break;
+        if (deps.signal?.aborted) {
+            failure = ABORTED_FAILURE;
+            break;
+        }
         // Re-run the pipeline each round: every successful compress renumbers
         // the surviving refs, so the previous round's range refs are stale.
         const turn = deps.core.processTurn({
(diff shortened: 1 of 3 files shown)


ranxianglei:

[bot] On it — reviewing PR #301 (`2026-08-28_preflight-fail-fast`). The branch is based on the current master tip (no rebase needed). Kicking off typecheck / test / build and reading through the `preflight.ts` + `server.ts` changes. Will post a full review shortly.

ranxianglei:

[bot] Full review of PR #301 (`2026-08-28_preflight-fail-fast`) — based on current master tip, no rebase needed.

**Pre-flight (ran all three, all green)**
- `npm run typecheck` — clean
- `npm test` — 694/694 pass (691 existing + 3 new)
- `npm run build` — success
- No `as any` / `@ts-ignore` / `console.error` in server-side modules; logging goes through `log(...)`.

**Core logic — sound.** I traced the full loop in `src/preflight.ts` and the gate in `src/server.ts`. The key design (trigger on the floored `max(lastInputTokens, payloadEstimate)`, but decide forward-vs-fail-fast on the payload-only `payloadEstimate`) is correct and strictly better than the previous always-forward behavior:
- `payloadEstimate <= currentTokens` always holds (`currentTokens = max(baseline, payloadEstimate)`), so whenever the caller fails fast (`payloadEstimate >= limit`), `currentTokens >= limit` and `result.failure` is guaranteed defined — the `?? "unknown reason"` at `server.ts:1750` is dead-but-harmless.
- `preflightCompress`'s own fit check is unchanged (still floors on `currentTokens`), so it still compresses aggressively until the *baseline* fits; the new gate only relaxes the caller's decision so a fitting payload is never failed fast on a stale baseline (#300). The partial compression is still persisted to session state even on fail-fast, so the next request starts from a smaller baseline — nice property.
- I walked every loop break path (abort, fit, no-ranges, unresolvable-range, upstream error, no-progress, budget-exhausted): each over-window exit records a `failure`, and `result.failure` is set iff `currentTokens >= limit`.
- Abort path is right: `respond: false` → call site skips both the write and `forward()`. The double guard (`outcome.respond && !res.headersSent && !res.destroyed`, with `!res.destroyed` re-evaluated at the call site) is robust to the client disconnecting between the `failFast()` computation and the write.
- The 3 new e2e tests match the implementation; I traced each (429→503+`retry-after:30`+zero forwards; stale-300k/fitting-13k+429→still 200; nothing-compressible→502, no summary call, no forward).
- Verified the PR's backoff claim: `REPLAY_MAX_ATTEMPTS = 3`, base 1500ms, `BILI_REPLAY_RETRY_MAX`/`_BASE_MS` tunable (`src/fetch-util.ts:112-127`).

**One thing to close before merge — missing CHANGELOG entry.** Every recent content fix/feature PR adds an entry under `[Unreleased]` → `### Fixes` (identity-binding, type-less-items, whitespace-drop, usage-credit all did). This is a notable behavioral change (fail-fast instead of forward-as-is) and has none. Suggested text, ready to drop in:

> **Preflight (overflow) compression fails fast on upstream errors instead of forwarding over-window payloads**: when proactive pre-forward compression (#247) can't bring an over-window payload under the window — the summarization call is rate-limited (429) or errors (5xx/network), or the compress budget is exhausted (nothing left to fold) — the proxy previously forwarded the original over-window payload as-is (a guaranteed upstream 400, wasted quota, retry storms; #292: a 261K-token payload forwarded after both summary attempts 429'd). It now answers with a structured error: `503` + `Retry-After: 30` for a rate-limited summary call (retryable), `502` for other upstream errors or an exhausted budget (not retryable), body `{ "error": { "type": "server_error", "code": "preflight_compress_failed", "message": "...", "retryable": bool } }` — the message names the cause and states the payload was NOT forwarded. No write when the client already disconnected. Forwarding as-is stays legal only when the payload's own estimate fits: the new `PreflightResult.payloadEstimate` (payload-only, no `lastInputTokens` floor) gates the decision, so a fitting payload is never failed fast on a stale baseline (#300).

Happy to push that to the branch if you want — just say the word (I'd rather not touch your branch unasked).

**Minor, non-blocking**
- The fail-fast message reports the pre-compress `tokenCount`, not the post-compress size — accurate (it really did exceed the window) but reads as the original size after partial compression. Cosmetic.
- `status: 0` sentinel in the abort path is never used (`respond: false`). Fine; omitting/null would read slightly cleaner.
- Optional test coverage: the abort path (client disconnect mid-preflight) and the non-429 upstream-error→502 path aren't directly exercised (502 is covered via the exhausted path). The non-429 case is easy to add; the abort one is fiddly.

**Verdict:** LGTM on the code. Only the CHANGELOG entry is a real pre-merge item. (Merging stays with you — I don't merge PRs.)

ranxianglei:
Model: glm-5.3

Rebased onto the #303 → #304 → #305 stack (ework is offline, taking over per maintainer request).

The original branch was based on pre-#311 master and conflicted with #304 in `server.ts` (`preflightCompressIfNeeded` signature: #304 added `instanceId: string,`, this PR changed the return type to `Promise<Prepared | PreflightFailFast>`). The rebase keeps BOTH — the parameter and the union return type. Call site auto-merges cleanly (`instanceId` threaded + `isPreflightFailFast` branch intact).

Validation on the rebased branch: typecheck ✓, **727/727 tests ✓**, build ✓ (includes all four PRs' tests: e2e-bili-chain, anthropic-beta-window, preflight-fail-fast, loop-adapters usage cases).

Merge order: #303 → #304 → #305 → #306 (all clean merges now).

PR #307, opened 2026-08-27
fix(logger): rotated log file orphan inode — reopen or degrade to stderr-only (#210)

### LINKED ISSUE #210: bug(logger): rotated bili.log orphan inode — writes vanish until restart

## Symptom

After log rotation (bili.log → bili.log.old at 10MB threshold), the running proxy's file descriptor points to the renamed inode (.old). New log lines are written to .old instead of the fresh bili.log. The fresh bili.log stays 0 bytes until the proxy is restarted.

## Root Cause

`src/logger.ts` opens a single `WriteStream` with `flags: "a"`. On rotation (`renameSync(file, file + ".old")`), the held fd becomes an orphan inode — writes silently go to the renamed file. The logger's error-detection (`s.on("error", ...)`) does NOT fire because the fd is still valid (just pointing to a different name).

## Impact

- Multiple restarts without exceeding 10MB: works fine (append to same file).
- After rotation: all subsequent writes go to .old; bili.log appears empty.
- Affects ALL startup modes (launcher, manual, test).

## Fix Direction

After rotation, explicitly close the old stream and reopen against the new file path. Or use `fs.watch` on the directory to detect the rename. Simplest: in the rotation branch of `openStream`, set `stream = undefined` so the next `getStream()` call reopens against the current path.

ranxianglei:
## What

Fixes #210: after `bili.log` is renamed to `bili.log.old` (internal 10MB rotation, logrotate, or a manual rename), the logger's held `WriteStream` fd points at the renamed inode and every subsequent line silently lands in `.old` while the fresh `bili.log` sits at 0 bytes — until a restart. No `error` event ever fires because the orphaned fd is still valid, so the previous error/writable-based recovery never triggered.

## How (per triage on #210)

- **Inode compare on every write**: `isOrphaned()` compares the fd's inode (`fstat`) with the current path's inode (`stat`); on divergence the file was renamed/replaced out from under us. The fd is captured via the stream's `open` event (the `WriteStream` type exposes no `fd` property).
- **Explicit close + reopen**: the old stream is closed with `end()` first (drains its buffer into the rotated file — no lost lines), `stream = undefined`, and the next `getStream()` reopens against the current path.
- **Reopen failure never crashes**: if a (re)open fails (disk full, perms, path clobbered), logging degrades to stderr-only with a single `[warn]` (one per degradation episode, reset on successful reopen) and the proxy keeps serving. Covers both runtime reopens and the startup `configureLogger` open (previously an unguarded `openStream` that could crash startup).

## Tests (tests/logger-rotation.test.ts, +3)

1. **External rename simulation** (the required repro): write → flush → `renameSync` → write more → assert the new `bili.log` receives the post-rename lines and `.old` size is frozen.
2. **Reopen failure**: clobber the log path so `mkdirSync` throws ENOTDIR → assert no crash, exactly one `[warn]` (via `setLogCapture`), and subsequent `log()` calls keep working stderr-only.
3. **Internal 10MB rotation regression**: single 11MB line + post-rotation line → assert the post-rotation line lands in the fresh file and `.old` holds the rotated file.

## Pre-flight

- `npm run typecheck` — clean
- `npm test` — 701/701 pass (698 + 3 new)
- `npm run build` — tsup success, dist/index.js 2.51MB

CHANGELOG entry added under Unreleased → Fixes.


ranxianglei:

[bot] Picked up. PR #307 (branch `2026-08-28_fix-logger-rotation-orphan`, commit 380e323) is already pushed with the logger fix + 3 new tests. I'm now verifying it: reviewing the diff, re-running pre-flight (typecheck / test / build) on the branch, and confirming the new tests actually catch the orphan-inode bug against the old logger. Will report back with results.

ranxianglei:

[bot] Verification complete — PR #307 checks out.

**Pre-flight (re-run on branch `2026-08-28_fix-logger-rotation-orphan`)**
- `npm run typecheck` — clean
- `npm test` — 701/701 pass (698 + 3 new)
- `npm run build` — tsup success, dist/index.js 2.51MB

**Bug-repro check** (ran the 3 new tests against the OLD `src/logger.ts` from master):
- Test 1 (external rename) — **fails** on old code (lines drift to `.old`, fresh file never created) ✓ catches the bug
- Test 2 (reopen failure) — **fails** on old code (no warn path existed) ✓ catches the bug
- Test 3 (internal 10MB rotation) — passes on both (regression guard for pre-existing behavior, as intended)

**Review notes** — implementation matches the triage approach on #210:
- `isOrphaned()` compares `fstat(fd)` vs `stat(path)` on `ino` + `dev` per write; open-in-flight (`streamFd` unset) is not treated as orphaned; stat/fstat throw (path gone/EBADF) → orphaned.
- `end()` before reopen drains the buffer into the rotated file — lossless (a `destroy()` would have dropped buffered lines).
- Exactly one `[warn]` per degradation episode, reset on successful reopen; the warn bypasses `log()` itself (no recursion) and mirrors through `setLogCapture`.
- `configureLogger` now opens via the guarded `getStream()`, so a failed startup open degrades to stderr-only instead of crashing.
- One deliberate perf trade-off: 2 sync syscalls per log line (fstat+stat) — acceptable at proxy log volumes.

**Follow-up commit pushed:** `07c8fd1` — devlog entry `devlog/2026-08-28_fix-logger-rotation-orphan/` (REQ.md + WORKLOG.md), which the original commit was missing per repo convention.

PR is ready for merge (human-only — I can't merge it). Merging will auto-close #210 via the `Fixes #210` reference.

PR #308, opened 2026-08-27
feat: #189 compress observability + gated staged/tail-bias steering (smooth transition)

ranxianglei:
## What
Lands request 2 (「平滑压缩过渡」/ smooth compression transition) from #189 — the structural fix for the GLM 3007 (`captcha verify failed`) that fires on the post-compress acp-loop replay request after a large context rewrite (131K→57K). Built on the already-shipped retry (`7fae454`), which makes 3007 recoverable; this lowers the *rate*.

Three parts:

1. **Observability (always on)** — `applyRanges` now records the last successful compress on the session (shrink ratio, fold point, blocks, tokens) and logs an `[acp-compress-obs] shrink N% (~X/Y tok) foldPoint=Z blocks=N` line per compress. This is the 「观测先行」 step: data before validating a fix.

2. **3007 correlation** — the replay/preflight retry callbacks (`src/loop/core.ts`, `src/compress-loop-responses.ts`, `src/preflight.ts`) now append `[after compress: shrink N% foldPoint=X blocks=N ~Ntok]` to the risk-control retry log, so a transient upstream rejection is correlated with the rewrite that preceded it.

3. **Staged compression + prefix survival (gated)** — behind `BILI_MAX_SHRINK_PER_COMPRESS` (a fraction in (0,1]; unset/out-of-range = off, legacy behavior). When a single compress shrinks context more than the configured fraction, the tool result steers the model toward a *smaller, TAIL-biased* range (compress the most recent large content, keep the stable prefix `m00001..foldPoint` intact), and the nudge text gains a smooth-transition guidance block. Rationale: the sharp request-shape change is what trips provider risk-control, and a late fold point keeps the prefix cache alive.

### Design notes
- **Staged compression is a soft steer, not a hard range-split.** A hard split would force the model's single full-range summary onto a sub-range (summary/range mismatch → quality loss). Since the shipped retry already makes 3007 recoverable (correctness), this is an optimization to cut the rate, so steering the model's *next* range choice is appropriate and safe.
- **Tail-bias is appended to the nudge text** (the moment the model picks the range) via a proxy-side append — it does NOT touch the kernel's load-bearing prompts, so there's no summary-quality regression risk.
- **Env var, not a file setting**, matching the existing retry-config pattern (`BILI_REPLAY_RETRY_MAX` / `BILI_REPLAY_RETRY_BASE_MS` in `src/fetch-util.ts`) and avoiding threading a new field through `RewriteCtx` + the three-level compress-config merge. One knob acts as both the switch and the threshold.
- `session.lastCompress` is **in-memory only** (not persisted) — a fresh process has none, which is correct (the correlation is only meaningful within a process that just compressed).

## Pre-flight
- `npm run typecheck` — clean
- `npm test` — 708 pass / 0 fail (698 prior + 10 new in `tests/compress-observability.test.ts`)
- `npm run build` — success

## Verify (reporter)
1. Update to the release containing this (≥ v0.1.60).
2. Set `BILI_MAX_SHRINK_PER_COMPRESS=0.3` (or your target per-compress shrink).
3. Run your 5×3007 scenario; watch the `[acp-compress-obs]` and the retry logs (now correlated with the preceding compress). Confirm the 3007 rate drops and the model starts choosing smaller, tail-biased ranges.

Issue stays open until the rate is thoroughly down (owner instruction).

### DIFF
--- a/src/compress-loop-responses.ts
+++ b/src/compress-loop-responses.ts
@@ -5,7 +5,7 @@ import {
     type Config,
     type CoreMessage,
 } from "acp-kernel";
-import type { Session } from "./session.js";
+import { lastCompressSuffix, type Session } from "./session.js";
 import { parseCompressInput, PROXY_TOOL_NAMES, MUTATING_PROXY_TOOLS, COMPRESS_TOOL_NAME, ACP_TEXT_OPEN, ACP_TEXT_CLOSE } from "./compress-tool.js";
 import { log as loggerLog } from "./logger.js";
 import { applyRanges } from "./stream.js";
@@ -219,8 +219,10 @@ export async function compressLoopResponsesJson(
             body: JSON.stringify(requestBody),
             ...(ctx.proxyUrl ? { dispatcher: proxyDispatcher(ctx.proxyUrl) } : {}),
         }, undefined, undefined, (info) => {
-            ctx.log(`[acp-proxy: responses upstream rejected replay (HTTP ${info.status}: ${info.detail.slice(0, 120)}); likely provider risk-control — retrying in ${info.delayMs}ms (attempt ${info.attempt}/${info.maxAttempts})]`);
-            loggerLog("warn", `[acp-compress-responses] upstream rejected replay (HTTP ${info.status}); retrying in ${info.delayMs}ms (attempt ${info.attempt}/${info.maxAttempts})`);
+            // #189: correlate the rejection with the rewrite that preceded it.
+            const lc = lastCompressSuffix(ctx.session.lastCompress);
+            ctx.log(`[acp-proxy: responses upstream rejected replay (HTTP ${info.status}: ${info.detail.slice(0, 120)}); likely provider risk-control — retrying in ${info.delayMs}ms (attempt ${info.attempt}/${info.maxAttempts})${lc}]`);
+            loggerLog("warn", `[acp-compress-responses] upstream rejected replay (HTTP ${info.status}); retrying in ${info.delayMs}ms (attempt ${info.attempt}/${info.maxAttempts})${lc}`);
         }).catch((e) => {
             if (e instanceof UpstreamHttpError) {
                 const suffix = e.attempts > 1 ? ` after ${e.attempts} attempt(s)` : "";
--- a/src/compress-tool.ts
+++ b/src/compress-tool.ts
@@ -10,6 +10,7 @@
  */
 import { parseCompressArgs } from "acp-kernel";
 import { log as loggerLog } from "./logger.js";
+import { maxShrinkPerCompress } from "./fetch-util.js";
 
 export {
     COMPRESS_TOOL_NAME,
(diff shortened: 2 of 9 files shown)


ranxianglei:

[bot] Reviewed `2026-08-28_smooth-compress-transition` (1 commit `a8685e1`, 9 files, +298/−13). Branch is clean master + 1 commit (no divergence); `package.json` version untouched, as required for a content branch.

**Pre-flight (ran locally on the branch):**
- `npm run typecheck` — clean
- `npm test` — 708 pass / 0 fail (698 prior + 10 new in `tests/compress-observability.test.ts`), matching the issue
- `npm run build` — success (dist/index.js 2.51 MB, acp-kernel inlined)

**Code review — all three parts land as described:**
1. **Observability**: `applyRanges` (`src/stream.ts:274-282`) records `session.lastCompress` only on the success path (the `blocksCreated === 0` failure returns early at `src/stream.ts:268-272`), and reads `preContext = stats.lastInputTokens` *before* the credit netting at `src/stream.ts:301-302`, so the shrink ratio is measured against pre-compress context. Logs `[acp-compress-obs] shrink N% (~X/Y tok) foldPoint=Z blocks=N` per compress. Fold point = earliest `startRef` across all ranges, order-independent (covered by a test).
2. **3007 correlation**: all three retry callbacks append `lastCompressSuffix(ctx/deps.session.lastCompress)`: `src/loop/core.ts:444-448`, `src/compress-loop-responses.ts:222-225`, `src/preflight.ts:178-181`. Verified `session` is in scope in each (`PreflightDeps.session` at `src/preflight.ts:35`, etc.). Empty string when no compress recorded, so no log change for fresh sessions.
3. **Gated staged/tail-bias**: `maxShrinkPerCompress()` (`src/fetch-util.ts:134-138`) parses `BILI_MAX_SHRINK_PER_COMPRESS` as (0,1], unset/out-of-range/non-numeric → `undefined` (off, legacy behavior) — read per call, matching the `BILI_REPLAY_RETRY_*` pattern. Two steering paths, both gated: the tool-result note in `applyRanges` (`src/stream.ts:286-293`, fires only when `shrinkRatio > max`) and `withStagedCompressGuidance` (`src/compress-tool.ts:64-68`) appended at all three nudge sites (`src/server.ts:1106, 1204, 1342`) — I grepped `renderNudgeText` and those are the only call sites. Kernel prompts untouched, as designed.

**In-memory-only claim verified**: `buildRecord` (`src/persist.ts:330-343`) uses an explicit field list that omits `lastCompress`, so it never hits disk.

- `refNum` (`src/stream.ts:224-226`) strips all non-digits, so a hypothetical mixed `mNNNNN`/`bN` range list would order across ref namespaces numerically. Compress ranges in practice are `mNNNNN` context refs, and foldPoint is an observability label only — no action needed.
- `shrinkRatio` divides by the upstream-reported `lastInputTokens` (an approximation, hence the `~` in the log). If that stat is stale/0, ratio = 0 and steering conservatively doesn't fire. Fine.
- No `as any` / `@ts-ignore` / `console.error`; server-side logging goes through `loggerLog`.

PR is ready for a human to merge. After merge + release (≥ v0.1.60), the open item is your Verify step: run the 5×3007 scenario with `BILI_MAX_SHRINK_PER_COMPRESS=0.3` and confirm the rate drops and the model starts preferring smaller tail-biased ranges — I can't exercise that scenario from here.

PR #322, opened 2026-08-28
feat: align launcher auto-compact budgets with bili window (#321 PR-D)

ranxianglei:
**Model: qwen3.8-27b**

PR-D of #321 — launcher budget coordination (codex + claude). Follow-up PRs: PR-C (semantic rebase gate), E1 (codex model-table snapshot + min() alignment), E2 (conditional interception forgery).

## Problem

codex's auto-compact budget is keyed off the window CODEX believes the model has (its bundled model table, `models-manager/models.json` — bili has no say in it), while bili's ACP compression is keyed off bili's own window resolution. Two uncoordinated budgets (#292): when bili's window exceeds codex's perception, codex's ledger (server-reported usage, which bili sees) crosses codex's ~90% threshold first and fires native compaction ahead of ACP.

claude-code has the same shape (its own model table + 200k unknown-model fallback) but a single env knob instead of `-c` config.

## Changes

**codex** (`src/launcher.ts`):
- `resolveCodexBudgetArgs()` — injects `-c model_context_window=<W> -c model_auto_compact_token_limit=<W>` into the spawned codex CLI (MITM mode only; direct-URL mode has no compression, skipped). W = bili's effective window for codex's configured model, resolved by `resolveLauncherWindow()`: per-route per-model config declaration → built-in `CONTEXT_LIMIT_TABLE` → models.dev registry (snapshot-first, offline-safe) — the same chain the proxy applies, minus the per-request-only sources (beta header / plugin report / launcher window).
- Effect: codex's auto-compact threshold becomes 90%×W; ACP (≈55%×W) always fires first; codex's LOCAL compaction (benign for bili — same-session truncation the kernel deactivates by message id) only backstops when ACP fails, before codex's 95% hard cap. codex clamps both values to its own `max_context_window`, so an over-generous W degrades to codex's own perception instead of overshooting.
- No injection when: no model configured; user already set `model_context_window` in `~/.codex/config.toml` (bili's proxy uses exactly that value as its launcher window — already self-aligned); or bili resolves no window. A user-set `model_auto_compact_token_limit` is honored, not overridden.
- `parseCodexToml` now stores top-level `model` and `model_auto_compact_token_limit` (`CodexConfig.model` / `.contextWindow` / `.autoCompactLimit`).

**claude** (`src/launcher.ts` + `src/client-config.ts`):
- `resolveClaudeBudgetEnv()` — injects `CLAUDE_CODE_AUTO_COMPACT_WINDOW=<W>` into the spawned claude env (MITM and direct-URL modes — both route through the proxy). The knob outranks claude's settings and is clamped DOWN to the model window claude perceives (never up), so injection is always safe: tightens claude's threshold to bili's budget when bili's window is smaller, no-op when larger.
- No injection when: no model resolvable (`ANTHROPIC_MODEL` env → settings.json `model`); user already set an explicit auto-compact window (settings `autoCompactWindow` or `env.CLAUDE_CODE_AUTO_COMPACT_WINDOW`, or shell-exported env var); or bili resolves no window.
- `readClaudeSettings` now parses `model` (env block beats same-file top-level) and `autoCompactWindow` (both settings forms).
- `CLAUDE_CODE_MAX_CONTEXT_TOKENS` deliberately NOT used — verified in claude 2.1.247 it is only read under `DISABLE_COMPACT`.

## Notes

- claude model can change mid-session (`/model`); the env value is fixed at spawn — the down-only clamp makes a stale value safe (worst case: claude compacts earlier than needed).
- MITM-only (non-launcher) claude users: manual alignment via settings.json `autoCompactWindow` / `/autocompact` — README docs to follow with the E-series.
- Source-verified against openai/codex: `config/src/config_toml.rs` (top-level `model_context_window` / `model_auto_compact_token_limit` keys), `core/src/session/context_window.rs` (Total scope: auto-compact limit = model_info value, 90% default; hard cap = 95%×window), `models-manager/src/model_info.rs:26-31` (config window override clamped to `max_context_window`), `model-provider-info/src/lib.rs:305` (default base URL `https://api.openai.com/v1`).
- ⚠️ E1 boundary correction (affects the follow-up, reported in #321): codex does NOT treat unknown models as "no perception" — `model_info_from_slug` (models-manager/src/model_info.rs:171-172) falls back to `context_window = max_context_window = 272000`, so unknown models auto-compact at ~244.8k. E1 must treat not-in-table as codex-perception=272k, not unlimited.

## Pre-flight

- `npm run typecheck` — clean
- `npm test` — 751 pass / 0 fail (11 new: value-chain unit tests for `resolveLauncherWindow` / `resolveCodexBudgetArgs` / `resolveClaudeBudgetEnv` / `codexUpstreamUrl`, `parseCodexToml` / `readClaudeSettings` parsing, and two `runLaunch` integration tests asserting the injected args/env end-to-end)
- `npm run build` — success

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -449,6 +449,20 @@ Run `codex login` as usual; the OAuth token travels in the `Authorization` heade
 "baseUrl": "http://localhost:8787/bili/https://api.anthropic.com"
 ```
 
+**Claude Code** — set the `ANTHROPIC_BASE_URL` env var to the `/bili/` URL. (claude's undici fetch ignores `HTTPS_PROXY`, so the `/bili/` URL form is the only manual option — cert MITM cannot intercept it.)
+
+```bash
+export ANTHROPIC_BASE_URL="http://localhost:8787/bili/https://api.anthropic.com"
+```
+
+> **Auto-compact alignment (manual mode only).** The `bili claude` launcher automatically sets `CLAUDE_CODE_AUTO_COMPACT_WINDOW` to bili's effective window for your model, so claude's own auto-compact threshold lines up with bili's compression budget. In manual `/bili/` mode you must do this yourself — otherwise claude may run its own local auto-compact (a "summarize the conversation" turn) on a threshold that doesn't match bili's window. That is usually harmless (same session-id, so bili re-derives state from the truncation) but noisier than needed. Set it to bili's effective window for your model:
+>
+> ```bash
+> export CLAUDE_CODE_AUTO_COMPACT_WINDOW=<bili effective window in tokens>
+> ```
+>
+> claude clamps this value **down** to the window it perceives for the model (never up), so over-setting is safe. You can also set it persistently via claude's settings (`autoCompactWindow`).
+
 **Other API-key clients** (Cursor / Aider / Continue …) — wherever the upstream URL is configured, prepend `http://localhost:8787/bili/`. Nothing else changes.
 
 The `/bili/` prefix doubles as a **self-detection signal**: billion-context client extensions (billion-context-pi / opencode-acp) recognize it in their own baseUrl and self-disable, so you never get double compression.
--- a/CONFIGURATION.zh-CN.md
+++ b/CONFIGURATION.zh-CN.md
@@ -449,6 +449,20 @@ openai_base_url = "http://localhost:8787/bili/https://chatgpt.com/backend-api/co
 "baseUrl": "http://localhost:8787/bili/https://api.anthropic.com"
 ```
 
+**Claude Code** —— 把 `ANTHROPIC_BASE_URL` 环境变量设成 `/bili/` URL。（claude 的 undici fetch 忽略 `HTTPS_PROXY`，所以 `/bili/` URL 形式是唯一的手动方式 —— 证书 MITM 拦不到它。）
+
+```bash
+export ANTHROPIC_BASE_URL="http://localhost:8787/bili/https://api.anthropic.com"
+```
+
+> **自动压缩对齐（仅手动模式）。** `bili claude` launcher 会自动把 `CLAUDE_CODE_AUTO_COMPACT_WINDOW` 设成 bili 对你模型的有效窗口，让 claude 自己的自动压缩阈值与 bili 的压缩预算对齐。手动 `/bili/` 模式下需要你自己设 —— 否则 claude 可能在与 bili 窗口不一致的阈值上跑它自己的本地自动压缩（一次“总结对话”轮次）。这通常无害（同一 session-id，bili 会从截断中重新推导状态），但比必要的更吵。把它设成 bili 对你模型的有效窗口：
+>
+> ```bash
+> export CLAUDE_CODE_AUTO_COMPACT_WINDOW=<bili 有效窗口 token 数>
+> ```
+>
+> claude 会把这个值**向下**钳制到它自己感知的模型窗口（不会向上），所以设大了是安全的。也可以持久化到 claude 的 settings（`autoCompactWindow`）里。
+
 **其他 API-key 客户端（Cursor / Aider / Continue ……）** —— 只要配置了上游 URL，前面加 `http://localhost:8787/bili/` 就行，其他都不用改。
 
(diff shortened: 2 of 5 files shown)


ranxianglei:

[bot] On it — reviewing PR-D (launcher budget coordination, codex + claude). Pulling branch `2026-08-28_codex-budget-alignment`, running pre-flight (typecheck / test / build), and auditing `resolveCodexBudgetArgs` / `resolveClaudeBudgetEnv` / `resolveLauncherWindow` against the described value-chain. Will report findings here.

ranxianglei:

[bot] **Review complete — LGTM, no blockers.** (Human merges; I won't.)

**Pre-flight (re-ran on `2026-08-28_codex-budget-alignment` @ 8adeef1):**
- `npm run typecheck` — clean
- `npm test` — **751 pass / 0 fail** (matches the PR claim exactly)
- `npm run build` — success
- Branch is based on current master tip (`744e658`); dry-run merge shows **no conflicts**.

**Correctness audit (all verified against source):**
- `resolveLauncherWindow` value chain is exactly `resolveConfiguredContextLimit → lookupContextLimit → contextFromRegistry` (src/launcher.ts:354-371). Signatures match `src/config.ts:192/139/251` and `src/registry.ts:225`. Correctly excludes the per-request-only sources.
- `resolveCodexBudgetArgs` (src/launcher.ts:407-420): skip on no-model / user-set `model_context_window` (self-aligned) / unresolvable; else injects both `-c` args with `limit = clientAutoCompactLimit ?? window`. Gated to **MITM only** — the direct-URL branch (src/launcher.ts:1540-1542) is untouched, which is right since codex direct-URL traffic bypasses the proxy (no ACP to coordinate).
- `resolveClaudeBudgetEnv` (src/launcher.ts:440-452): skip on no-model / settings `autoCompactWindow` / shell `CLAUDE_CODE_AUTO_COMPACT_WINDOW` / unresolvable; else injects the env. Applied in **both** MITM and direct-URL (src/launcher.ts:1564-1574) — correct, both route through the proxy.
- `parseCodexToml` / `readClaudeSettings` (src/client-config.ts): env-block-beats-top-level precedence for claude `model` and both `autoCompactWindow` forms are right.
- Tests: 11 new, covering the value chain + 2 `runLaunch` integration tests asserting the injected args/env end-to-end. Good coverage.

**One thing I chased down (not a bug):** whether excluding the `launcherWindow` source could make the injected budget diverge from the window the proxy actually uses for ACP (server.ts:784-789 ranks `launcherWindow` *above* config/table/registry). It can't, for these two clients: `collectModelWindows` (src/client-config.ts:573-584) pulls codex windows only from the top-level `model`+`model_context_window` pair — which is precisely the value that trips the skip-injection path (`clientWindow`) — and claude isn't in `collectModelWindows` at all. So the budget and the ACP window stay consistent.

**Non-blocking notes (worth the radar, none block merge):**
1. **Claude down-clamp is a version-dependent external guarantee.** The injected `CLAUDE_CODE_AUTO_COMPACT_WINDOW=<W>` has no bili-side clamp — safety rests entirely on claude clamping W *down* to its perceived model window. If W > claude's perception (a model bili's registry lists larger than claude's 200k fallback) and a future claude drops the down-clamp, claude would believe it has a W-sized budget and hit its real hard cap before auto-compact. You've verified it in 2.1.247 and documented it in CONFIGURATION.md — good. Suggest a re-verify note on claude major updates (there's no clean bili-side fix; unlike codex's two-knob clamp, claude exposes only this one knob).
2. **Spawn-time registry fetch for untabled models (minor).** `resolveLauncherWindow → contextFromRegistry → loadRegistry` (src/registry.ts:128) can trigger a bounded models.dev fetch at spawn when the model is in neither the config routes nor the built-in table and the disk cache is stale. Adds a brief (timeout-bounded) startup delay to `bili codex`/`bili claude` for such models + one extra fetch in the parent process. Tabled models short-circuit via `lookupContextLimit` — zero impact. Acceptable.
3. **Alignment is inherently partial when bili's window ≫ client's max (by design).** codex clamps the injected `model_context_window` to its own `max_context_window`, so for a very large bili window the injected value degrades to codex's perception and native compaction can still lead ACP. You already call this out ("over-generous W degrades to codex's own perception") — this is exactly the gap E1 (model-table snapshot + min() alignment) closes. Not a PR-D concern.

Ship it. 👍

PR #323, opened 2026-08-28
fix: gate native-compaction rebase on semantic terminal state (#321 PR-C)

ranxianglei:
**Model: qwen3.8-27b**

PR-C of #321 — the semantic rebase gate (adopts #249's response-observer design, which was never merged). Follows PR-D (#322); precedes E1/E2.

## Problem

Both `markNativeCompactionBoundary` call sites fired on ANY HTTP 2xx upstream response. When a native compaction request actually FAILED (stream cut before `response.completed`, `response.failed`/`response.incomplete`, empty body, 2xx error JSON), codex kept its history — but bili scheduled a rebase anyway: on the next turn it reset its compression state (blocks cleared, usage zeroed) on top of a compaction that never happened. That is the #249 false-rebase bug: discarded valid blocks + a ledger that no longer matches the client's real context.

A second gap: the trigger form (plain `/responses` ending in `compaction_trigger`) had NO rebase marker at all — it flowed through the rewriter path (`compressInjected: true`), which never marks the boundary, even though a completed trigger response IS a client-side history swap. It also had bili's compress prompt/tools injected into the very request whose job is to produce the compaction.

## Changes

**`src/stream-terminal.ts` (new)** — terminal-state observer for Responses responses:
- SSE: `response.completed` → `completed`; `response.failed` / `response.incomplete` → `failed`; stream ends without a terminal event (truncation) → `unknown`. Checks the `event:` line, with a `data:` JSON `type` fallback for relays that strip `event:` lines. CRLF-safe, chunk-boundary-safe; cancels its tee branch at the first terminal event so it never backpressures the client branch.
- JSON (non-stream): `output` array → `completed`; `error` field → `failed`; unparseable / non-object / empty / >16MB → `unknown`.
- `unknown` is the safe default: no rebase; the next request self-heals via kernel message-id deactivation.

**`src/server.ts`**:
- Both rebase call sites now gate on observed `terminal === completed`. The 2xx empty-body path no longer rebases (logs a warn instead).
- Trigger-path hardening in `prepareResponses`: a trailing `compaction_trigger` is detected BEFORE injection → no compress system prompt, no compress tools, `renderTags: "none"`, `compressInjected: false` (response flows through plain passthrough instead of the rewriter, so the compaction stream reaches codex byte-exact), and `resetAfterSuccess: true` so its terminal state gates the rebase. The nudge skip now keys off the same pre-injection detection (replacing the post-injection `triggerFinal` check).

## Behavior matrix (all e2e-tested through the proxy)

| Response | Rebase? |
|---|---|
| SSE `response.completed` (trigger form) | ✅ yes |
| SSE `response.failed` / `response.incomplete` | ❌ no |
| SSE truncated (no terminal event) | ❌ no |
| 2xx JSON with `output` (endpoint form) | ✅ yes |
| 2xx JSON with `error` | ❌ no |
| 2xx empty body | ❌ no |
| non-2xx | ❌ no (unchanged) |

## Tests

- 16 new unit tests (`tests/stream-terminal.test.ts`): every terminal state, data-line fallback, CRLF/chunk-split, terminal-wins, non-JSON data lines, JSON shapes, oversized body.
- New e2e (`tests/native-compact-terminal.test.ts`): the full matrix above + trigger-forwarding assertions (no compress prompt/tools in the forwarded body, client instructions preserved, trigger stays final) + reconcile-on-next-turn (blocks cleared, marker consumed).
- `tests/compaction-trigger-finality.test.ts`: the nudge control now runs on a separate session — a completed trigger response rebases the compacting session (usage reset), which legitimately disarms its nudge; the control gets a warmup turn to arm usage.
- Pre-flight: `npm run typecheck` clean; `npm test` 757/757 pass (three-protocol regression unchanged); `npm run build` success.

## Notes

- This is the correctness backstop for E2's pass-through path (when ACP is unhealthy, native compaction passes through and this gate decides the rebase) and for all non-codex Responses clients.
- With E2's interception mode eventually defaulting, the boundary marker retires for codex (deterministic handoff) — this gate stays as the safety net for the pass mode and other clients.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -55,6 +55,7 @@ import { runCompressLoop, pickAdapter } from "./loop/index.js";
 import { sanitizeResponsesInputIds, dropWhitespaceResponsesMessages, normalizeResponsesMessageItems } from "./loop/adapter-responses.js";
 import { rewriteOpenaiJsonResponse } from "./stream-openai.js";
 import { rewriteResponsesJsonResponse } from "./stream-responses.js";
+import { observeResponsesTerminalState } from "./stream-terminal.js";
 import { emitStreamError } from "./stream-error.js";
 import { affinityToken, clientConversationHeader, preferPromptCacheKeyIdentity, type ConversationIdentity } from "./session-id.js";
 import { prefixAffinity, type AnonymousAffinity } from "./prefix-affinity.js";
@@ -1401,6 +1402,12 @@ function prepareResponses(
 
     const shouldInject = opts.compress.injectTool;
     const injectTools = shouldInject && !pluginMode;
+    // Codex native remote-compact request: no compress prompt/tools (the model
+    // produces the compaction itself), no acp tags, plain passthrough so the
+    // response terminal state can gate the rebase marker.
+    const isCompactionTrigger = Array.isArray(parsed.input)
+        && parsed.input.length > 0
+        && parsed.input[parsed.input.length - 1]!.type === "compaction_trigger";
     // Route config is keyed by the upstream THIS request goes to (#286: a
     // session can outlive its first relay — session.meta.upstreamOrigin is
     // first-wins and would silently ignore the new relay's route settings).
@@ -1416,7 +1423,7 @@ function prepareResponses(
             log("info", `[${sessionId}] input items: ${Array.isArray(parsed.input) ? parsed.input.map((i: ResponseInputItem) => i.type).join(",") : "(string)"}`);
         }
         const tokenCount = session.stats.lastInputTokens;
-        const turn = core.processTurn({ messages: msgs, state: session.state, config, tokenCount, renderTags: process.env.ACP_RENDER_NONE ? "none" : "text-only" });
+        const turn = core.processTurn({ messages: msgs, state: session.state, config, tokenCount, renderTags: process.env.ACP_RENDER_NONE || isCompactionTrigger ? "none" : "text-only" });
         session.state = turn.state;
         // The fold from last turn's compress has now materialized in state —
         // future usage reports are post-fold reality, drop the credit.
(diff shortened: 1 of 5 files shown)


ranxianglei:

[bot] Verified PR #323 (branch `2026-08-28_semantic-rebase-gate` @ `8d27bca`, single commit on current master `744e658`). All pre-flight checks pass locally:

- **typecheck**: clean (`tsc --noEmit --project tsconfig.build.json`)
- **test**: 757/757 pass (includes the new `tests/stream-terminal.test.ts` — 16 unit cases — and `tests/native-compact-terminal.test.ts` e2e)
- **build**: success (`dist/index.js` 2.52 MB, acp-kernel inlined)

Implementation review — matches the description:

1. **`src/stream-terminal.ts`** (new): `observeResponsesTerminalState(stream, isStream)` → `completed | failed | unknown`. SSE path checks the `event:` line first with a `data:` JSON `type` fallback, is CRLF/chunk-boundary-safe, and cancels its tee branch at the first terminal event (client branch keeps draining). JSON path: `error` → failed (checked first), `output` array → completed; unparseable/non-object/empty/>16 MB → unknown. `unknown` is the safe default — no rebase, self-heals via kernel message-id deactivation on the next turn.
2. **`src/server.ts`**: both `markNativeCompactionBoundary` call sites now gate on `terminal === "completed"`; 2xx-with-no-body logs a warn and does not schedule a rebase.
3. **Trigger hardening**: `prepareResponses` detects a trailing `compaction_trigger` *before* injection — no compress prompt, no compress tool, `renderTags: "none"`, `compressInjected: false` (plain passthrough, byte-exact), `resetAfterSuccess: true`. Nudge skip now keys off the same pre-injection detection.
4. **Invariant**: `resetAfterSuccess` is only ever true when `compressInjected: false`, and the rewriter path requires `compressInjected: true` — so the rewriter path can never bypass the gate.
5. **Tests**: e2e covers the full matrix (SSE completed/failed/truncated, JSON output/error, empty body) plus trigger-forwarding assertions (trigger stays final, client instructions preserved, no compress prompt/tools) and reconcile-on-next-turn (marker consumed, blocks cleared). `compaction-trigger-finality.test.ts` control correctly moved to a separate session with a warmup turn, since a completed trigger response now legitimately rebases the compacting session.

No `version` change on the branch (correct for a content branch). No problems found. PR is open and built by CI (`billion-context@pr-323`); merge is yours: https://github.com/ranxianglei/billion-context/pull/323

PR #324, opened 2026-08-28
feat: clamp effective window to codex's model perception for codex clients (#321 PR-E1)

ranxianglei:
**Model: qwen3.8-27b**

## What

PR-E1 of #321 (budget coordination for anonymous MITM codex clients): when the request client is codex, bili's effective window is capped at codex's own perception of that model — `min(bili window, codex perception)` — so ACP always compresses before codex's native auto-compact can fire.

## Why

codex's ledger base is the server-reported usage (which bili controls), but its auto-compact threshold comes from codex's own model table (`auto_compact_token_limit` = 90% of the model's `resolved_context_window`), with zero coordination with bili (#292's root cause). If bili's budget exceeds what codex believes, codex's native compaction fires first. Capping the effective window at codex's perception makes the budgets coincide: ACP (~55%) always goes first; if ACP fails, codex backstops at 90% before its 95% hard cap (local compaction for custom-provider users — benign for bili).

## Boundary correction vs the issue spec

The issue's E1 boundary clause ("模型不在 codex 表内 → codex 无认知 → 永不自动压缩") does not hold: codex's `model_info_from_slug` fallback gives unknown models `context_window = max_context_window = 272000` (models-manager/src/model_info.rs:171-172), i.e. unknown models auto-compact at ~244.8k — exactly the "~244k" trigger observed in #292. This PR treats not-in-table as **codex perception = 272000** (not unlimited), which is what fixes the #292 class. Already flagged in the issue thread.

## Changes

- `src/codex-models-snapshot.json` — bundled codex model table (10 models) generated from `openai/codex` main (commit 868c9ed), `codex-rs/models-manager/models.json`, same snapshot pattern as `registry-snapshot.json`.
- `scripts/update-codex-models-snapshot.mjs` + `npm run codex-models:snapshot` — refresh script (proxy-aware fetch, shape validation, keeps existing snapshot on failure).
- `src/codex-models.ts` — `codexWindowForModel` (emulates codex's lookup: longest-prefix slug match + namespaced-suffix retry + 272K fallback; user config overrides intentionally not emulated — they can only clamp DOWN to `max_context_window`, never raise, so the table value is a safe upper bound), `isCodexClient` (User-Agent starts `codex_cli_rs/` — the default originator; the `originator` header is only sent for non-default thread originators), `codexAlignedWindow` (the min()), test hooks.
- `src/server.ts` — after `resolveRequestConfig`, clamp `reqConfig.modelContextLimit` at codex's perception for codex clients; the clamped value is authoritative (codex's own config), so `nativeFromFallback` is cleared — otherwise the low-confidence fallback floor (100K, applied after output-headroom reservation) would push the budget back above codex's threshold and re-introduce the misalignment. Info log on clamp.

## Verification

- `npm run typecheck` clean
- `npm test`: 754/754 (master baseline 740 + 14 new)
  - `tests/codex-models.test.ts` (8): snapshot integrity; exact slug (context_window wins over max); longest-prefix; namespaced suffix; not-in-table → 272K (incl. double-slash miss); `isCodexClient` variants; min() acceptance matrix (in-table / not-in-table / below-perception untouched / equal untouched / non-codex untouched / user-override-below untouched); test hooks
  - `tests/codex-window-align.test.ts` (6, e2e through the proxy, observed via `metadata.effectiveContextLimit`): in-table clamp (gpt-5.5 400K→272K); non-codex control (400K); prefix (gpt-5.5-mini→272K); not-in-table fallback (glm-5 1M→272K); below-perception untouched (claude-sonnet-4-5 200K); per-request resolution (UA disappears → reverts to 400K)
- `npm run build` success; snapshot verified inlined in `dist/index.js`

Independent of PR #322 (PR-D) and PR #323 (PR-C); branches from master.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
   "scripts": {
     "build": "tsup",
     "registry:snapshot": "node scripts/update-registry-snapshot.mjs",
+    "codex-models:snapshot": "node scripts/update-codex-models-snapshot.mjs",
     "typecheck": "tsc --noEmit --project tsconfig.build.json",
     "test": "node --import tsx --test tests/*.test.ts",
     "start": "node dist/index.js"
--- a/scripts/update-codex-models-snapshot.mjs
+++ b/scripts/update-codex-models-snapshot.mjs
@@ -1,0 +1,88 @@
+#!/usr/bin/env node
+/** Refresh src/codex-models-snapshot.json from openai/codex (main branch).
+
+The snapshot is the SLIM form of codex's bundled model table
+(codex-rs/models-manager/models.json) — only the fields bili consumes for
+budget alignment (#321 PR-E1): slug, contextWindow, maxContextWindow, and the
+optional autoCompactTokenLimit / effectiveContextWindowPercent. The full
+upstream file is ~424KB of tool/modality metadata bili never reads; the slim
+form keeps the bundle small and the contract explicit.
+
+Run manually or before a release:
+    npm run codex-models:snapshot
+
+Node's global fetch ignores http(s)_proxy env vars, so the script tries the
+configured shell proxy first (undici ProxyAgent) and falls back to a direct
+connection. If BOTH fail the existing snapshot is left untouched (exit 1).
+*/
+import { writeFile } from "node:fs/promises";
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+
+const SOURCE_URL = "https://raw.githubusercontent.com/openai/codex/main/codex-rs/models-manager/models.json";
+const OUT_FILE = path.join(path.dirname(fileURLToPath(import.meta.url)), "..", "src", "codex-models-snapshot.json");
+const proxyUrl = process.env.https_proxy || process.env.HTTPS_PROXY || process.env.http_proxy || process.env.HTTP_PROXY;
+
+async function attempt(dispatcher) {
+    const res = await fetch(SOURCE_URL, {
+        ...(dispatcher ? { dispatcher } : {}),
+        signal: AbortSignal.timeout(20_000),
+        headers: { Accept: "application/json" },
+    });
+    if (!res.ok) throw new Error(`HTTP ${res.status}`);
+    return res.json();
+}
+
+async function fetchFull() {
+    if (proxyUrl) {
+        try {
+            const { ProxyAgent } = await import("undici");
+            const full = await attempt(new ProxyAgent({ uri: proxyUrl }));
+            console.log(`fetched openai/codex via proxy ${proxyUrl}`);
+            return full;
+        } catch (e) {
+            console.log(`proxy attempt failed (${e.message}); trying direct`);
+        }
+    }
+    const full = await attempt(undefined);
+    console.log("fetched openai/codex direct");
+    return full;
+}
+
+let full;
+try {
+    full = await fetchFull();
+} catch (e) {
+    console.error(`could not fetch ${SOURCE_URL}: ${e.message}`);
+    console.error(`keeping the existing ${path.basename(OUT_FILE)} untouched`);
+    process.exit(1);
+}
+
+const models = Array.isArray(full?.models) ? full.models : null;
+if (!models || models.length === 0) {
+    console.error("unexpected upstream shape: expected { models: [...] }");
+    console.error(`keeping the existing ${path.basename(OUT_FILE)} untouched`);
+    process.exit(1);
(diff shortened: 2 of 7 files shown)


ranxianglei:

[bot] **Review: verified, LGTM** — all claims in the PR body hold up under independent re-verification. One minor comment-accuracy nit (non-blocking).

**Pre-flight (branch `2026-08-28_codex-model-alignment` @ c6d1f2c, based exactly on master 744e658):**
- `npm run typecheck` — clean
- `npm test` — **754/754** on branch; ran master in a worktree: **740/740**, so the "740 baseline + 14 new" claim checks out (8 unit + 6 e2e)
- `npm run build` — success; snapshot verified inlined in `dist/index.js` (`codex_cli_rs`, `gpt-daybreak-red-latest` present)

**Cross-checked against openai/codex @ 868c9ed (fetched the actual sources):**
- `models-manager/src/model_info.rs` `model_info_from_slug`: unknown models get `context_window: Some(272_000), max_context_window: Some(272_000)` — the boundary correction vs the issue spec is correct; not-in-table = perceived at 272K, not unlimited
- `protocol/src/openai_models.rs:488-506`: `resolved_context_window()` = `context_window.or(max_context_window)` — matches `resolvedWindow()`; `auto_compact_token_limit()` = `min(config_limit, resolved*9/10)` — an explicit config limit can only LOWER the threshold, so clamping bili to the resolved window is a safe upper bound, and the "don't emulate user overrides" rationale is sound
- `models-manager/src/manager.rs:617-671`: lookup is longest-prefix → single-segment namespaced-suffix retry (charset `[A-Za-z0-9_-]`, double-slash/empty-namespace rejected) → fallback. The TS `lookupWindow`/`longestPrefixMatch` is a line-for-line match of that
- `login/src/auth/default_client.rs:40,164-195,137-145`: `DEFAULT_ORIGINATOR = "codex_cli_rs"`, UA = `{originator}/{version} (…)`; the `originator` header is only sent when the thread originator differs from the process default — the PR's UA-detection claim is accurate
- Snapshot fidelity: all 10 models match upstream `models.json` field-for-field (null `auto_compact_token_limit` correctly omitted by the slim script)
- e2e values match `CONTEXT_LIMIT_TABLE` (gpt-5→400K, glm-5→1M, claude-→200K); MITM path is covered by construction — CONNECT-terminated requests flow through the same `handle()` with the client's original headers, so the clamp is mode-agnostic even though the e2e rig exercises `/bili/` mode
- Refresh script failure path verified live: fetch timeout → existing snapshot kept untouched, exit 1 (success path couldn't be exercised here — live `main`-ref resolution times out from this network even with bare curl; pinned-commit fetches work, so this is network-side, not a script defect)

**Minor (non-blocking):**

1. **The `nativeFromFallback` rationale is inaccurate** (PR body + comment at `src/server.ts:808-810`). It says the 100K fallback floor "would push the budget back above codex's threshold and re-introduce the misalignment" — that can't happen: the floor only raises `reserved` up to `FALLBACK_EFFECTIVE_WINDOW_FLOOR = 100_000` (`src/config.ts:156`) and only when `reserved < 100_000` (`src/server.ts:1043-1046`), while codex's perception is always ≥ 272K, so a floored budget (100K) stays below codex's 90% threshold (≥ 244.8K). Clearing the flag is still *semantically* correct — the clamped value derives from codex's own table (an authoritative source), so it shouldn't be treated as low-confidence — and the only behavioral delta is that when headroom reservation drops `reserved` below 100K, the budget stays at `reserved` instead of being raised to 100K (both below codex's threshold). Suggest rewording the comment/PR body to "the clamped value is authoritative, so it must not be treated as a low-confidence fallback" rather than the floor-mechanism claim.

2. (Design note, acceptable for E1) Codex clients with a custom originator (`CODEX_INTERNAL_ORIGINATOR_OVERRIDE_ENV_VAR` / `set_default_originator`) send a UA that doesn't start with `codex_cli_rs/`, so they escape the clamp and the #292-class misalignment can return for them. Documented in the PR — fine to leave as-is.

3. (Future-proofing, one-liner) `codexWindowForModel` ignores the `autoCompactTokenLimit` field the snapshot schema carries. Harmless today (all null upstream, and the 55%-vs-90% band gap absorbs a lower explicit limit), but if a future upstream table sets explicit limits, that field would be the input to tighten the cap.

No merge from my side — that's human-only. Ready to merge once you're happy (or after the comment reword in #1, if you want it).

PR #328, opened 2026-08-29
fix: warn when running process is stale relative to on-disk install (#327)

ranxianglei:
## What

`checkForUpdate` reads the version from the **on-disk** install (`diskVersion ?? opts.currentVersion`) before comparing with the npm registry. When an auto-update has already replaced the on-disk install but the process was never restarted, every periodic check logs `[update] current=0.1.62 latest=0.1.62 (up to date)` — misleading, because the *running* code is still the old version. The one-time "Restart to finish" message at install time is the only signal, and once missed, it is gone forever.

This PR makes the up-to-date branch detect that state and log a persistent warning instead:

```
[update] running v0.1.55 but v0.1.62 is installed — restart bili to activate (auto-update replaced the on-disk install; this process is still on the old code)
```

## Why (issue #327)

The reported symptom (context meter jumping 0% → 405% then an EMERGENCY compress) was caused by a stale v0.1.5x process: preflight payload-estimate trigger (v0.1.60) and credit netting (v0.1.56) were not running. Disk already had 0.1.62, so the updater kept saying "up to date" and the user had no way to know a restart was needed. Restarting fixes the symptom; this PR fixes the *silence*.

## Changes

- `src/update.ts`: new exported pure fn `staleInstallStatus(diskVersion, runningVersion) → "restart" | "current"` (reuses existing strict `isNewer`); up-to-date branch of `checkForUpdate` warns when disk > running.
- `tests/update-restart-reminder.test.ts`: 5 unit tests (disk-newer / equal / disk-older / undefined / numeric-vs-lexicographic compare).

No control-flow change to the download/extract/install path — log-only in the check path.

## Pre-flight

- `npm run typecheck` — clean
- `npm test` — 745/745 pass (740 existing + 5 new)
- `npm run build` — success

## Release note (§5)

This touches `src/update.ts`, so per the no-op-release rule this is flagged for human decision: the change is log-only in the version-check path (no download/extract/install/version-compare behavior change), but a no-op release first is the conservative route if you want the rule applied strictly.

### DIFF
--- a/src/update.ts
+++ b/src/update.ts
@@ -67,6 +67,15 @@ function isNewer(latest: string, current: string): boolean {
     return false;
 }
 
+/** Stale-install decision for the up-to-date branch of an update check —
+ *  exported for tests. "restart" means the on-disk install is newer than the
+ *  running process: an auto-update landed in-place but the process was never
+ *  restarted, so the one-time "Restart to finish" message is long gone and
+ *  every "up to date" line since has been misleading (#327). */
+export function staleInstallStatus(diskVersion: string | undefined, runningVersion: string): "restart" | "current" {
+    return diskVersion !== undefined && isNewer(diskVersion, runningVersion) ? "restart" : "current";
+}
+
 async function readLastCheck(): Promise<number> {
     try {
         const data = await readFile(THROTTLE_FILE, "utf-8");
@@ -354,7 +363,11 @@ export async function checkForUpdate(opts: UpdateOptions, force = false): Promis
         const currentVersion = diskVersion ?? opts.currentVersion;
 
         if (!isNewer(latest, currentVersion)) {
-            loggerLog("info", `[update] current=${currentVersion} latest=${latest} (up to date)`);
+            if (staleInstallStatus(diskVersion, opts.currentVersion) === "restart") {
+                loggerLog("warn", `[update] running v${opts.currentVersion} but v${diskVersion} is installed — restart bili to activate (auto-update replaced the on-disk install; this process is still on the old code)`);
+            } else {
+                loggerLog("info", `[update] current=${currentVersion} latest=${latest} (up to date)`);
+            }
             return;
         }
 
--- a/tests/update-restart-reminder.test.ts
+++ b/tests/update-restart-reminder.test.ts
@@ -1,0 +1,24 @@
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import { staleInstallStatus } from "../src/update.ts";
+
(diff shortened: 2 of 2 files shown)


ranxianglei:

[bot] **PR #327 reviewed — LGTM.** I checked out `2026-08-29_update-restart-reminder` (452904e) in a worktree and verified everything independently.

**Diff vs. description** — matches exactly:
- `src/update.ts` (+15/−1): new pure `staleInstallStatus(diskVersion, runningVersion) → "restart" | "current"` reusing the existing strict 3-component numeric `isNewer` — no new compare logic. The up-to-date branch compares `diskVersion` against `opts.currentVersion` (the startup version), which is the correct semantics: warn only when disk is strictly newer than running.
- `tests/update-restart-reminder.test.ts`: 5 tests as described.
- Confirmed no control-flow change in the download/extract/install path — log-only in the check branch.
- `currentVersion` comes from `VERSION` in `src/cli.ts:35` (read from package.json, no `v` prefix), so the `v${...}` formatting in the message is correct.

**Pre-flight (fresh worktree, `npm ci`):**
- `npm run typecheck` — clean
- `npm test` — 745/745 pass (the 5 new tests also pass in isolation)
- `npm run build` — success

**Functional smoke test** (stubbed registry fetch, isolated `XDG_CACHE_HOME`/`XDG_STATE_HOME` so the running daemon's throttle/lock files were untouched):
- running 0.1.55 / disk 0.1.62 → `[warn] [update] running v0.1.55 but v0.1.62 is installed — restart bili to activate (auto-update replaced the on-disk install; this process is still on the old code)` — exactly the #327 scenario ✅
- running 0.1.62 / disk 0.1.62 → `[info] current=0.1.62 latest=0.1.62 (up to date)` ✅
- running 0.1.70 / disk 0.1.62 (manual downgrade) → `[info] up to date`, no false warning ✅

**Edge cases considered:**
- If npm `latest` is newer than disk, the check takes the download path as before (no warning) — correct, since the install flow re-logs "Restart to finish".
- The warning repeats on every periodic check (3 min) — persistent by design, which is the point of the PR.

**Nit (non-blocking):** the new warning line uses a literal em dash `—` while other log strings in this file use the `\u2014` escape (e.g. the "throttled" line). Cosmetic only.

**Release decision (§5):** my read is this does not require a no-op release first — the rule targets changes to the upgrade mechanism (download/extract/install/version-compare logic), and this is a log-only line in the version-check path that cannot affect the upgrade path's behavior. Shipping directly is safe; a no-op first is also fine if you want the rule applied strictly. Your call.

PR #325, opened 2026-08-28
feat: intercept + forge codex native compaction when ACP is healthy (#321 PR-E2)

ranxianglei:
**Model: qwen3.8-27b**

Closes the last major piece of #321 (PR-E2): **conditional interception + forgery of codex's native compaction requests**, giving a deterministic handoff from codex's ledger to the bili/ACP state.

## What

When **all** of these hold, bili forges a success response and **never contacts upstream**:
1. `BILI_CODEX_COMPACT=intercept` (kill-switch; **default `pass`** until real-device validation flips it),
2. the client is codex (User-Agent starts `codex_cli_rs/`),
3. the ACP state is **healthy**: transform ok + steady-state `lastInputTokens < 90%` of the effective window + at least one **active block** to hand off.

Two request shapes are handled, both matching codex's source-verified parsers:
- **Trigger form** (input ends with `compaction_trigger`): forged **2-frame SSE** — `response.output_item.done` carrying exactly one `{type:"compaction", id:"fc_bili_*", encrypted_content:"bili:acp:<ACP summaries>"}` + `response.completed` with honest usage. Matches `collect_compaction_output` (exactly one compaction item + `response.completed`; every other event is ignored; missing completed → retryable error).
- **Endpoint form** (`POST /responses/compact`): forged **JSON `{output:[...]}`** where `output` is the rebuilt history (recent items + ACP summary via the normal `responsesToCore → processTurn → stripKernelSummaries → patchResponsesInput` pipeline). Matches `CompactHistoryResponse`.

Otherwise (kill-switch off, ACP not keeping up ≥90%, or nothing compressed yet) the request **passes through** to upstream and native compaction backstops — rebased correctly by the PR-C semantic gate. Bili-generated compaction items (`fc_bili_` prefix / `bili:acp:` sentinel) are stripped from subsequent requests so codex's echoed item never double-counts; the summaries come from ACP state, not re-sent blobs.

## Why this is the "fundamental" fix

With E2, a healthy ACP state is handed to codex deterministically: codex's ledger resets to the (small) forged usage and its history becomes the ACP summary + recent items. The next request matches the ACP state exactly — no heuristic rebase needed on the happy path. The `markNativeCompactionBoundary`/reconcile machinery remains only as the fallback for the pass-through (unhealthy) case, where PR-C's semantic gate does the rebase.

## Files
[...]

PR #317, opened 2026-08-28
feat: partition codex subagent threads by turn-metadata (PR-A of #316)

ranxianglei:
Model: glm-5.3

Part 1 (PR-A) of #316 — codex subagent isolation via turn-metadata partition. Independent of PR-B (tail-window reattach); both are based on master.

## Problem

Codex (>=0.147) **reuses the root `session-id` header on subagent threads** (verified in captures: `codex exec` root vs `codex review` subagent both send the same `session-id`, but each subagent thread carries a unique `thread-id` and an `x-codex-turn-metadata` JSON with `thread_source: "subagent"`). The legacy identity chain therefore mapped every thread of one task onto a single session — subagents inherited the root's compression state. This breaks #150's isolation guarantee: a guardian subagent must read the user's original authorization verbatim, never a compressed summary.

The kernel's `subagentNamespace` is a no-op for codex today: both root and subagent send empty `instructions`, so the kernel's empty-instructions early-return never anchors and the subagent lands on the root session.

## Change

`src/session-id.ts` — new `codexTurnIdentity(headers)`:
- When `x-codex-turn-metadata` is present **and** cross-checked (`metadata.thread_id === thread-id` header), partition by `thread_source`:
  - `"user"` → the `session-id` header (current root semantics, stable across turns)
  - `"subagent"` → the `thread-id` header (fresh independent state per thread, #150 semantics)
- Any other `thread_source`, unparseable JSON, missing/mismatched `thread_id`, or a missing `session-id` on a `user` turn → `undefined`: the caller falls through to the legacy chain unchanged.

`src/server.ts` — wire it into the Responses identity resolution:
- The trusted id enters the **verbatim session chain directly** (no hashing, no new dimensions, #286/#288 philosophy) and **bypasses `subagentNamespace`**, leaving the kernel's empty-instructions non-anchoring path untouched for metadata-less clients.
- No-metadata clients (omp `prompt_cache_key`, etc.) see **zero regression** — `codexTurn` is `undefined` so the legacy chain (including pck promotion) runs as before.

## Tests

- `tests/proxy-session-id.test.ts`: unit coverage for both `thread_source` values, cross-check mismatch degradation, JSON parse failure, unknown `thread_source`, missing headers, and the omp pck chain staying intact.
[...]

PR #329, opened 2026-08-29
Model: glm-5.3 — test: e2e suite — real codex through bili (warmup/load/compress/purity/forge)

ranxianglei:
Model: glm-5.3 (vllm)

## What

`tests/e2e/e2e-codex.test.ts` — a single gated e2e test that runs the **real codex CLI** (0.147) through a bili proxy against a **real Responses-compatible upstream**, asserting the full context-management lifecycle:

1. **warmup** — plants a known contamination value (`1400`)
2. **load growth** — payload user-messages ride the wire verbatim (models routinely shrink tool output — `seq 1 1500 | tail -n 1` was observed), so growth is deterministic
3. **ACP compress** — under a forced window (`BILI_LAUNCHER_MODEL_WINDOWS`), asserts `preflight compressed`/`compress requested` and a usage drop
4. **purity** — post-compress recall (row count + sentinel value) must be correct and must NOT echo the planted contamination (validated end-to-end: `1500, 30217` survived T1 fold + T2 distill)
5. **forge** (gated `E2E_FORGE=1`) — restart under a 128k window (persist reloads the same session; blocks survive), then a codex-side auto-compact limit below the ACP trigger makes codex emit its native compaction request; asserts `codex compact intercepted … upstream not contacted`, a `fc_bili_` compaction item in the codex rollout, and intact recall after the forged compact

Plus `test:e2e` script, `tests/e2e/README.md`, and a `workflow_dispatch` CI job (`ci-e2e.yml`) with upstream URL/key from repo secrets.

## Gating

- Suite **skips by default** (`ACP_TEST_E2E=1` to run) — `npm test` stays free
- `E2E_CHECK=1` preflight: codex version + dist path + upstream probe, no tokens
- Forge phase additionally needs `E2E_FORGE=1` and a dist with `BILI_CODEX_COMPACT` support (#325)

## Real-device validation (local, full stack incl. #317–#328 + #325 with UA fix)

`pass 1 fail 0` in 5.5 min against the local relay; forge evidence: `gate=true steady=45818/128000 activeBlocks=3 → forged SSE with 3 block summary(s), upstream not contacted`.

## Bugs found while building this (report-only; not fixed here)

- **codex UA regression**: codex 0.147 exec sends `user-agent: codex_exec/0.147.0 …`, not `codex_cli_rs/…` — `isCodexClient` never matched, so the #325 intercept was dead code for exec-mode clients. Fix pushed to #325 (c507e0c).
- **preflight fail-fast × soft-protection livelock**: an over-window payload whose foldable ranges are all inside the protected zone 502s forever; the EMERGENCY nudge recovery path is cut because the request never reaches the model. Suggests a follow-up issue.
- relay usage reporting inflates ~2×, which pins steady usage near the window ceiling — the forge gate's 90% margin needs a window with headroom (the suite restarts under 128k for this reason).

## Pre-flight

- typecheck ✓ (no new errors)
- `npm test` 816/816 ✓
- build ✓; `E2E_CHECK` preflight ✓ on this tree

### DIFF
--- a/.github/workflows/ci-e2e.yml
+++ b/.github/workflows/ci-e2e.yml
@@ -1,0 +1,44 @@
+name: E2E (real codex)
+
+on:
+  workflow_dispatch:
+    inputs:
+      forge:
+        description: "run forge phase (needs BILI_CODEX_COMPACT support in master)"
+        type: boolean
+        default: false
+
+jobs:
+  e2e:
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+      - name: Install codex CLI
+        run: npm install -g @openai/codex@latest
+      - run: npm ci
+      - run: npm run build
+      - name: Preflight
+        run: E2E_CHECK=1 node --import tsx --test tests/e2e/e2e-codex.test.ts
+        env:
+          E2E_UPSTREAM_URL: ${{ secrets.E2E_UPSTREAM_URL }}
+          E2E_UPSTREAM_KEY: ${{ secrets.E2E_UPSTREAM_KEY }}
+      - name: E2E suite
+        run: ACP_TEST_E2E=1 node --import tsx --test tests/e2e/e2e-codex.test.ts
+        env:
+          E2E_UPSTREAM_URL: ${{ secrets.E2E_UPSTREAM_URL }}
+          E2E_UPSTREAM_KEY: ${{ secrets.E2E_UPSTREAM_KEY }}
+          E2E_FORGE: ${{ inputs.forge && '1' || '' }}
+      - name: Collect artifacts
+        if: always()
+        run: |
+          mkdir -p artifacts && cp -r tmp/e2e-codex-* artifacts/ 2>/dev/null || true
+      - uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: e2e-logs
+          path: artifacts/
+          retention-days: 7
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -106,6 +106,41 @@ because `package.json` has proper `bin` + `files` fields. Either approach is
 fine; the registry install is preferred for testing the real published
 artifact.
 
+### E2E Regression (real client through bili)
+
+`tests/e2e/e2e-codex.test.ts` drives the **real `codex` CLI** through a built
+proxy against a real Responses-compatible upstream and asserts the full
+context lifecycle end-to-end: warmup → load growth → ACP compress → purity →
+native-compact interception (last phase gated by `E2E_FORGE=1`). Full phase
+details, env vars, and mechanics: `tests/e2e/README.md`.
+
+```bash
+# zero-token preflight (codex binary + dist + upstream reachable)
+E2E_CHECK=1 node --import tsx --test tests/e2e/e2e-codex.test.ts
+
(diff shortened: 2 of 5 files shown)


ranxianglei:

[bot] Picked up. PR #329 is already open on branch `2026-08-29_e2e-suite` (head `eed2c8b`): https://github.com/ranxianglei/billion-context/pull/329

Verifying the tree now: typecheck, `npm test`, build, and `E2E_CHECK=1` preflight. Will also file the follow-up issue for the preflight fail-fast × soft-protection livelock you flagged. Results shortly.

ranxianglei:

[bot] Verification complete on branch `2026-08-29_e2e-suite` (commit `eed2c8b`):

**Pre-flight (this tree = master + this commit)**
- `npm run typecheck` ✓ no errors
- `npm test` ✓ **745/745 pass, 0 fail** (your 816 was on your fuller tree with #325 etc.; this branch only adds the 4 e2e files)
- `npm run build` ✓ (`dist/index.js` 2.52 MB, acp-kernel inlined)
- `E2E_CHECK=1 node --import tsx --test tests/e2e/e2e-codex.test.ts` ✓ **pass 1 fail 0 skip 1** — printed `codex: codex-cli 0.150.1; dist: …/dist/index.js; upstream: http://127.0.0.1:8199/v1`, probe `unreachable` (no local relay in this sandbox, as expected)

**Full-suite note:** the `ACP_TEST_E2E=1` run needs a real Responses upstream + tokens, which this sandbox doesn't have — so your real-device validation (`pass 1 fail 0` in 5.5 min, forge evidence `gate=true steady=45818/128000 activeBlocks=3 → forged SSE with 3 block summary(s), upstream not contacted`) plus the `ci-e2e.yml` `workflow_dispatch` job (secrets-fed upstream) remain the full-verification path. One heads-up: codex here is **0.150.1** (you targeted 0.147) — preflight passes on it; the UA fix from #325 (c507e0c) should be confirmed once #325 lands, since the forge phase depends on `isCodexClient` matching exec-mode UAs.

**Design checks that look right:**
- Gating is sound: default `npm test` glob is `tests/*.test.ts` (no subdirs), so the suite can't leak into normal runs even before the `ACP_TEST_E2E` gate; `test:e2e` script + `E2E_CHECK`/`E2E_FORGE` layers match the README.
- `turnExpect`'s 3-attempt retry with the "直接给出上一请求的答案…" prompt correctly isolates the relay #221 no-content family from real failures.
- Forge phase's `-c model_auto_compact_token_limit=999999` warm turn → settle → `3000` trigger sequence is the right way to force codex's native compaction below the ACP trigger.
- `WORK_ROOT = cwd/tmp` (not `TMPDIR`) sidesteps codex's TMPDIR-home refusal — good call, and the CI artifact step picks it up.

**Follow-up issue filed:** the preflight fail-fast × soft-protection livelock is now tracked at **[#330](http://127.0.0.1:3002/ranxianglei/billion-context/issues/330)** — over-window payload with all foldable ranges inside the protected zone → permanent `preflight fail-fast 502 (retryable=false)`, EMERGENCY nudge path cut because the request never reaches the model. The issue includes the exact code path (`src/server.ts:1802-1807`, `src/preflight.ts:264-266`), impact, and a suggested fix order (relax soft protection under overflow → emergency truncation → actionable error).

The other two report-only bugs need nothing from this PR: the UA regression is already fixed in #325 (c507e0c), and the relay's ~2× usage inflation is a relay property the suite already works around via the 128k forge restart (noted in #330's impact section as a compounding factor).

PR stays open for your review: https://github.com/ranxianglei/billion-context/pull/329

Issue #330, opened 2026-08-29
preflight fail-fast × soft-protection livelock: over-window payload with nothing foldable outside the protected zone 502s forever


ranxianglei:
## Symptom

A session whose payload exceeds the model window and whose **entire foldable content sits inside the soft-protected recent zone** gets stuck in a permanent 502 loop. Every request fails fast before forwarding, and no recovery path exists — the request never reaches the model, so the EMERGENCY nudge (the only other recovery mechanism, #299) can never fire.

Found while building the e2e suite (#329): the suite's load-growth phase rides large user-messages on the wire verbatim, and under a forced small window (`BILI_LAUNCHER_MODEL_WINDOWS`) the recent zone ends up covering all foldable content.

## Mechanism

1. Preflight trigger: `tokenCount >= config.modelContextLimit` (`src/server.ts:1788`).
2. The kernel's nudge computes `compressibleRanges` **excluding** the soft-protected recent zone (`preserveRecentMessages`, default 5, plus the last user message) — acp-kernel: "messages there are excluded from the range but do not fail it (an entirely-protected range fails with a clear error)".
3. When nothing viable remains (>= 200 tokens per `viableRanges`):
   - `src/server.ts:1802-1807` → `failFast(502, "no part of the conversation is compressible (nothing left to fold)", false)`, or
   - `src/preflight.ts:264-266` → `failure = { kind: "exhausted", detail: "no compressible ranges remain in the conversation" }` → `src/server.ts:1853` → `failFast(502, ..., false)`.
4. The payload is never forwarded ("The over-window payload was NOT forwarded." — `src/server.ts:1798`). The model never sees it, so the EMERGENCY nudge / model-driven compress can never fire.
5. Nothing changes between attempts (same session state, same payload, same window) → every retry is an identical 502. `retryable: false` tells the client it is a hard failure, but it is actually a *permanent* one — the session is bricked until the window is raised or history is trimmed externally.

This is the same stuck-session class that #247 fixed for model switches (preflight was that fix); #301's fail-fast introduced a new variant: when preflight itself cannot help, the proxy 502s instead of forwarding (where the upstream 400 would at least surface to the user and the reactive path could still try).

## Impact

- A session can be **permanently stuck with no recovery path** — worse than a transient error.
- Likely in real operation: any workload where one large recent message (big tool output, big user payload) pushes context over a small forced window. The e2e suite hits this deterministically under a forced window.
[...]

PR #337, opened 2026-08-29
fix: preflight relaxes the soft-protected recent zone under overflow (#330)

ranxianglei:
## What

Fixes the #330 livelock: an over-window payload whose only foldable content sits inside the soft-protected recent zone (`preserveRecentMessages` / `preserveRecentTokens` / most-recent user message) used to 502 forever with no recovery path.

Two things bricked it:
1. The server pre-check failed fast on the **normal-config** `compressibleRanges` (which exclude the soft zone) *before* `preflightCompress` ever ran.
2. Even when preflight ran, it had no recovery once nothing was foldable outside the zone — it just set an exhausted failure.

## Changes

- **`src/preflight.ts`**
  - When no range is foldable outside the soft zone **and** the payload's own estimate still overflows, relax the soft zone to zero (oldest-first within it) and retry. The hard `protectedTools` exclusion is computed independently and still applies. Only when truly nothing is foldable does it fail fast — now with an **actionable** error (raise the model context window / restart the session).
  - The relaxation is gated on the payload's own estimate (`result.payloadEstimate`), **not** `currentTokens` — the latter is floored by a possibly-stale `lastInputTokens` from a prior model, which would fold protected content even when the real payload already fits. (This regression was caught by the existing #247 model-switch test.)
  - The per-round pipeline now runs with the kernel's **emergency-truncate node suppressed** (window inflated so `usage < truncate.threshold`; compressible ranges derive from the protected zone, not usage). Without this, the node trims the very tool output overflowing the window, making the fit check undershoot and break *before* the soft zone is relaxed.
- **`src/server.ts`**
  - Decide forward/fail on the payload **actually forwarded** (re-measured after the rebuild), not on `result.payloadEstimate` — the relaxed-zone `processTurn` trims that estimate more than the normal-config prepare does, which could turn a guaranteed-400 forward into a false "fits".

## Tests

- New e2e `#330`: over-window payload whose only foldable content is in the protected recent zone → soft zone relaxed, folded, forwarded (200).
- The "truly incompressible" e2e is now a **hard-protected** tool result (`protectedTools: ["bash"]`): 502 + actionable remedy, no summarization call, no forward.
- Full suite **746/746**, `npm run typecheck` clean, `npm run build` clean.

## Follow-up (not in this PR)

The issue also suggested the kernel's `truncateLargeToolOutputs` as a second lever. It's stateless/non-persistent and only touches tool-results (the repro is a large user message), so it's largely redundant with the soft-zone relaxation here — left as a possible follow-up.

### DIFF
--- a/src/preflight.ts
+++ b/src/preflight.ts
@@ -224,6 +224,27 @@ async function summarizeRange(deps: PreflightDeps, content: string, startRef: st
 
 const ABORTED_FAILURE: PreflightFailure = { kind: "aborted", detail: "the client disconnected during preflight compression" };
 
+// #330: the soft-protected recent zone (preserveRecentMessages /
+// preserveRecentTokens / most-recent user message) can cover ALL foldable
+// content when one large recent message pushes the payload over a small
+// window — preflight then 502s forever with no recovery path. Under overflow
+// the soft zone is a preference, not a constraint: relax it to zero so its
+// oldest content becomes foldable. The hard protectedTools exclusion is
+// computed independently of preserveRecent* and still applies.
+function relaxedConfig(config: Config): Config {
+    return { ...config, preserveRecentMessages: 0, preserveRecentTokens: 0 };
+}
+
+// #330: the preflight's fit check must reflect the durable payload, not the
+// per-turn emergency-truncate side effect — that node would trim the very tool
+// output overflowing the window, making currentTokens undershoot and the loop
+// break before the soft zone is relaxed. Inflate the window so usage stays
+// below truncate.threshold; compressible ranges derive from the protected zone,
+// not usage, so this is safe.
+function noEmergencyTruncate(config: Config): Config {
+    return { ...config, modelContextLimit: config.modelContextLimit * 100 };
+}
+
 export async function preflightCompress(deps: PreflightDeps, messages: CoreMessage[]): Promise<PreflightResult> {
     const limit = deps.config.modelContextLimit;
     const result: PreflightResult = { compressedRanges: 0, savedTokens: 0, payloadEstimate: estimateCoreMessages(messages) + (deps.imageFloor ?? 0) };
@@ -239,6 +260,12 @@ export async function preflightCompress(deps: PreflightDeps, messages: CoreMessa
     let currentTokens = deps.session.stats.lastInputTokens;
     let startTokens = -1;
     let failure: PreflightFailure | undefined;
+    let activeConfig = deps.config;
+    let relaxed = false;
+    const relaxedExhaustedDetail =
+        `the payload still exceeds the window after folding everything compressible, including the soft-protected recent zone ` +
+        `(last ${deps.config.preserveRecentMessages} messages + most recent user message), which was relaxed under overflow; hard protectedTools remain excluded. ` +
+        `Raise the model context window or restart the session to recover.`;
     for (let round = 0; round < MAX_PREFLIGHT_ROUNDS; round++) {
         if (deps.signal?.aborted) {
             failure = ABORTED_FAILURE;
@@ -249,7 +276,7 @@ export async function preflightCompress(deps: PreflightDeps, messages: CoreMessa
         const turn = deps.core.processTurn({
             messages,
             state: deps.session.state,
-            config: deps.config,
+            config: noEmergencyTruncate(activeConfig),
             tokenCount: currentTokens,
             renderTags: "text-only",
         });
@@ -265,7 +292,19 @@ export async function preflightCompress(deps: PreflightDeps, messages: CoreMessa
         if (currentTokens < limit) break;
         const ranges = viableRanges(turn.nudge?.compressibleRanges ?? []);
         if (ranges.length === 0) {
-            failure = { kind: "exhausted", detail: "no compressible ranges remain in the conversation" };
+            // #330: nothing foldable outside the soft-protected recent zone.
+            // Relax the soft zone (oldest-first within it) and retry — the hard
+            // protectedTools exclusion still applies. Gate on the payload's own
+            // estimate (not currentTokens, which is floored by a possibly-stale
+            // lastInputTokens from a prior model): if the real payload already
+            // fits, stop instead of folding protected content.
+            if (!relaxed && result.payloadEstimate >= limit) {
+                activeConfig = relaxedConfig(deps.config);
+                relaxed = true;
+                deps.log("warn", "[preflight] no compressible ranges outside the protected recent zone; relaxing soft protection (preserveRecentMessages/Tokens -> 0) and retrying");
+                continue;
+            }
+            failure = { kind: "exhausted", detail: relaxed ? relaxedExhaustedDetail : "no compressible ranges remain in the conversation" };
             break;
         }
         const range = [...ranges].sort((a, b) => refNum(a.startRef) - refNum(b.startRef))[0];
@@ -315,7 +354,7 @@ export async function preflightCompress(deps: PreflightDeps, messages: CoreMessa
             if (!summary) continue;
             const ctx: RewriteCtx = {
                 core: deps.core,
-                config: deps.config,
+                config: activeConfig,
                 messages,
                 session: deps.session,
                 log: (msg) => deps.log("info", msg),
@@ -342,7 +381,12 @@ export async function preflightCompress(deps: PreflightDeps, messages: CoreMessa
         }
     }
     if (!failure && currentTokens >= limit) {
-        failure = { kind: "exhausted", detail: `the compress budget was exhausted after ${MAX_PREFLIGHT_ROUNDS} rounds` };
+        failure = {
+            kind: "exhausted",
+            detail: relaxed
+                ? relaxedExhaustedDetail
+                : `the compress budget was exhausted after ${MAX_PREFLIGHT_ROUNDS} rounds`,
+        };
     }
     if (result.compressedRanges > 0) deps.session.stats.lastInputTokens = currentTokens;
     result.savedTokens = Math.max(0, startTokens - currentTokens);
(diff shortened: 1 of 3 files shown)


ranxianglei:

[bot] Reviewed `2026-08-29_preflight-soft-zone-relax` (74fe46d) — checked out the branch and verified every claim locally. **Verdict: approve.** The fix is correct, well-reasoned, and genuinely regression-tested.

**Verified clean:**
- `npm run typecheck` — no errors
- `npm run build` — success
- `npm test` — **746/746 pass**

**The tests actually catch the bug (not vacuous):** I ran the PR's test file against **master's** `src/preflight.ts` + `src/server.ts` (keeping the PR's tests). Both new/modified e2e cases **fail on master** and pass on the branch:
- `#330` → 502 on master (old server pre-check bails on normal-config `compressibleRanges` before `preflightCompress` runs), 200 on the branch.
- Modified `#301` (hard-protected `bash` tool result) → also fails on master.

**Design claims validated against acp-kernel 0.0.46 source:**
- `noEmergencyTruncate` (inflate `modelContextLimit` ×100) is safe: `buildCompressibleRanges` derives ranges **structurally** (excludes `protectedZoneRefs` + `protectedTools`), not from usage — so inflating the window suppresses only the usage-gated `emergency-truncate` node, never the range set. Exactly as the comment claims.
- `relaxedConfig` is **complete**, no gap: in `computeProtectedRefs`, the most-recent-user-message protection is gated on `preserveN > 0`, so zeroing `preserveRecentMessages`/`preserveRecentTokens` removes the *entire* soft zone (including the last user message).
- The `noEmergencyTruncate` change is **load-bearing and tested**: with it disabled, the `#301` case's emergency-truncate node trims the overflowing tool output → preflight loop breaks early via `currentTokens < limit` → `result.failure` stays `undefined` → the actionable "Raise the model context window" remedy is lost (falls back to the generic message). Status is still 502 either way, so this is a message-quality fix that the modified `#301` test now pins down.
- `server.ts` correctly re-measures the **rebuilt** (normal-config) payload instead of trusting `result.payloadEstimate` — the relaxed-zone `processTurn` trims that estimate more, which would otherwise turn a guaranteed-400 forward into a false "fits". Right call.

**Code quality (per AGENTS.md):** no `as any`, no `@ts-ignore`, no `console.error` in server modules (uses `log`/`deps.log`), no `<acp>` XML introduced, config objects are spread (no mutation of the caller's config). Comments are explanatory and match the file's existing `#NNN` style.

**Two non-blocking notes:**
1. The "e2e" tests live in `tests/preflight-fail-fast.test.ts` (mock-upstream e2e via `startServer`), not the `tests/e2e/` real-client suite from #329. The "e2e" label in the PR body is accurate for what they exercise (real HTTP through the proxy) — just flagging the location differs from the `tests/e2e/` dir in case that was the intent.
2. The ×100 inflation factor in `noEmergencyTruncate` is a magic number; it's documented by the adjacent comment and is safe (only pushes usage below the truncate/nudge thresholds), so no action needed — just noting it.

The `truncateLargeToolOutputs` follow-up you scoped out looks right to leave for later (stateless, tool-results-only, redundant with the soft-zone relax for this repro). No changes requested.

ranxianglei:
Re-reviewed on current master (2cbbeab, 252 commits ahead of the branch base) and pushed a verified merge head `f78a63d`.

**The merge applied cleanly**, but the post-flight restructure regressed **#496**: deciding forward/fail on `estimateCoreMessages(...)` alone drops the `imageFloor` (images ride the payload verbatim, #488), so an image-dominated payload with overflow evidence looked "fitting" on its text estimate and was forwarded again — the exact 400 loop #496 forbids. Fixed in `f78a63d`: both post-flight fit gates now measure `estimateCoreMessages(...) + imageTokens < limit`.

**Verified on `f78a63d`:** typecheck ✓ · full suite 1047/1049 (the 2 fails are the known `plugin-agent.test.ts` env flakes, pre-existing) · build ✓ · `ACP_TEST_E2E=1` codex 4-phase PASS (242 s). The #330 e2e (soft zone relaxed → folded → forwarded) passes with kernel 0.0.52, so the `noEmergencyTruncate` window-inflation trick is still compatible with the pressure-band gate.

Ready to merge. PR#334 is the weaker alternative (forward-as-is) for the same livelock and is being closed as superseded — this PR implements the issue's preferred fix (relax soft protection, keep hard `protectedTools`).

ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] Independently verified the merge head **f78a63d** (checked out the branch, `npm ci` → acp-kernel 0.0.52). Green light.

**Pre-flight (my run):**
- `npm run typecheck` — clean
- `npm run build` — success
- Targeted `#496` / `#488` / `#301` / `#330` e2e — **12/12 pass**
- Full suite — **1048/1049** on my runner; the single failure is **not from this PR** (below)

**The #496 fix is real and load-bearing.** I confirmed it empirically: stripping just the `+ imageTokens` term from the two post-flight fit gates makes `tests/issue496-forward-once-learn.test.ts` **fail** (AssertionError — request 2 gets forwarded again, reintroducing the 400 loop); restoring it makes it pass. Traced the logic too — all four forward/fail sites now measure text + image cost consistently:
- early-forward gate uses `payloadEstimate < limit`, where `payloadEstimate = estimateCoreMessages(...) + imageTokens` (src/server.ts:2528)
- rebuild gate: `estimateCoreMessages(rebuilt.processedMessages) + imageTokens < limit` (src/server.ts:2618)
- no-progress gate: `estimateCoreMessages(prepared.processedMessages) + imageTokens < limit` (src/server.ts:2619)

The rebuild path reuses the original `imageTokens` constant, which is correct — images ride the payload verbatim (#488) and are never folded, so their cost is invariant across the rebuild. So yes: the rebase dropped these terms when it rewrote the gates around `estimateCoreMessages(...)`, and f78a63d puts them back. Matches your description exactly.

**On the count (my 1048/1049 vs your 1047/1049):** the one failure in my run is `tests/launcher.test.ts → resolveClientCommand: codex/claude resolve to themselves`. It asserts the resolver returns the *bare* command name when the binary isn't on PATH; in this sandbox a `codex`/`claude` executable *is* resolvable on the runner's PATH, so the resolver correctly returns the resolved path and the `deepEqual` trips. I confirmed it's **pre-existing and environmental**, not a rebase artifact — it fails identically on master (`2cbbeab`) and this PR never touches `launcher.ts`. Your 2-fail (plugin-agent flakes) is the same category surfacing in your cleaner env. Net: **zero failures attributable to this change.**

**Not re-run:** the `ACP_TEST_E2E=1` codex 4-phase needs a live local LLM upstream I don't have in this sandbox, so I'm taking your 242 s PASS on that. Everything I could execute independently confirms the head.

**#334 supersession — agreed.** Forward-as-is (let the upstream arbitrate) is strictly weaker than relaxing the soft zone and actually folding the content back under the window; the two are semantically exclusive, so keeping both alongside #337 would conflict. It's already closed (your floor 6) with the right rationale.

Ready to merge on my end — per repo rules the merge itself stays yours.

ranxianglei:
已修复 PR 显示的巨大 diff 问题。

**为什么之前显示 251 commits / 143 files:** 该分支此前用 merge 方式同步 master（head `f78a63d` 是个 merge commit），GitHub 把 diff 锚在 08-29 的老 base 上，于是把 master 一周的演进全部计入了 Files changed。

**修复方式:** 从当前 master（`a184ff9`, v0.1.87）rebase，`cherry-pick -m 1 f78a63d` 单提交落地。现在显示恢复为真实改动：**3 files, +163/-29**。

## 改动内容

| 文件 | 内容 |
|---|---|
| `src/preflight.ts` (+52) | 溢出时软区放宽：`relaxedConfig()` 将 preserveRecentMessages/Tokens 归零重试（硬 protectedTools 仍独立生效）；`noEmergencyTruncate()` 让 fit 检查反映持久 payload 而非 per-turn 截断副作用；失败信息改为可操作文案（提窗口/重开会话） |
| `src/server.ts` (+45/-) | 删除溢出时提前 502 "nothing left to fold"（#330 livelock 元凶）；压缩后 fit 判定改用重建 payload 估计 + imageTokens（修 #496 图主导 payload 的假 fit 回归）；stale-baseline 转发仅在 payload 确实装得下时 |
| `tests/preflight-fail-fast.test.ts` (+95) | e2e #330 场景：唯一可折叠内容都在软保护区内 → 放宽、折叠、转发 + upstream 捕获断言 |

## Pre-flight

- typecheck ✓
- tests/preflight-fail-fast.test.ts 4/4 ✓
- 全量 **1063/1065**（2 个已知 plugin-agent 环境失败，与 master 基线一致）✓
- mergeable ✓（GitHub 显示 clean，无冲突）

Closes #330

ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] Verified the rebased head **bfe9e4c**. The display fix is correct and the change is green on the new base. Green light.

**Display fix confirmed:** bfe9e4c is a **single commit with a single parent = `a184ff9`** (current master tip), so GitHub now anchors the diff correctly → clean **3 files / +163/−29** instead of the old 251-commit / 143-file view. That parent structure is what fixed it.

**Content unchanged from what I validated last round:** `git diff f78a63d bfe9e4c -- src/preflight.ts src/server.ts tests/preflight-fail-fast.test.ts` is **empty** — the rebase moved the base but didn't touch the PR's code, so all prior logic/code-quality review carries over verbatim.

**Re-validated on the newer base** (`a184ff9`, v0.1.87; note acp-kernel bumped **0.0.52 → 0.0.54**, so I re-ran rather than assumed):
- `npm run typecheck` ✓ · `npm run build` ✓
- Targeted #330 / #301 / #496 / #488 e2e — **12/12 pass** (confirms both the `noEmergencyTruncate` window-inflation trick and the #496 `+ imageTokens` fit-gate fix still hold against kernel 0.0.54)
- Full suite — **1064/1065**; the one failure is again the known environmental `resolveClientCommand` flake (a client CLI is resolvable in this sandbox's PATH; fails identically on master; PR doesn't touch launcher.ts). Your 1063/1065 (2 plugin-agent flakes) is the same denominator with your env's flakes instead. **Zero failures attributable to this change.**

`Closes #330` is accurate. Ready to merge on my end — the merge itself stays yours per repo rules.

PR #340, opened 2026-08-29
fix: omp pck stamp for chat-completions with max_tokens; anthropic wire consumes + strips pck (#268)

ranxianglei:
## Problem (#268, still broken in 0.1.62)

`/acp` keeps reporting "No model request in this conversation yet" for `bili omp` sessions on the openai chat-completions wire. User-reported root cause, verified against source and the omp 18.0.6 dist:

`stampPromptCacheKey()` (src/agent/pi.ts) excluded payloads carrying `max_tokens` — intended to skip the anthropic wire — but omp's openai-compat providers send `max_tokens` in **every** chat-completions body (provider config `maxTokensField:"max_tokens"`, confirmed in the omp 18.0.6 dist). So the pck was never stamped, the proxy's openai pck identity path (added in #267) never fired, the conversation map stayed empty, and `/acp` 404'd forever. The #267 test matrix used a synthetic chat payload *without* `max_tokens`, which is why this slipped through.

## Fix

**Plugin (src/agent/pi.ts)** — the user's one-line guard fix: drop `max_tokens` from the stamp exclusion (chat shape = `messages` array + no `input` + no native pck).

**Proxy (src/server.ts)** — make the anthropic wire consume the field deliberately, because the plugin cannot tell the wires apart by shape (both carry `max_tokens`):

- anthropic path mirrors the openai pck promotion: `preferPromptCacheKeyIdentity` over the content fingerprint (session id becomes the omp sid verbatim, `clientProvided=true`), plus the conversation mapping is now recorded for `responses | openai | anthropic`.
- `prepareAnthropic` **strips `prompt_cache_key`** from the outgoing body — it is a proxy-internal identity marker, not part of the Anthropic Messages API, so the real upstream never sees a field it doesn't know (same pattern as the existing `prompt_cache_retention` strip in `prepareOpenai`).

### Why not ship the one-line plugin patch alone

The patch alone would also stamp the anthropic wire (indistinguishable by shape — `BeforeProviderRequestEvent` carries only `{type, payload}`, no URL to scope on). The pck would then be forwarded to api.anthropic.com as an unknown top-level field: zero benefit (the proxy ignored body-pck on that wire) and an unverified upstream-rejection risk. With the proxy half, the anthropic wire is *also* fixed (`/acp` works for omp + Claude models) with no unknown field reaching the real Anthropic.

### Other report claims, checked

- ✅ omp has no `before_provider_headers` (header path is dead code on omp) — confirmed in the 18.0.6 type defs.
- ✅ register-consumption timing deadlock (`!anonAffinity` / `requests === 0` gates) — confirmed; the pck path bypasses it since the first request now carries identity.
- ❌ "omp natively sends `x-session-id` on every model request" — not supported: in the 18.0.6 dist, `x-session-id` appears only in Codex-Live WebSocket signaling and the auth-broker header allowlist, never in the normal model-request path. The reporter's own observed `anonymousPrefixAffinity: true` session metadata proves it (that flag is only set when the request carries no client identity). Consequence: the header-identity route is dead on omp too — pck is the only route, which is what this PR fixes.

## Tests

- `tests/plugin-agent.test.ts` — injection matrix: new real-world case (chat-completions **with** `max_tokens` → stamped), anthropic-shape case flipped to stamped (documents the proxy strip contract).
- `tests/anthropic-pck-identity.test.ts` (new) — mapping recorded by pck on the anthropic wire, field stripped upstream, identity register + pck binds pluginMode (wire injection suppressed), anonymous no-pck requests unchanged.

Checks: `npm run typecheck`, `npm test` (827 pass), `npm run build` — all green.

### DIFF
--- a/src/agent/pi.ts
+++ b/src/agent/pi.ts
@@ -65,9 +65,14 @@ function sessionIdOf(ctx: Ctx): string | undefined {
 // by the omp session uuid). The before_provider_request return value REPLACES
 // the whole outgoing payload (omp onPayload chain, verified in the omp 17.3.8
 // dist), so stamp prompt_cache_key with the omp session id: the proxy binds
-// pluginMode by that identity and /acp finds the session by it. Chat shape only
-// (messages array, no responses `input`, no anthropic `max_tokens`, no native
-// prompt_cache_key); pi is untouched (it stamps x-bili-plugin-conversation in
+// pluginMode by that identity and /acp finds the session by it.
+// Chat shape = messages array, no responses `input`, no native prompt_cache_key.
+// max_tokens is NOT a discriminator: omp's openai-compat providers send it in
+// every chat-completions body (maxTokensField:"max_tokens") exactly like the
+// anthropic wire — excluding it meant the target shape was never stamped
+// (#268). The anthropic wire gets stamped too: the proxy records the mapping
+// from the body pck there as well and strips the field before forwarding to
+// the real Anthropic. pi is untouched (it stamps x-bili-plugin-conversation in
 // before_provider_headers, which outranks the body field).
 function stampPromptCacheKey(event: unknown, ctx: Ctx, agent: string): Record<string, unknown> | undefined {
     if (agent !== "omp") return undefined;
@@ -75,7 +80,7 @@ function stampPromptCacheKey(event: unknown, ctx: Ctx, agent: string): Record<st
     if (payload === null || typeof payload !== "object" || Array.isArray(payload)) return undefined;
     const p = payload as Record<string, unknown>;
     if (!Array.isArray(p.messages)) return undefined;
-    if (p.input !== undefined || p.max_tokens !== undefined) return undefined;
+    if (p.input !== undefined) return undefined;
     if (typeof p.prompt_cache_key === "string" && p.prompt_cache_key.trim().length > 0) return undefined;
     const sid = sessionIdOf(ctx);
     if (sid === undefined || sid.length === 0) return undefined;
--- a/src/server.ts
+++ b/src/server.ts
@@ -896,8 +896,24 @@ async function handle(
                   parsed as { prompt_cache_key?: unknown },
               )
             : undefined;
-        const conversation = protocol === "anthropic"
+        // Anthropic mirrors the openai pck promotion (#268): the omp plugin
+        // stamps prompt_cache_key on every chat-shaped payload — it cannot
+        // tell the anthropic wire apart by shape (both carry max_tokens). The
+        // proxy consumes the field on this wire too (identity + mapping);
+        // prepareAnthropic strips it before the real Anthropic sees it.
+        const anthropicSignal = protocol === "anthropic"
             ? conversationSignalAnthropic(parsed as AnthropicRequestBody, convHeader)
+            : "";
+        const anthropicIdentity = protocol === "anthropic"
+            ? preferPromptCacheKeyIdentity(
+                  convHeader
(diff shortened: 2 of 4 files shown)


ranxianglei:

[bot] Picked up — reviewing the PR now: verifying the root-cause claims against source, checking the plugin/proxy split, and running typecheck/test/build. Will report findings shortly.

ranxianglei:

[bot] Review of PR #340 complete — **LGTM**, with two small nits I've already pushed to the branch (commit `2925e90`).

## Verification (ran on the PR branch)

- `npm run typecheck` — clean
- `npm test` — **827/827 pass** (matches your claim)
- `npm run build` — success
- e2e: preflight can't run in this sandbox (no codex binary, no local upstream at 127.0.0.1:8199). Not a blocker: the diff touches only the anthropic branch of the identity chain, the mapping condition, and `prepareAnthropic`/`prepareCountTokens` — the responses/codex pipeline paths that e2e exercises are byte-identical.

## Code review

**Plugin (`src/agent/pi.ts:83`)** — guard is now `messages array + no input + no native pck`; `max_tokens` no longer discriminates. Stamp is a copy (`{ ...p, prompt_cache_key: sid }`), original payload never mutated; pi still early-returns (`agent !== "omp"`). The updated comment correctly documents the proxy-strip contract for the anthropic wire.

**Proxy (`src/server.ts`)** — verified the anthropic branch mirrors the openai one exactly:
- `preferPromptCacheKeyIdentity` (src/session-id.ts:83) only promotes pck when the kernel would fall to `content-fingerprint`; a real conversation header (`source: "header"`) stays stronger — no override risk.
- `clientProvided` behavior matrix checked: header±pck → unchanged (true); no header + pck → true (the fix — session id = omp sid verbatim, bypasses the anonymous-affinity path and the `!anonAffinity`/`requests === 0` binding gates); no header + no pck → false → anonymous prefix affinity, unchanged (covered by test 4).
- Mapping recording now includes anthropic (src/server.ts:1043) — this is the actual `/acp` fix; it runs unconditionally on pck presence, so it works even before the MCP shell's register lands.
- `prepareAnthropic` strip sits at the single return point (src/server.ts:1313) and covers the kernel-failure fallback too (`rebuilt` is spread from `parsed` in both paths).
- Side effect noted and fine: with pck, `affinityToken` now forwards the sid upstream as an `x-session-id` header (src/server.ts:1922) — identical to the existing openai-wire behavior; unknown headers are ignored by Anthropic.

**Tests** — the new `tests/anthropic-pck-identity.test.ts` covers all four contracts (mapping lookup, upstream strip, pluginMode binding with wire-injection suppression, anonymous no-pck unchanged), and the plugin-agent matrix now has the real-world `max_tokens` case plus the flipped anthropic-shape case documenting the strip contract.

## Nits (pushed as `2925e90`, re-verified green: typecheck + 827 tests + build)

1. **`prepareCountTokens` leaked the field too** — both its success and catch paths spread `parsed` into the forwarded body without the strip, so a stamped `/messages/count_tokens` request (explicit-protocol routes only) would have reached the real Anthropic with the unknown field. Same one-line `delete` added on both paths.
2. **Stale hint** — `NO_IDENTITY_MESSAGE` (src/server.ts:153) said "prompt_cache_key (responses/openai)"; now "(responses/openai/anthropic)".

(`--passthrough` mode still forwards bodies verbatim, pck included — that's the existing debug-mode semantics, left as-is.)

Merging stays with you: https://github.com/ranxianglei/billion-context/pull/340

PR #343, opened 2026-08-29
fix: show real CA path in web UI + Windows ~ expansion doc note (#342)

ranxianglei:
## What

#342: ZCode on Windows cannot find the CA cert because the web UI / docs tell the user to enter `~/.local/share/billion-context/ca/root-ca.pem`, but ZCode on Windows does not expand `~` (and the real file lives at `%USERPROFILE%\.local\share\billion-context\ca\root-ca.pem`). The user has to hand-compute the absolute path per machine — with N working directories this is exactly the pain reported.

## Changes

- `src/web/page.ts`: the ZCode card now renders the machine's **actual** CA path (`rootCaPath()` from `src/ca.ts`) instead of the hardcoded `~/...` string, with a one-click **复制 CA 路径** copy button (reuses the existing `.copy-btn`/`data-copy` mechanism). If the CA file has not been generated yet (MITM never started), the card says so. Added a server-side `escapeHtml` for the injected path (Windows usernames can contain `&`). Hint added: Windows ZCode does not expand `~` — copy-paste the shown path, don't hand-type `~/...`.
- `CONFIGURATION.md` / `CONFIGURATION.zh-CN.md`: MITM one-time-setup step 2 now lists both the POSIX and Windows CA locations and points to the web UI copy button; added an explicit Windows note that ZCode does not expand `~` and the full absolute path must be entered.
- `tests/web-routing.test.ts`: the routing-page test now asserts the UI contains the real `rootCaPath()` (escaped), the copy button, and no longer the hardcoded `~/.local/share/billion-context/ca/root-ca.pem`.

## Pre-flight

- `npm run typecheck` — clean
- `npm test` — 823/823 pass
- `npm run build` — success

No changes to `src/update.ts`, request pipeline, or version.


ranxianglei:

[bot] Reviewed PR #342 (branch `2026-08-29_ca-path-windows`, single commit `78a2f53`). **LGTM — verified, ready for a human to merge.**

## Independent verification

Ran the full pre-flight on the branch myself:
- `npm run typecheck` — clean
- `npm test` — **823/823 pass**
- `npm run build` — success

Plus a live smoke test of the rendered page (built `dist`, two server instances with isolated XDG dirs):

1. **CA exists (normal case):** the ZCode card renders the machine's real absolute path (correctly reflecting an `XDG_DATA_HOME` override), with the `复制 CA 路径` copy button (`data-copy="<real path>"`). No "尚未生成" suffix.
2. **CA absent (`BILI_MITM=0`):** card shows the path + `（尚未生成，启动代理后自动创建）`.
3. **No hardcoded path left:** zero occurrences of `~/.local/share/billion-context/ca/root-ca.pem` in the rendered UI.

## Code review notes

- `rootCaPath()` (`src/ca.ts:38`) is pure (`path.join(caDir(), "root-ca.pem")`) — safe to call per-request in `renderPage`. On Windows `homedir()` → `%USERPROFILE%`, so the rendered path matches the docs' `%USERPROFILE%\.local\share\billion-context\ca\root-ca.pem`.
- Copy button correctly reuses the existing mechanism: `src/web/client.ts:19` reads `btn.dataset.copy`, and the browser decodes HTML entities before `clipboard.writeText`, so the `escapeHtml`'d `&amp;` copies back as a literal `&`. Escaping order (`&` first) is correct.
- `caReady` via `existsSync` is accurate: the CA is created by `ensureRootCA()` at `src/mitm.ts:76`, invoked from `setupMitm` at startup when `mitm.enabled` — which defaults to **true** (`src/config.ts:355`), so the "not yet generated" state only appears when MITM is explicitly disabled.
- Test additions in `tests/web-routing.test.ts` assert the real escaped path, the copy button, and the absence of the hardcoded `~/...` string — good coverage of the regression.
- Scope check: only 4 files touched (`src/web/page.ts`, 2 docs, 1 test). No changes to `src/update.ts`, the request pipeline, or `version`. Docs (EN + ZH) consistently list both POSIX and Windows locations and the no-`~`-expansion note.

## Minor nits (non-blocking, no changes required)

1. The test re-implements the `escapeHtml` mapping inline instead of importing it from `src/web/page.ts` — if the escaping ever changes, the test silently diverges. Could export `escapeHtml` for reuse.
2. Edge case: with MITM explicitly disabled, the hint `启动代理后自动创建` is slightly inaccurate (a restart alone won't create the CA — MITM must be enabled). Very minor since the ZCode card is moot in that mode.

Per repo rules I won't merge — please merge #342 yourself: https://github.com/ranxianglei/billion-context/pull/342

Issue #344, opened 2026-08-29
zcode 用的是gpt-5.6-sol 上下文设置了272000，识别错了上下文空间了


ranxianglei:
相关日志
2026-08-29T11:53:38.257Z [info] mitm <private-host>:443 tunnel established (TLS terminated locally)
2026-08-29T11:53:38.297Z [info] [eb37a6f8-e344-4562-85e0-061fe856f968] processTurn: 372 msgs, renderTags=text-only, 96 text tagged, 0 tool tagged (should be 0 with text-only)
2026-08-29T11:53:38.297Z [info] [eb37a6f8-e344-4562-85e0-061fe856f968] nudge idle: usage=23% (208054/922000), growth=21792/22500 (ref=186262, interval=50000), pendingT1=135045/50000 model=gpt-5.6-sol, reason="growth 21792 < floor 22500, ready: T1 135045, blocked: T1 (cadence)"
2026-08-29T11:53:38.302Z [info] forward POST → https://<private-host>/v1/chat/completions
2026-08-29T11:53:46.248Z [info] [update] checking npm registry for billion-context (last check 50s ago)…
2026-08-29T11:53:46.913Z [info] [update] current=0.1.63 latest=0.1.63 (up to date)
2026-08-29T11:53:49.172Z [info] mitm <private-host>:443 tunnel established (TLS terminated locally)
2026-08-29T11:53:49.209Z [info] [eb37a6f8-e344-4562-85e0-061fe856f968] processTurn: 380 msgs, renderTags=text-only, 96 text tagged, 0 tool tagged (should be 0 with text-only)
2026-08-29T11:53:49.209Z [info] [eb37a6f8-e344-4562-85e0-061fe856f968] nudge INJECT T1: usage=23% (212143/922000), growth=25881/22500 (ref=186262, interval=50000), pendingT1=137864/50000 model=gpt-5.6-sol, reason="T1 effective 137864 >= 50000, growth 25881, usage 23%"
2026-08-29T11:53:49.213Z [info] forward POST → https://<private-host>/v1/chat/completions
2026-08-29T11:54:53.527Z [info] mitm <private-host>:443 tunnel established (TLS terminated locally)
2026-08-29T11:54:53.552Z [info] [eb37a6f8-e344-4562-85e0-061fe856f968] processTurn: 388 msgs, renderTags=text-only, 96 text tagged, 0 tool tagged (should be 0 with text-only)
[...]

PR #347, opened 2026-08-29
fix: 404 unknown /__bili/ paths + actionable MITM CA-rejection warning (#346)

ranxianglei:
## Context (from #346 log analysis)

Analyzing the reporter's actual `bili.log` (ZCode 3.10.1, Windows, MITM via private relay) showed the session-loss symptom is **not** the #344 window bug (no overflow 400s; long sessions up to 290 msgs processed fine). The dominant signal was **160× `TLS error: ... ssl/tls alert certificate unknown`** — the client (ZCode) rejecting bili's MITM root CA on its non-model HTTP clients (session/telemetry/billing/config), which breaks ZCode's session management. The fix for the user is to install the root CA in the Windows trust store; these two changes make that failure **diagnosable at a glance** and stop a related path-leak.

## Changes

1. **`src/server.ts`** — unknown `/__bili/` or `/__acp/` management paths now return **404 locally** instead of falling through to the proxy and being forwarded to the upstream (which 403s it and leaks the internal path to a third party). The log showed `/__bili/api/sessions` / `/__bili/api/status` being forwarded to `api.anthropic.com` → 403.

2. **`src/mitm.ts`** — when a TLS handshake fails with a cert-rejection alert (`certificate unknown` / `unknown ca` / `bad certificate` / ...), log **one deduplicated, actionable warning** pointing at the root CA file path and the fix (install in the trust store, restart the client), instead of only 160 raw `[info] TLS error` lines. Extraction into `noteMitmTlsError()` keeps it unit-testable.

## Pre-flight
- `npm run typecheck` — clean
- `npm test` — 826/826 pass (added `tests/mitm.test.ts` cert-rejection dedup cases + `tests/admin-origin.test.ts` unknown-`/__bili/`-404 case)
- `npm run build` — success

No `version` bump (content change; release goes through the standard §5 flow).

### DIFF
--- a/src/mitm.ts
+++ b/src/mitm.ts
@@ -168,6 +168,7 @@ function tunnelThrough(
         if (head.length > 0) upstream.write(head);
         upstream.pipe(clientSocket);
         clientSocket.pipe(upstream);
+        log(`tunnel ${maskHostForLog(host)}:${port} established (blind TCP, not decrypted)`);
         const cleanup = (where: string, err: Error) => {
             log(`tunnel ${maskHostForLog(host)}:${port} ${where} closed: ${maskHostInText(err.message, host)}`);
             upstream.destroy();
--- a/src/mitm.ts
+++ b/src/mitm.ts
@@ -1,7 +1,7 @@
 import http from "node:http";
 import net from "node:net";
 import tls from "node:tls";
-import { ensureRootCA, getSecureContext } from "./ca.js";
+import { ensureRootCA, getSecureContext, rootCaPath } from "./ca.js";
 import { connectThroughProxy } from "./upstream-proxy.js";
 import { discoverMitmDomains } from "./discover.js";
 import { isLoopbackAddress } from "./util.js";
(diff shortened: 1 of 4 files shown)


ranxianglei:

[bot] Picked this up — checking out branch `2026-08-29_mitm-ca-warning-and-admin-404` (29cb0e9) and verifying the two changes (admin-path 404 + MITM cert-rejection warning) plus re-running the pre-flight (typecheck / test / build). Will report back with review findings shortly.

ranxianglei:

[bot] Reviewed branch `2026-08-29_mitm-ca-warning-and-admin-404` (29cb0e9). Both changes verified in code and the pre-flight is clean.

**Pre-flight (re-ran locally)**
- `npm run typecheck` — clean
- `npm test` — 826/826 pass, including the 3 new cases (`noteMitmTlsError` dedup + non-cert no-op, unknown-`/__bili/`-404)
- `npm run build` — success

**1. `src/server.ts` — local 404 for unknown admin paths** ✅
- The new `if (isAdminPath) → 404` block sits at `src/server.ts:653`, *after* every real management handler (`/__bili/stats|health|/|config|config/reload|upstream|upstream/test|plugin/*`, lines 526–648) and *after* the loopback + trusted-origin gates (513–525). So it can only fire for genuinely unknown paths, and the security gates still run first — a remote caller hitting an unknown `/__bili/` path still gets 403 (loopback-only), not a forwarded request.
- Confirmed no legitimate endpoint is shadowed: `/__acp/` has **no** handlers anywhere in `src/` (only the `isAdminPath` prefix definition at `src/server.ts:512`), and no `/__bili/` handler exists after line 653. So 404-ing both prefixes locally breaks nothing and stops the path leak to upstream.

**2. `src/mitm.ts` — deduplicated cert-rejection warning** ✅
- `noteMitmTlsError()` (`src/mitm.ts:31`) always logs the raw `TLS error` line, then emits the actionable "install the root CA" warning **once** when `CERT_REJECT_RE` matches. `rootCaPath()` is a pure path join (no file I/O), so it can't throw in the error handler. `maskHostForLog` + `Logger` are already in scope.
- Regex matches the real OpenSSL string from the reporter's log (`ssl/tls alert certificate unknown`) plus the other rejection alerts. Non-cert errors (`socket hang up`, `ECONNRESET`) correctly don't trigger the warning (covered by a test).

**Minor (non-blocking, report only):** the dedup flag `warnedCertRejected` is a module-level singleton, so the actionable warning fires exactly once per proxy *process*. If the user fixes the CA for one client and a *different* client later rejects it, no fresh warning is emitted (the raw `TLS error` line still is). That matches the "warn once" intent and is fine — just noting the tradeoff in case you ever want per-host or per-restart dedup instead.

No `version` bump present (correct — content change; release goes through the §5 flow). Ready to merge when you are — merge is human-only per AGENTS.md, so I'll leave that to you: https://github.com/ranxianglei/billion-context/pull/346

ranxianglei:

[bot] 追加两个改动（都属 #346 诊断/加固，已推送到本分支）：

3. **`src/mitm.ts`** — `tunnelThrough` 现在在盲隧道建立时打 `tunnel <host>:<port> established (blind TCP, not decrypted)`。之前非白名单 host（如 `zcode.z.ai`）连上后在上层失败，日志完全看不到（只记 connect 超时/失败/关闭）。现在能区分「隧道到了、上层失败」vs「根本没到隧道」。

4. **`tests/mitm.test.ts`** — 新增 e2e：盲隧道到非白名单 host **双向**透传字节（本地 TCP server 顶替上游，断言往返 + 新日志）。这正是 `zcode.z.ai` 走的路径，**测试通过** → 证明 bili 盲隧道本身没问题，#346 的 `zcode.z.ai` 连不上是 ZCode/网络侧（修复：ZCode 代理 bypass 掉 `zcode.z.ai`），不是 bili 隧道 bug。

Pre-flight：typecheck 干净，`npm test` 827/827 通过，build 成功。仍无 version bump。

PR #345, opened 2026-08-29
fix: per-route model context declaration outranks the models.dev registry (#344)

ranxianglei:
## Problem (#344)

ZCode (MITM, private relay) on `gpt-5.6-sol` with a route-declared `context: 272000` — but the nudge logs show denominators of **922000** and **988782**, i.e. the proxy sized the window from **models.dev** (`openai/gpt-5.6-sol` = 1,050,000) minus the per-request `max_tokens` headroom (1,050,000 − 128,000 = 922,000; 1,050,000 − 61,218 = 988,782), ignoring the user's declaration.

## Root cause

`src/server.ts` native-window resolution order was:

```
beta ?? plugin ?? launcher ?? peekRegistry ?? configured ?? table
```

The warm models.dev registry peek (`peekRegistryContext`) sat **above** the user's explicit per-route per-model declaration (`resolveConfiguredContextLimit`). The #212 priority change intended "registry > static table" but landed above the whole `configured ?? table` chain. For a private relay host (absent from the registry host table) the lookup falls back to scanning all providers, so the official `openai/gpt-5.6-sol` 1M window was adopted over the relay's real 272K window.

This contradicts the documented contract in `src/config.ts` ("The per-route declaration (which the user controls) always wins, because the same model name can have different windows behind different relays") and `CONFIGURATION.md` ("When a model is not declared, the proxy falls back to its built-in context table or the models.dev registry").

## Fix

Swap the two sources — an explicit declaration now outranks the auto-fetched registry:

```
beta ?? plugin ?? launcher ?? configured ?? peekRegistry ?? table
```

Registry still beats the built-in table when nothing is declared (unchanged behavior).

## Tests

New `tests/configured-window-outranks-registry.test.ts` (regression for #344, exact issue numbers):
1. route declares `context: 272000` + registry lists 1,050,000 → effective window is **272,000** (fails on master: 1,050,000)
2. no declaration → registry 1,050,000 still outranks the built-in table's 400K (pins the registry path)

## Pre-flight

- `npm run typecheck` — clean
- `npm test` — 825/825 pass
- `npm run build` — success

PR #352, opened 2026-08-30
docs: align per-model context declaration precedence with actual resolution order

ranxianglei:
## What

Docs-only fix for the docs-vs-code contradiction surfaced by #351.

`CONFIGURATION.md` (per-model `models` section) claimed the per-model `context` declaration is **primary** and the models.dev registry is only a **fallback for undeclared models**. The code does the opposite: in the native-window resolution chain (`src/server.ts:791-796`), the **warm** models.dev registry (`peekWindow`) is checked **before** the per-model declaration (`configuredWindow`). For a relay/private host whose model is listed in the registry (matched by bare name against provider-prefixed entries via the relay fallback in `src/registry.ts:271-282`), the registry value silently **outranks** the user's declaration.

This is exactly what bit the #351 reporter: their `deepseek-v4-flash` deployment serves a 200K window, but the registry lists 1M, so the effective window resolved to 967232 (1M − 32768 output headroom) and the 75%/95% compression thresholds were unreachable — no compression until the upstream died.

## Change

Rewrite the `models` description in `CONFIGURATION.md` to document the **actual** resolution order (per-request sources → warm registry → per-model `context` declaration → built-in table) and steer users to `compress.modelContextLimit` (highest-priority source, always wins) for reliably pinning the effective window — especially for relay/private deployments where the registry's standard window differs from the served window.

**Behavior-preserving** (markdown only; no code touched).

## Why docs and not code

The code order is deliberate (detailed comment at `src/server.ts:767-784`). Whether the per-model declaration *should* outrank the warm registry (at least for relay/private hosts, where the declaration is the only source of truth about the served window) is a design decision. This PR resolves the contradiction on the docs side; the code-side option (host-dependent precedence) is left to your call — happy to follow up if you want it.

## Minor note (not fixed here)

The committed `package-lock.json` is stale: it records version `0.1.57` while `package.json` is `0.1.63`. Per the release flow the lockfile should be kept in sync — flagging for the next release commit.

## Pre-flight

Docs-only change; no typecheck/test/build impact. Verified the resolution-order claim against `src/server.ts:785-808`, `src/registry.ts:255-284`, and `src/compress-settings.ts:16-25`.

Closes the docs half of #351.

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -146,7 +146,9 @@ A shallow key (`https://open.bigmodel.cn`) matches every path on that host. A de
 - **Type:** `Record<string, { context?: number; output?: number; compress?: CompressSettings }>`
 - **Default:** *(none)*
 - **Status:** ACTIVE
-- **Description:** Maps a model name to its context-window declaration. The LLM `/models` endpoint does **not** return context windows (verified across OpenAI, Anthropic, zhipu, comfly), so the proxy cannot discover them at runtime — you must declare them here. `context` is the model's context window in tokens; `output` is the max output size. When a model is not declared, the proxy falls back to its built-in context table or the models.dev registry. Each model entry may also carry a per-model `compress` block (see [Compression Tuning](#compression-tuning)).
+- **Description:** Maps a model name to its context-window declaration. The LLM `/models` endpoint does **not** return context windows (verified across OpenAI, Anthropic, zhipu, comfly), so the proxy cannot discover them at runtime. `context` is the model's context window in tokens; `output` is the max output size.
+
+  **Resolution order (first match wins):** (1) per-request sources — the client's `anthropic-beta` larger-context negotiation, a cooperative plugin's report, and the launcher's per-model windows; (2) the **warm** models.dev registry cache, when the model is listed (relay/private hosts match the bare model name against the registry's provider-prefixed entries); (3) this per-model `context` declaration; (4) the built-in context table. So when the models.dev registry already lists the model, the registry value **outranks this declaration** — this matters for relay/private deployments, where the registry's standard window can differ from the window the relay actually serves. To pin the effective window reliably, set `compress.modelContextLimit` instead (highest-priority source, always wins). Each model entry may also carry a per-model `compress` block (see [Compression Tuning](#compression-tuning)).
 
 ### `proxy`
 


ranxianglei:

[bot] Reviewed PR #352 (branch `2026-08-30_docs-window-resolution-order`, single commit `675ce55`, `CONFIGURATION.md` only).

**Claims verified against master:**
- `src/server.ts:791-796`: the chain is `betaWindow ?? pluginWindow ?? launcherWindow ?? peekWindow ?? configuredWindow ?? lookupContextLimit(model)` — the warm registry (`peekWindow`) is checked **before** the per-model declaration (`configuredWindow`). ✓
- `src/registry.ts:271-282`: the relay fallback scans `*/<model>` and adopts the value only when every match agrees. ✓
- `compress.modelContextLimit` is the highest-priority source: `resolveRequestConfig` (`src/compress-settings.ts:169-170`) resolves it against the native window, and the comment at `src/server.ts:783` confirms it outranks everything. ✓
- The new text is also consistent with the existing `compress.modelContextLimit` entry (CONFIGURATION.md:195: "highest-priority source … overrides the built-in table, the legacy per-model `context` field"). No other statement in CONFIGURATION.md contradicts the new wording, and the async cold-start registry fetch (`src/server.ts:804-807`) only runs when all sync sources miss, so the documented "first match wins" order holds.

**Lockfile note confirmed:** `package.json` is `0.1.63` while `package-lock.json` records `0.1.57` (top-level and root package entry). Flagged for the next release commit — release commits may carry a drifted lockfile per the release flow.

**⚠️ One interaction to decide before merging:** open PR #345 (from #344 — same root cause as #351) already implements the code-side change this PR defers: it swaps the chain to `configuredWindow ?? peekWindow` (declaration **outranks** the warm registry), with a regression test (`tests/configured-window-outranks-registry.test.ts`). If #345 lands, this PR's load-bearing sentence — "the registry value **outranks this declaration**" — becomes wrong again.

So the two PRs are alternative final states:
1. **Keep the current code order** → merge #352 as-is and close #345.
2. **Adopt the #345 order** (explicit declaration wins — which is what the original docs promised, per the #344 analysis) → merge #345 first, then update this PR's text to the new order (swap items (2)/(3); the "outranks" sentence flips to "this declaration outranks the registry").

Given #345 is already complete with tests and restores the originally documented precedence, my suggestion is option 2. Happy to prepare the updated docs text for the post-#345 order once you confirm the direction.

I can't merge PRs — that's human-only. Merge at: https://github.com/ranxianglei/billion-context/pull/352 (and https://github.com/ranxianglei/billion-context/pull/345 if you go with option 2).

Issue #353, opened 2026-08-30
Auto mode classifier requires confirmation for this command. 3 consecutive actions were blocked.


xlr1012514182:
<img width="1464" height="320" alt="Image" src="https://github.com/user-attachments/assets/80c674e1-413a-41e2-abfb-9540f3d15c95" />
Auto mode classifier requires confirmation for this command. 3 consecutive actions were blocked. Please review the transcript before continuing. Latest blocked action: Auto mode could not evaluate this action and is blocking it for safety - run with--debug for details

This problem occurred after using bili claude. And there can also be exceptions when using sub-agents.

ranxianglei:
ework  接单

ranxianglei:
> 🤖 ework agent · vllm-qwen/qwen3.8-27b

On it. I reproduced the proxy pipeline locally (a mock Anthropic upstream routed through `bili`) and confirmed that **plain one-shot requests pass through unchanged** — streaming and non-streaming, with and without `x-claude-code-session-id`, and even when the session's reported context is over the window (the preflight "stale baseline but payload fits → forward as-is" guard fires correctly). So the failure depends on something specific to Claude Code's auto-mode classifier call that a mock can't reproduce. I have no real model here to test against, so I need your log to pin it down.

**What I need (fastest path):**

1. **The proxy log around the blocked actions.** Default: `~/.local/state/billion-context/bili.log`. Grep for the session id and look for `400`, `upstream`, `preflight`, `no stable conversation identity`, or a long time gap (timeout). This shows the exact forwarded request + upstream status/body.
2. **`claude --debug` output** for one blocked action — the error itself says "run with --debug for details"; the classifier's actual exception (HTTP status / parse error / timeout) will be there.
3. **Fresh session or grown session?** Does it fail on a brand-new conversation, or only after the conversation has grown large? That splits an identity/state problem from an always-on injection problem.

**What I've already ruled out (in isolation):** the `#286` no-identity 400 (a one-shot with a user message gets a `pfa-` session, not a 400) and the preflight fail-fast (guarded when the payload fits).

**Leading hypothesis:** the auto-mode classifier is a throwaway one-shot that still gets bili's compress system-prompt + the 4 ACP tools injected, and (if it shares `x-claude-code-session-id`) is processed against the main session's state. A strict JSON classifier can choke on that. Quick confirm you can run — disable tool/nudge injection for the proxy and see if the classifier recovers (this disables compression globally, so it's a diagnostic, not the fix):

```
[...]

PR #354, opened 2026-08-30
fix: pass through Claude auto-mode classifier requests untouched (#353)

ranxianglei:
## What
`bili claude` broke Claude Code's **auto-mode safety classifier**: it reported "Auto mode could not evaluate this action and is blocking it for safety" and blocked consecutive actions (#353).

## Root cause
The auto-mode classifier is a throwaway one-shot API call (not a compressible conversation). Reverse-engineered from the `@anthropic-ai/claude-code` v2.1.251 binary, its default `xml_2stage` mode expects a **strict XML verdict** and terminates the response with `stop_sequences: ["</severity>"]` (or `"</block>"`), e.g. `<severity>72</severity><category>Data Exfiltration</category>`.

bili's `prepareAnthropic` injects the compress system-prompt + the 4 ACP tools into **every** anthropic request unconditionally (and runs the kernel round-trip). For the classifier that derailed the small model away from the expected XML verdict → "could not evaluate" → blocked actions.

## Fix
Detect the classifier by its in-body signal — `stop_sequences` containing `</severity>` or `</block>` — and forward the request **byte-for-byte**: no compress system-prompt, no ACP tools, no kernel round-trip, and `compressInjected: false` so the response is piped through unrewritten (the verdict reaches the client intact).

Normal conversation requests never carry those stop sequences, so they are unaffected (the existing e2e tests that assert injection on plain turn-1 requests still pass).

## Tests
- New e2e test: a classifier request (`stop_sequences: ["</severity>"]`) reaches the upstream with **no** compress tools and **no** compress prompt appended, and the verdict stream reaches the client unrewritten.
- `npm run typecheck` clean · `npm test` 824/824 · `npm run build` clean.

## Known limitations (follow-up)
- The `twoStageClassifier: "fast"` mode (opt-in) omits `stop_sequences`, so it is not detected by this signal. The default `both`/`thinking` modes are covered.
- The separate "sub-agent exceptions" mentioned in #353 are a distinct path (sub-agents don't use these stop sequences) and are not addressed here.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -1225,6 +1225,11 @@ function prepareAnthropic(
     ++session.stats.requests;
     const injectTools = opts.compress.injectTool && !pluginMode;
 
+    if (isAutoModeClassifier(parsed)) {
+        log("info", `[${sessionId}] auto-mode classifier passthrough (skipping compress injection)`);
+        return { body: JSON.stringify(parsed), session, processedMessages: [], originalMessages: [], anthropicSystem: parsed.system, protocol: "anthropic", stream, compressInjected: false, pluginMode, nudge: undefined, prompts } as Prepared;
+    }
+
     let processedMessages: CoreMessage[] = [];
     let originalMessages: CoreMessage[] = [];
     let nudge: NudgeDecision | undefined;
@@ -1791,6 +1796,20 @@ export function resolvePromptCacheKey(
     if (explicit?.trim()) return explicit;
     if (!identity.clientProvided || !shouldInjectPromptCacheKey(routing, upstream)) return undefined;
     return identity.value;
+}
+
+// Claude Code's auto-mode safety classifier one-shots expect a strict XML
+// verdict — the default `xml_2stage` mode stops the response at `</severity>`
+// or `</block>`. These are not compressible conversations: the compress
+// system-prompt + ACP tools (or the kernel round-trip) derailed the small model
+// from that verdict, so the classifier reported "could not evaluate" (#353).
+// The magic stop sequences are the only in-body signal; forward them untouched.
+const AUTO_MODE_CLASSIFIER_STOPS = new Set(["</severity>", "</block>"]);
+
+function isAutoModeClassifier(parsed: AnthropicRequestBody): boolean {
+    const stops = parsed.stop_sequences;
+    if (!Array.isArray(stops)) return false;
+    return stops.some((s) => typeof s === "string" && AUTO_MODE_CLASSIFIER_STOPS.has(s));
 }
 
 function injectSystem(
--- a/tests/e2e-anthropic.test.ts
+++ b/tests/e2e-anthropic.test.ts
@@ -321,3 +321,42 @@ test("e2e anthropic: compress tool_use round-trip — 2nd upstream request carri
         await h.close();
     }
 });
+
+test("e2e anthropic: auto-mode classifier (stop_sequences </severity>/</block>) bypasses compress injection (#353)", async () => {
+    const h = await startHarness([textScript()]);
+    try {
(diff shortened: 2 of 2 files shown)


ranxianglei:

[bot] On it — reviewing PR #353 (classifier passthrough). Pulling the branch, running typecheck/tests/build, and checking the detection signal against the server pipeline. Will report back with results.

ranxianglei:

[bot] Review of PR #353 (`2026-08-30_passthrough-auto-mode-classifier`, single commit `2af21ab` on current master) — **LGTM**.

**Verified locally** (branch checked out, `npm ci` from committed lockfile):
- `npm run typecheck` — clean
- `npm test` — **824/824 pass** (matches the PR claim), including the new test: `e2e anthropic: auto-mode classifier (stop_sequences </severity>/</block>) bypasses compress injection (#353)` — ok
- `npm run build` — clean
- Existing injection test ("text streams through … injected tools + system reach upstream") still passes, confirming plain turn-1 requests are unaffected.

**Code review** (`src/server.ts`):
- Detection is sound: `isAutoModeClassifier` reads `parsed.stop_sequences` (typed `unknown` via the index signature on `AnthropicRequestBody`, narrowed with `Array.isArray` + `typeof s === "string"` — no `as any`). `</severity>`/`</block>` are specific enough that normal conversation requests (which don't carry stop sequences) can't realistically false-positive.
- Early return in `prepareAnthropic` is placed after `++session.stats.requests` (stats stay accurate) and before `processTurn`/`snapshotMessages`/`markDirty` — so the one-shot **never touches session compression state**. That's the right isolation: the classifier turn can't pollute the real conversation's fold state, and since the plain `pipeThrough` path doesn't sniff usage, `session.stats.lastInputTokens` keeps tracking the real conversation (nudge thresholds unaffected — better than the pluginMode sniff path).
- Response path: `compressInjected: false` → `useRewriter` false → `pipeThrough` verbatim (`src/server.ts:2347-2367`). Verdict reaches the client unrewritten. ✔
- Checked `preflightCompressIfNeeded` (runs after prepare): with empty `processedMessages`, `payloadEstimate = 0`, and `nudge` undefined, it can only hit the "stale baseline … forwarding as-is" branch — the classifier body is never re-processed or fail-fast'ed, even if the co-located real conversation is over the window. ✔
- `JSON.stringify(parsed)` re-serializes rather than forwarding raw bytes, but JSON.parse→stringify preserves every field (order included), so "untouched" holds semantically. Fine.
- The comment block above `AUTO_MODE_CLASSIFIER_STOPS` captures the reverse-engineered `xml_2stage` context that isn't obvious from the code — justified under the no-comments rule.

**Minor observations (non-blocking, no separate issues):**
1. The new test only exercises `</severity>`; a second case (or parameterization) for `</block>` would cover the full detection set.
2. The response-side assertion only checks `message_stop` count. The mock script emits "Hello world!", which would survive an (incorrectly) active rewriter unchanged — so that assertion alone wouldn't catch a broken response path. Sending the actual verdict (e.g. `<severity>72</severity>…`) in the script and asserting it appears in the client stream would make the test airtight on both ends. The request-side assertions (no ACP tools, no compress prompt, stop_sequences verbatim) are the primary defense and they're solid.
3. Cosmetic: when the real conversation on the same session is over the window, the classifier one-shot logs a spurious "preflight trigger fired on a stale baseline" warning. Harmless (forwards as-is), could be suppressed by skipping preflight when `compressInjected === false` && messages empty — optional.

Known limitations from the PR description (`twoStageClassifier: "fast"` opt-in mode not covered; sub-agent exceptions are a distinct path) are acknowledged and reasonable for a follow-up.

Ready to merge whenever you are — I can't merge PRs (system rules), so the merge is yours: https://github.com/ranxianglei/billion-context/pull/353

PR #357, opened 2026-08-30
fix: default upstream proxy to direct; add gost-style -F forward flag (#346)

ranxianglei:
## Summary

Fixes the #346 root cause (silent upstream proxy) and adds an explicit way to set the upstream proxy from the CLI.

### 1. Unset upstream-proxy mode now means **direct**, not env auto-detect

The web UI advertises 直连（默认） for an unset `upstreamProxyMode` (the radio defaults to `direct`), and ZCode's own default is `mode:"direct"` — but the code parsed an unset mode as "no preference" and fell through to env/system proxy auto-detection (`HTTPS_PROXY`/`HTTP_PROXY`/`ALL_PROXY`, then the Windows system proxy).

So a user who never configured a proxy could **silently** have bili route the upstream through whatever proxy was in the environment (e.g. Clash), changing the egress IP. On domestic endpoints with IP-bound OAuth (the #346 ZCode `zcode.z.ai` case) that produces 401s + slowness — and it contradicts what the UI says.

**Behavior change:** unset mode now resolves to `direct` (an empty global proxy short-circuits to direct instead of env discovery).
- To follow the system/env proxy, set mode **`auto`** explicitly (web UI 自动（跟随系统） or `BILI_UPSTREAM_PROXY_MODE=auto`).
- An explicit proxy (`-F` / `BILI_UPSTREAM_PROXY`) still wins regardless of mode. (A top-level `proxy` config still requires mode `auto`/`manual` to take effect — pre-existing, unchanged.)

> Overseas users who relied on the silent auto-detect to reach `api.anthropic.com` should set mode `auto` (or a manual proxy). Per discussion, this is acceptable — the default is now the least-surprise "direct".

### 2. `-F <url>` gost-style forward flag

`bili -F http://host:port <client>` (and `bili -F ... start`) sets bili's upstream proxy for that run, mirroring gost's `-F` (forward). HTTP/HTTPS only for now (SOCKS5 is a follow-up).

bili flags must **precede** the client name — everything after the client name is forwarded to the client. The launcher already isolates bili from the shell's proxy env (`loadOptions(overrides)` uses the overrides as the env), so `-F` is the explicit way to give a launched proxy an upstream.

### 3. `[upstream-proxy]` log line

The resolved upstream-proxy decision is now logged once per unique `(host, proxy, source)`:
```
[upstream-proxy] zcode.z.ai via http://<private-host>:7890 (source=HTTPS_PROXY)
[upstream-proxy] api.anthropic.com direct (source=direct)
```
Non-public hosts are masked (#255). This makes a silently-picked env proxy visible instead of a mystery.

### 4. Help fix

The `bili pi --mitm-domain api.foo.com` example was wrong (that form forwarded `--mitm-domain` to the client). Corrected to `bili --mitm-domain api.foo.com pi`, plus a note that bili flags precede the client name.

## Tests

- `default (unset mode) is direct, not env auto-detect (#346)` — unset mode ignores `HTTPS_PROXY`.
- `explicit 'auto' mode still follows the env proxy (#346 opt-in)`.
- `tests/cli-parseargs.test.ts` — `-F` before/after the client, `start` cmd, composition with other flags.
- Full suite: **829 pass, 0 fail** (was 823; +6 new). Typecheck clean.

## Files

- `src/config.ts` — unset mode → direct; `explicitDirect` refined (`&& !biliProxy`).
- `src/cli.ts` — `-F` flag + help.
- `src/server.ts` — `[upstream-proxy]` log (deduped, masked).
- `tests/upstream-proxy-routing.test.ts`, `tests/cli-parseargs.test.ts` — new tests.
- `CHANGELOG.md` — entries.

## Review follow-ups (non-blocking, from #357 review)

- **`--forward` alias**: dropped from the help/PR copy (only `-F` is implemented, per "暂时只有 -F 即可"); adding the `--forward` long alias is a trivial follow-up if wanted.
- **Top-level `proxy` config doc gap**: a config file with `"proxy": "http://..."` and no `upstreamProxyMode` does not take effect unless mode is `auto`/`manual` (pre-existing; `rawProxyMode` only defaults to `manual` for the web-UI `upstreamProxy` key, not top-level `proxy`). The README's "Upstream proxy" section should note this; making a set top-level `proxy` imply `manual` is a behavior change, so it's deferred to a follow-up.


ranxianglei:

[bot] Reviewed PR #357 (fixes #346). **LGTM** — the core fix is correct and well-tested. Two minor nits, neither blocking.

## Verified

**The #346 bug is real and fixed.** I reproduced it on `master`: with unset `upstreamProxyMode` + `HTTPS_PROXY` in env, `loadOptions` yields `proxy: ""` but `explicitDirect: false`, so `resolveProxyDecision` falls through to env auto-detect and silently routes upstream through the env proxy (`decision: { proxy: "http://env.example:8080/", source: "HTTPS_PROXY" }`). On this branch the same case resolves `{ source: "direct" }`. The mechanism: `parseUpstreamProxyMode(undefined)` already returned `"direct"`, so the old `explicitDirect = ... && rawProxyMode === "direct"` was the only thing keeping an empty global proxy from short-circuiting — the new `effectiveMode = rawProxyMode ?? "direct"` + `explicitDirect = proxyMode === "direct" && !biliProxy` closes that gap. The `!biliProxy` refinement is behaviorally neutral (a set `biliProxy` makes the global proxy non-empty, so the short-circuit at `upstream-proxy.ts:245` is skipped regardless) but makes the intent explicit.

Behavior matrix I probed on this branch (all correct):
- unset mode, no proxy anywhere → `direct` (the fix)
- unset mode + `HTTPS_PROXY` → `direct` (the fix; master: env proxy)
- `BILI_UPSTREAM_PROXY_MODE=auto` + `HTTPS_PROXY` → env proxy, `source=HTTPS_PROXY` (opt-in preserved)
- explicit `direct` + `HTTPS_PROXY` → `direct`
- web-UI `upstreamProxy` config (no explicit mode) → `manual`/`web-manual` (the `webProxy ? "manual"` default still works)
- `mode=auto` + top-level `proxy` config → config proxy, `source=config`

**`-F` flag:** parsed in the value-flag loop (`src/cli.ts:203,220`), requires a non-empty value (exit 2 otherwise), must precede the client name (everything after the first launch-client positional is forwarded — verified by the new `tests/cli-parseargs.test.ts` cases, including `-F` after the client landing in `clientArgs`). The launcher path works because the spawned proxy's env is `{ ...stripInheritedProxy(process.env) }` (`src/launcher.ts:1308`) — the shell's `*_proxy` vars are stripped but `BILI_UPSTREAM_PROXY` (set from the `-F` override into `process.env` at `src/cli.ts:1623`) is inherited. All four new CLI tests + two routing tests pass.

**`[upstream-proxy]` log:** dedup key uses the unmasked host (internal only, never logged); the log line masks both the upstream host (`maskHostPortForLog`) and the proxy URL (`maskUrlForLog`); the Set is bounded in practice by unique (host, proxy, source) tuples. The `resolveProxy` → `resolveProxyDecision().proxy` swap in `buildForwardTarget` is provably equivalent — `resolveProxy` is a one-line wrapper (`src/upstream-proxy.ts:272-279`).

**Help fix:** correct — `bili pi --mitm-domain ...` would indeed have forwarded `--mitm-domain` to the client per the parse loop; `bili --mitm-domain api.foo.com pi` is the working form. Web UI consistency confirmed: the radio defaults to `direct` (`src/web/client.ts:9`, `input.value===(data.upstreamProxyMode||"direct")`), matching the new code default.

**Pre-flight on this branch:** `npm run typecheck` clean; `npm test` → **829 pass, 0 fail** (matches the PR body; +6 = 4 cli-parseargs + 2 routing). E2E could not be run in this review sandbox (no `codex` binary, no local sglang upstream at 127.0.0.1:8199) — noting it since `server.ts` was touched, but the change there is a log line plus the equivalent resolver swap above, and the routing unit suite covers the decision logic deterministically.

## Minor nits (non-blocking)

1. **`--forward` is advertised but not implemented.** The help text says "gost-style -F/--forward" (`src/cli.ts:101`) and the PR body says "mirroring gost's `-F/--forward`", but `parseArgs` only handles `-F` — `bili --forward http://... codex` hits the default case and exits 2 with `unknown option --forward`. Either add the `--forward` alias to the same case or drop it from the help/PR copy.

2. **Top-level `proxy` config is silently ignored unless mode is `auto`/`manual` (pre-existing, but this PR's territory).** A config file with `"proxy": "http://..."` and no `upstreamProxyMode` resolves to `direct` on this branch (and to env auto-detect on master) — the config proxy is never used, because `rawProxyMode` only defaults to `"manual"` for the web-UI `upstreamProxy` key, not for top-level `proxy` (`src/config.ts:294`). The README's "Upstream proxy (firewall / GFW)" section presents top-level `"proxy"` as "Global default: ALL providers route through this proxy" with no mention of needing `upstreamProxyMode: "auto"`, and the new CHANGELOG line "an explicit proxy (`-F` / `BILI_UPSTREAM_PROXY` / config) still wins regardless of mode" is only true for the first two (and for web-UI `upstreamProxy`, which forces `manual`). Since this PR is exactly about silent-proxy surprises, worth closing the gap — either make a set top-level `proxy` imply `manual` (mirroring `webProxy ? "manual"`), or document that it requires mode `auto`/`manual`. I'd lean toward the doc fix in this PR and the code change as a follow-up, since making a stale config proxy suddenly take effect is itself a behavior change.

ranxianglei:
Thanks — both nits addressed:

1. **`--forward` advertised but not implemented** — fixed. Dropped `--forward` from the help text and PR copy (only `-F` is implemented, per the "暂时只有 -F 即可" decision). Adding the `--forward` long alias is a trivial follow-up if wanted.

2. **Top-level `proxy` config silently ignored unless mode is `auto`/`manual`** — agreed, and good catch. Fixed the inaccurate CHANGELOG/PR claim (it now says only `-F` / `BILI_UPSTREAM_PROXY` win regardless of mode; a top-level `proxy` config still requires mode `auto`/`manual`, pre-existing). Per your lean, I'm keeping the code change (making a set top-level `proxy` imply `manual`) as a follow-up since it's itself a behavior change, and noting the README doc gap in the PR body.

Re-pushed; full suite still 829 pass / 0 fail.

ranxianglei:

[bot] Verified the re-push (head `cfb34d1`) — both nits are properly addressed:

1. **`--forward`** — removed from both help-text spots (`src/cli.ts` example line now "gost-style -F", option line now "gost-style;"). No code changes, so no behavior drift.
2. **Proxy-priority claim** — CHANGELOG now reads "an explicit proxy (`-F` / `BILI_UPSTREAM_PROXY`) still wins regardless of mode. (A top-level `proxy` config still requires mode `auto`/`manual` to take effect — pre-existing, unchanged.)" — accurate. The PR body's new "Review follow-ups" section captures the README doc gap and the deferred behavior change (set top-level `proxy` implying `manual`) nicely.

Re-ran on the updated head: `npm run typecheck` clean, `npm test` → **829 pass, 0 fail**.

No further objections from my side — the PR is in good shape for a human to merge whenever. (Reminder per AGENTS.md: the merge itself stays human-only.)

PR #364, opened 2026-08-30
fix: surface raw-dump failures and rate-limit telemetry path noise (#362)

ranxianglei:
Part of #362 (request #4 + the telemetry-noise note). Observability-only changes to `src/server.ts`; no change to forwarding or request behavior.

## What changed

1. **Raw-dump failures are no longer silent.** All five dump write paths (INCOMING / REQ / RES / SSE stream) previously swallowed their `catch {}` (best-effort), so a dump that started failing (disk full, locked dir, EPERM) stopped writing with zero log lines — which is exactly what happened to this user's `raw/` (4,192 `.sse` files, silent stop after 08-29 18:40). They now call `logDumpFailure(where, err)`: first failure logs a `[warn]` immediately, repeats are suppressed within a 60s window, and it re-logs after the window with a cumulative count (`[dump] <where> failed (total Nx): <msg>`). The request is still never broken by a dump failure.

2. **`unrecognized path` telemetry warn is rate-limited per path.** Client telemetry endpoints (e.g. `/api/v1/event/report`) are forwarded unchanged — expected, not an error — but every hit logged a warn (~20k lines in one capture, 14,240 for a single path). Now per masked-path: first 3 hits at `warn`, one `info` "further occurrences suppressed" notice, then silent. (The logger has no debug-level filtering, so rate-limiting is the only way to cut the volume.)

## Notes
- Independent of the acp-kernel PR #172 (persist EPERM data-loss fix); no dependency bump here.
- Content PR — no version bump.
- New test `tests/dump-observability.test.ts` covers both rate-limiters.

## Pre-flight
- `npm run typecheck` — clean
- `npm test` — 845 pass / 0 fail
- `npm run build` — success (dist/index.js 2.55MB)

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -98,7 +98,9 @@ export function logDumpFailure(where: string, err: unknown): void {
 // "suppressed" notice, then silent.
 const unrecognizedPathCounts = new Map<string, number>();
 export function logUnrecognizedPath(log: (level: string, msg: string) => void, url: string): void {
-    const key = maskUrlsInText(url);
+    // Strip the query before masking: a varying query (?ts=…) would otherwise
+    // split one endpoint into unbounded keys and defeat the rate limit.
+    const key = maskUrlsInText(url.split("?")[0]);
     const n = (unrecognizedPathCounts.get(key) ?? 0) + 1;
     unrecognizedPathCounts.set(key, n);
     if (n <= 3) {
--- a/src/server.ts
+++ b/src/server.ts
@@ -77,6 +77,37 @@ function bodyDumpEnabled(): boolean {
     return process.env.ACP_DUMP_BODY === "1";
 }
 
+// Raw dumps are best-effort: a failure (disk full, locked dir, EPERM) must not
+// break the request, but a silently-stopped dump hides real problems (#362).
+// Rate-limited so a stuck dir doesn't spam the log.
+let dumpFailCount = 0;
+let lastDumpFailLog = 0;
+export function logDumpFailure(where: string, err: unknown): void {
+    dumpFailCount++;
+    const now = Date.now();
+    if (dumpFailCount === 1 || now - lastDumpFailLog >= 60_000) {
+        lastDumpFailLog = now;
+        const msg = err instanceof Error ? err.message : String(err);
+        loggerLog("warn", `[dump] ${where} failed (total ${dumpFailCount}x): ${msg}`);
+    }
+}
+
+// Non-protocol paths (client telemetry like /api/v1/event/report, ...) are
+// forwarded unchanged — expected, not an error. Logging every hit spammed the
+// log (~20k lines in one user's capture, #362). Per-path: first 3 at warn, one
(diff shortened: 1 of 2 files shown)


ranxianglei:

[bot] Reviewed `2026-08-31_dump-observability` @ `a2bce38` (checked out locally, diff vs master = `src/server.ts` +45/−7, new `tests/dump-observability.test.ts` +61).

**Pre-flight — all claims verified, reproduced locally:**
- `npm run typecheck` — clean
- `npm test` — **845 pass / 0 fail** (matches claim)
- `npm run build` — success, `dist/index.js` = 2,672,896 bytes (2.55 MB, matches claim)
- New test file alone: 3/3 pass (`logUnrecognizedPath` warn×3→info→silent; per-path independence; `logDumpFailure` first-logs/suppresses/re-logs-after-window with cumulative count)
- Content PR confirmed: no `version` change, no dependency changes (diff touches only the two files)

**Code review:**
- All five raw-dump write paths are instrumented: INCOMING (`src/server.ts:773`), REQ (`src/server.ts:2212`), RES (`src/server.ts:2263`), SSE stream `ws.on("error")` handler and SSE stream outer catch (`src/server.ts:2607`/`2619`). Grep confirms no other `unrecognized path` log site remains — the only occurrences are inside the new `logUnrecognizedPath`.
- Rate-limiter semantics match the issue: `dumpFailCount === 1 || now - lastDumpFailLog >= 60_000` (warn with cumulative `total Nx`); per masked-path `n <= 3` warn / `n === 4` info "suppressed" / silent after. Requests are never broken — all calls are inside `catch`/error handlers.
- Remaining silent `catch {}` blocks in `server.ts` are out of scope and correctly left alone (JSON-body dump `ACP_DUMP_REQ` with its intentional pretty→raw fallback, `mkdirSync` best-effort, error-body reads, JSON parses).

1. In `dumpStreamToFile`, both the `ws.on("error")` handler and the outer catch call `logDumpFailure("SSE stream dump", …)` — a single write failure can be counted twice (error event + catch). Cosmetic only; the count is rate-limited and approximate by design.
2. Edge: if `mkdirSync(ACP_RAW_DUMP_DIR)` fails inside `forward()`, `rawBase=""` silently skips REQ/RES dumps; if the request also has no JSON body, the INCOMING dump is skipped too → zero log lines for that request. Rare in practice (JSON requests are the norm), and the reported incident (SSE dumps stopping) is covered.
3. `unrecognizedPathCounts` grows one entry per distinct masked path — unbounded in theory, bounded in practice (a handful of telemetry endpoints).

**E2E:** skipped — local upstream `127.0.0.1:8199` is not reachable from this environment. The change is observability-only (log calls inside catch blocks + one log call site); no forwarding, routing, or request-pipeline behavior changed, so the 845-test suite + typecheck + build are the appropriate gate here.

**Verdict: LGTM.** Ready for merge — merging is human-only, so please merge yourself: https://github.com/ranxianglei/billion-context/pull/364

PR #350, opened 2026-08-29
fix: actionable no-valid-ranges compress error + defensive object tool args

ranxianglei:
Fixes #349.

Three low-risk changes (no compression-semantics change) addressing the "no
valid ranges" compress error and @liyangbing's runtime-verification
suggestions:

### 1. Actionable error message (`src/stream.ts`)

When `applyRanges` gets 0 ranges it returned:

> `[Compression FAILED: no valid ranges parsed from the tool call. Check your startId/endId parameters.]`

That hint is misleading for the most common cause — the model called `compress`
with **no arguments at all** (`kind=missing-content`, `keys=[]`), not with bad
startId/endId. The new message tells the model exactly what to do so it can
self-correct on the re-request round:

> `[Compression FAILED: no valid ranges parsed. compress requires a non-empty 'content' array of {startId, endId, summary} ranges, where startId/endId are mNNNNN message refs from the conversation. Re-issue the compress call with a valid content array.]`

Keeps the `[Compression FAILED` prefix so `preflight.ts` detection is unaffected.

### 2. Defensive object tool args (`src/loop/adapter-openai.ts`)

The OpenAI adapter only accepted `function.arguments` as a **string**; a
non-standard upstream that emits it as an **object** was silently dropped
(`""` → empty → the same `missing-content` failure). Now `JSON.stringify`
object args. Standard OpenAI (always a string) is unchanged.

### 3. Categorize the reject log + include callId (`src/compress-tool.ts`)

Per @liyangbing's suggestion to separately count the failure modes, the
`[acp-compress-input] rejected` log now carries a `category` field (by `kind`)
and the `callId` (when the path supplies one):

- `empty-call` = `missing-content` / `empty-input` — model empty call
- `stream-concat` = `malformed-json` / `truncated` — adapter-attributable
- `validation` = `no-valid-ranges` / `content-not-array` / `not-object`

So `grep "category=..." bili.log | wc -l` gives per-mode rates with no
session-structure change. With change #2, `missing-content` now maps cleanly to
"model empty call" (the adapter no longer swallows object args).

## Root cause note

Per the log analysis in #349, the primary trigger is **upstream/model** —
deepseek-v4-flash emits an empty `compress` call in some turns (10/12 calls in
the log succeeded, so proxy parsing is sound). These changes are the proxy-side
mitigation: clearer feedback, a hardened arg path, and better observability.

## Session-state safety (verified, no code change)

@liyangbing's concern that a failed compress might replace the original tool
[...]

PR #366, opened 2026-08-31
feat: categorize compress-reject log + include callId (#349 observability)

ranxianglei:
Follow-up to #349 (observability), per @liyangbing's runtime-verification
suggestions. The error-message + defensive-object-args changes already landed
via #350 (merged); this is the remaining observability piece.

## What

The `[acp-compress-input] rejected` log now carries a `category` field (by
`kind`) and the `callId` (when the path supplies one):

- `empty-call` = `missing-content` / `empty-input` — model empty call
- `stream-concat` = `malformed-json` / `truncated` — adapter-attributable
- `validation` = `no-valid-ranges` / `content-not-array` / `not-object`

So `grep "category=..." bili.log | wc -l` gives per-mode rates with no
session-structure change. With #350's object-args hardening, `missing-content`
now maps cleanly to "model empty call" (the adapter no longer swallows object
args), so attribution is unambiguous.

## Why

@liyangbing asked to separately count (1) model empty calls, (2) streaming
parameter concatenation failures, and (3) valid-range validation failures —
only (2) is adapter-attributable. The `category` field makes that split
directly greppable from the log, and `callId` lets a rejection be correlated to
the specific tool call.

## Session-state safety (verified, no code change)

@liyangbing's concern that a failed compress might replace the original tool
result with an empty "success" placeholder does not hold:
- `src/stream.ts` — on `ranges.length===0`, `applyRanges` returns the error
  string without calling `ctx.core.applyCompression`; session state (blocks) is
  untouched. Log: `[acp-proxy: compress call had no valid ranges; nothing compressed.]`.
- `src/loop/core.ts:347-363` — the failed compress is written back as a
  `tool-call` (with the **original** `arguments`) + `tool` (with the error
  result), not an empty placeholder.
- `src/loop/core.ts:373-383` — on `anyCompressFailed`, `hideConsumedCompressCalls`
  is skipped, so the failure stays visible for the model to self-correct.

## Pre-flight

- `npm run typecheck` — clean.
- `npm test` — 850/850 pass.
- `npm run build` — ok.

### DIFF
--- a/src/compress-tool.ts
+++ b/src/compress-tool.ts
@@ -51,7 +51,12 @@ export { ACP_TOOL_NAMES as PROXY_TOOL_NAMES, ACP_MUTATING_TOOLS as MUTATING_PROX
 export function parseCompressInput(input: unknown, callId?: string) {
     const { ranges, diagnostics } = parseCompressArgs(input, { callId });
     if (!diagnostics.ok && diagnostics.kind !== "ok") {
-        loggerLog("warn", `[acp-compress-input] rejected: kind=${diagnostics.kind} invalidItems=${diagnostics.invalidItems}${diagnostics.keys ? ` keys=[${diagnostics.keys.join(",")}]` : ""}${diagnostics.length !== undefined ? ` len=${diagnostics.length}` : ""}`);
+        const category = diagnostics.kind === "missing-content" || diagnostics.kind === "empty-input"
+            ? "empty-call"
+            : diagnostics.kind === "malformed-json" || diagnostics.kind === "truncated"
+                ? "stream-concat"
+                : "validation";
+        loggerLog("warn", `[acp-compress-input] rejected: category=${category} kind=${diagnostics.kind} invalidItems=${diagnostics.invalidItems}${diagnostics.keys ? ` keys=[${diagnostics.keys.join(",")}]` : ""}${diagnostics.length !== undefined ? ` len=${diagnostics.length}` : ""}${callId ? ` callId=${callId}` : ""}`);
     }
     return ranges;
 }


ranxianglei:

[bot] Picked this up. Plan: add the `category` field (by `kind`) + `callId` (when supplied) to the `[acp-compress-input] rejected` log in `src/compress-tool.ts:54`, cover all 7 non-ok `CompressParseKind` values, add tests asserting the mapping + callId presence/absence via `setLogCapture`, then run typecheck/test/build and open a PR. Will follow up with the PR link shortly.

ranxianglei:

[bot] Done — PR opened: https://github.com/ranxianglei/billion-context/pull/368

**What changed**
- `src/compress-tool.ts` — added `REJECT_CATEGORY: Record<CompressParseKind, string>` covering all 7 non-ok kinds:
  - `empty-call` = `missing-content` / `empty-input`
  - `stream-concat` = `malformed-json` / `truncated` (the only adapter-attributable category)
  - `validation` = `no-valid-ranges` / `content-not-array` / `not-object`
  The rejected log line is now:
  `[acp-compress-input] rejected: kind=X category=Y [callId=Z] invalidItems=N [keys=[...]] [len=M]`
  `callId=` is only emitted when the path supplies one — the loop path (`src/loop/core.ts:106`, `executeProxyTool`) does; the legacy stream paths don't pass one, so they log without it.
- `tests/compress-observability.test.ts` — 2 new tests via `setLogCapture`:
  1. each of the 7 kinds logs at `warn` with the correct `kind=` + `category=` pair, and a successful parse logs no rejected line;
  2. `callId=call_abc123` appears when supplied and the field is absent otherwise.

**Pre-flight**
- `npm run typecheck` — clean
- `npm test` — 852/852 pass (850 + 2 new)
- `npm run build` — ok

`grep "category=stream-concat" bili.log | wc -l` now gives the adapter-attributable failure rate directly, and `callId=` correlates a rejection to the specific tool call. No session-state changes (the safety verification in the issue holds as-is).

ranxianglei:
[bot] Closing in favor of #368.

This PR implements the same feature (categorize the `[acp-compress-input] rejected` log + add `callId`) but with an incomplete approach:
- inline nested-ternary for the category (no `REJECT_CATEGORY` map)
- no tests

#368 (`2026-08-31_compress-reject-category`) supersedes it: a `REJECT_CATEGORY: Record<CompressParseKind, string>` map covering all 7 non-ok kinds, `callId=` when supplied, and 2 tests in `tests/compress-observability.test.ts`. Both PRs modify the same log line in `src/compress-tool.ts`, so they conflict — #368 is the one to merge.

Reopen if I've misjudged this.

PR #365, opened 2026-08-30
fix: warn on tool-call XML fragments + close-side truncated acp echo (#361)

ranxianglei:
Fixes #361.

## What changed

The 0.1.63 leak had two fragment classes reaching the final message. This
addresses both, plus the flush-path question:

1. **Non-acp tool-call XML fragments (the main gap).** </invoke>,
   </tool_calls>, </parameter> (and `antml:`-namespaced / JSON-escaped
   forms) matched nothing in the filter and always passed through. Added a
   `containsToolCallXmlFragment` detector for this family and a **once-per-stream
   warn** in the server streaming loop, so a "fake completion" (model wrote the
   tool call as text, no `tool_use` block) is attributable in `bili.log`.
   **Warn-only, not stripped** — these fragments are indistinguishable from
   legitimate prose discussing tool-call code (the #295 review deliberately did
   not strip them; stripping would corrupt real content).

2. **Close-side truncated acp tags.** `TRUNC_OPEN` only handled the open side
   (`<acp …`); a truncated close (`</acp`, `</acp …`) held at
   stream end was re-emitted verbatim. Added `TRUNC_CLOSE` mirroring
   `TRUNC_OPEN` (strip at flush + drop in the streaming filter), and extended
   `PARTIAL_TAIL` / the `definite` check so the close side is held instead of
   leaking.

3. **Regression fix.** A dangling `RENDER_TAG_DETECT` reference (left by an
   earlier uncommitted refactor) broke `tsc`; restored the constant.

## On the "TAG_OPEN_CAP flush path" question (#361 ask 3)

A truncated open **with attributes** (`<acp …`) is already dropped at flush by
`TRUNC_OPEN` — that path works. The residue that survived was the **close side**
(fixed above) and the non-acp tool-call fragments (now warned, ask 1/2). A bare
`<acp` (no space) remains an intentional conservative pass-through (covered
by existing tests) and is not changed here.

## Pre-flight

- `npm run typecheck` — exit 0
- `npm test` — 835/835 pass (833 baseline + 2 new; 1 previously-failing assertion fixed)
- `npm run build` — exit 0

No `version` bump (content branch).

PR #378, opened 2026-08-31
feat: detect + retry fake completions (tool call written as text, #371)

ranxianglei:
## What

Fixes the "fake completion" from #371 (root-cause follow-up to #361): a small model (e.g. MiniMax-M3 via a gateway) can WRITE a tool call as plain text — echoing the tool-call XML template from the context — instead of emitting a real tool block. With no tool block the agent ends the turn early and shows success, but the tool never ran.

This adds a proxy-side **detect + retry** fallback:

- **Detect**: a response carrying a complete tool-call structure (an opening `<invoke>`-style tag, or 2+ distinct closing tags like `</invoke>` + `</tool_calls>`) AND no real tool block for the wire protocol (Anthropic `tool_use` / OpenAI `tool_calls` / Responses `function_call`). Reuses PR #365's `containsToolCallXmlFragment` detector; the structural guard (opening tag OR 2+ distinct closers) avoids false-positives on legit prose that merely mentions a single closing tag.
- **Retry**: re-request upstream with a corrective hint appended as a trailing user message ("you wrote the tool call as text; use the real tool-calling mechanism"). Bounded per turn (max retries) and per session (`session.metadata.fakeCompletionStreak` — consecutive fake-completion turns; resets on a clean turn).

## Opt-in (disabled by default)

The fallback must **buffer the whole response** before the client sees it (a fake completion is only knowable at end-of-stream), which defeats incremental streaming. That cost is only worth paying for the low-frequency fake-completion case, so it is **opt-in**:

- `BILI_FAKE_COMPLETION_RETRIES=2` — enable (2 retries/turn + 2 consecutive-turn cap).
- Default `0` — pre-#371 passthrough (streaming untouched).

This matches the issue's "先收集数据…确认频率后再决定是否上重试（避免为低频问题加复杂逻辑）" guidance: detection is already observable (PR #365's warn), and the retry can be turned on for the models/scenarios where it is confirmed to help.

## Trade-off (documented)

When enabled, streaming responses are buffered before being sent to the client (the client waits for the complete response before seeing tokens). This is the price of being able to retry before the client sees a fake completion.

## Tests

`tests/fake-completion.test.ts` (21 tests):
- Detection per protocol (anthropic/openai/responses): tool-XML + no tool block → fake; tool-XML + tool block → not fake.
- `hasToolCallStructure`: opening tag / 2+ distinct closers → structure; single closer (legit prose) → not.
- `injectFakeCompletionHint` per protocol (merge into last user msg / append).
- Integration (full proxy vs scripted mock upstream):
  - fake → retry → real `tool_use` reaches the client (fake text discarded).
  - retry cap: always-fake upstream → 3 requests turn 1, 6 turn 2, 7 turn 3 (streak cap blocks the retry).
  - legit prose (single closing tag) → NO retry.

## Pre-flight

- `npm run typecheck` — green
- `npm test` — 883 passing (all green)
- `npm run build` — green

Rebased onto latest master (v0.1.66).

### DIFF
--- a/src/fake-completion.ts
+++ b/src/fake-completion.ts
@@ -1,0 +1,124 @@
+import type { WireProtocol } from "./util.js";
+import { containsToolCallXmlFragment } from "./loop/tag-echo-filter.js";
+
+// #371 (root-cause follow-up to #361): a small model can WRITE a tool call as
+// plain text (echoing the tool-call XML template from the context) instead of
+// emitting a real tool block. With no tool block the agent ends the turn early
+// — a "fake completion": the tool never ran but the client shows success. The
+// proxy detects this shape (tool-call structure present + no real tool block)
+// and retries once with a corrective hint, bounded per turn and per session.
+
+export const FAKE_COMPLETION_HINT =
+    "[billion-context] Your last reply described a tool call as plain text instead of invoking it, so no tool actually ran. " +
+    "To act, invoke the tool through the proper tool-calling mechanism (emit a tool_use / tool_calls / function_call block). " +
+    "Do not write tool-call markup as text in your reply.";
+
+// One value serves two bounds: max retries within a single request, and the
+// consecutive-fake-completion-turn cap after which a session stops retrying.
+//
+// DISABLED BY DEFAULT (0): the fallback must buffer the whole response before
+// the client sees it (a fake completion is only knowable at end-of-stream),
+// which defeats incremental streaming. That cost is only worth paying for the
+// low-frequency fake-completion case (small models via gateways), so it is
+// opt-in: set BILI_FAKE_COMPLETION_RETRIES=2 to enable. 0 = pre-#371 passthrough.
+export function maxFakeCompletionRetries(): number {
+    const n = Number.parseInt(process.env.BILI_FAKE_COMPLETION_RETRIES ?? "0", 10);
+    return Number.isFinite(n) && n >= 0 ? n : 0;
+}
+
+// OOM guard for a pathological upstream; LLM responses are bounded by max_tokens
+// and normally well under 1 MiB.
+export function fakeBufCap(): number {
+    const n = Number.parseInt(process.env.BILI_FAKE_BUF_CAP ?? String(16 * 1024 * 1024), 10);
+    return Number.isFinite(n) && n > 0 ? n : 16 * 1024 * 1024;
+}
+
+// "Does this raw response (full SSE stream or JSON body) carry a REAL tool
+// block?" A real block means the model actually invoked a tool → not a fake
+// completion. Each protocol marks its tool block differently on the wire.
+const ANTHROPIC_TOOL_BLOCK = /"type"\s*:\s*"tool_use"/;
+const OPENAI_TOOL_BLOCK = /"tool_calls"\s*:\s*\[\s*\{/;
+const RESPONSES_TOOL_BLOCK = /"type"\s*:\s*"function_call"/;
+
+export function hasToolBlock(protocol: WireProtocol, rawText: string): boolean {
+    switch (protocol) {
+        case "anthropic":
+            return ANTHROPIC_TOOL_BLOCK.test(rawText);
+        case "openai":
+            return OPENAI_TOOL_BLOCK.test(rawText);
+        case "responses":
+            return RESPONSES_TOOL_BLOCK.test(rawText);
+    }
+}
+
+// Opening tool tag (no leading '/'): the model started writing a call as text —
+// the strongest signal. Literal '<' or JSON-escaped '\u003c'; optional antml:
+// namespace; case-insensitive, matching containsToolCallXmlFragment.
+const TOOL_OPEN =
+    /\x3c(?!\s*\/)(?:antml:)?(invoke|tool_calls|tool_call|parameter|parameters)\b|\\u003c(?!\s*\/)(?:antml:)?(invoke|tool_calls|tool_call|parameter|parameters)\b/i;
+
+// Closing tool tags (global): an echoed template tail carries 2+ distinct
+// closers (</invoke> + </tool_calls>) even when the openers were folded away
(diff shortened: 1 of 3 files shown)


ranxianglei:

[bot] Picked up #378. Reviewing the `2026-08-31_fake-completion-retry` branch now: checking out the diff, re-running the pre-flight (typecheck / test / build), and reviewing the detect+retry logic for correctness. Will report findings here.

ranxianglei:

[bot] **Review complete — LGTM, no blocking issues.** Reviewed the diff on `2026-08-31_fake-completion-retry` (commit `5cccaa7`, 3 files) and re-ran the pre-flight myself.

**Pre-flight (re-run on the branch, all match your claims):**
- `npm run typecheck` — green
- `npm test` — 883 pass / 0 fail
- `npm run build` — green (`dist/index.js` 2.56 MB)
- `tests/fake-completion.test.ts` — 21/21 pass

**What I verified in the code:**
- **Retry invariant holds**: the retry re-sends exactly what the original request sent — same `prepared.body` (the body the original fetch uses at `src/server.ts:2256`), same `upstreamUrl`, same `proxyDispatcher`, same `clientAbort.signal`, same 10-min default timeout. Headers are rebuilt via `buildForwardHeaders`, which strips the stale `host`/`content-length` before the hinted body (different length) — correct.
- **Placement is right**: the buffer+retry sits after the `!upstream.ok` early return (`src/server.ts:2408`), the empty-body return (2418), and the plugin-mode verbatim return (2435) — so it only touches 2xx, non-plugin, client-streaming responses, exactly as documented.
- **Detection logic checks out against the real gate**: `containsToolCallXmlFragment` (`src/loop/tag-echo-filter.ts:64`) passes both opening *and* closing tags, so the structural guard (opening tag OR 2+ distinct closers) is precisely the narrowing that excludes legit single-closer prose. The `hasToolBlock` regexes match the actual wire shapes (Anthropic `content_block_start` `"type":"tool_use"`, OpenAI first `tool_calls` delta, Responses `function_call` item).
- **Streak semantics match the tests** (max=2 → 3/6/7 requests across turns 1–3); the read-modify-write of `session.metadata.fakeCompletionStreak` is serialized under `withSessionLock` (`src/server.ts:1158`) and persisted via `markDirty`.
- **Hint injection** handles all three protocols, string/array content shapes, and Anthropic's user/assistant alternation (merge into last user msg vs. append after assistant).
- Code quality per AGENTS.md: no `as any`/`@ts-ignore`, hex-escaped XML, `opts.log` (no `console.error`).

**Minor findings (report-only, non-blocking):**
1. **Silent truncation when a response exceeds `BILI_FAKE_BUF_CAP` (default 16 MiB)** — confirmed by repro: `readStreamToBuffer` (`src/server.ts:2623`) does not throw on overflow; it retains the first `maxBytes` and drains the rest. With the feature enabled, an 8810-byte SSE response under a 1024-byte cap reached the client as exactly 1024 bytes, cut mid-event with no `message_stop`. At the default 16 MiB cap this needs a >100K-token-class response (SSE overhead ≈ 100 B/token), so it's rare — but it's a *new* failure mode introduced by the buffering (pre-PR the success path streamed directly), and it fails silently (corrupted stream) rather than loudly. Suggested fix: detect overflow and fail loudly (502) instead of serving truncated bytes, or raise the default cap well beyond any realistic LLM response.
2. **Non-streaming responses are out of scope**: the `prepared.stream` gate (`src/server.ts:2440`) skips `stream:false` JSON responses. Buffering costs nothing there (the client already waits for the full body), so covering them would be nearly free. Small gap — agents stream by default.
3. **Retry fetch skips `recordUpstreamConnection` telemetry** — the original fetch records it (`src/server.ts:2270`), the retry (`src/server.ts:2683`) doesn't, so retry traffic is invisible to connection-status telemetry.
4. (Footnote) `markDirty` is called unconditionally even when the streak is unchanged (0→0) — an extra persist per streaming request when enabled; harmless since it's debounced.

**E2E regression**: per AGENTS.md, changes to the request pipeline (`server.ts`) should run `tests/e2e/e2e-codex.test.ts` before merge. I could not run it in this environment — no `codex` binary on PATH and the default upstream `http://127.0.0.1:8199/v1` is unreachable (probe `000`); the `E2E_CHECK=1` preflight also needs the codex binary. Please run the 4-phase core locally (or via the manual-dispatch CI) before merging.

ranxianglei:
Pushed to this branch (merge master + one fix), same drill as #464:

1. **Merged post-#466 master** in — textual auto-merge was clean, but **semantically broken**: the #460 non-inject SSE strip branch (`forward()`) read `upstream.body` directly, which `resolveFakeCompletion` has **already consumed** when `BILI_FAKE_COMPLETION_RETRIES > 0` — that path would hand the client an exhausted stream (empty response / hang).
2. **Fixed** in 892ffa2: the strip branch now reads `responseBody` (the buffered stream), identical to `upstream.body` when retries are disabled — zero behavior change at default `0`.

Verified on the merged tree:
- `npm run typecheck` ✓
- `tests/fake-completion.test.ts` + `tests/proxy-noinject-tag-strip.test.ts`: 24/24 ✓
- full suite 942/945 (3 failures are pre-existing on master HEAD: 2 permanent env failures in `plugin-agent.test.ts` + the #466 call-order breakage fixed by pending #472)

One design gap noted for follow-up (not blocking this PR): the fake-completion gate sits **after** the plugin-mode early return in `forward()`, so plugin-mode streams (omp/pi — the original reporters of #361/#371) are not covered. Tracking issue to follow.

ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] Verifying the new push (master merge + 892ffa2): checking out the branch, reviewing the strip-branch fix, and re-running the pre-flight (typecheck / targeted tests / full suite) plus confirming the 3 suite failures are pre-existing on master HEAD. Will report here.

PR #379, opened 2026-08-31
fix: merge all system/developer messages into one leading message (SGLang #377)

### LINKED ISSUE #377: [Bug] ACP: ≥2 compressed blocks render as multiple system messages → SGLang (vLLM-compatible) upstreams return 400
### Bug description

When a session has **2 or more active ACP compressed blocks**, bili renders each block's
summary as its **own `system` message** (one `[Compressed conversation section]…` per block).
OpenAI-compatible backends that enforce "system message only at index 0" — SGLang in particular —
reject the request with a hard 400, so **the session becomes unusable after the second fold**.

- Anthropic upstream: unaffected (multiple system blocks are legal there).
- Single-block sessions: work (one leading system message).
- This means the core ACP use case (repeated folding over long sessions) silently breaks on
  any SGLang/vLLM-style OpenAI-compatible endpoint.

### Exact error

Upstream (SGLang) response:

```json
{"object":"error","message":"System message must be at the beginning.","type":"BadRequest","param":null,"code":400}
```

bili log line:

```
[warn] [pfa-...] ← upstream 400: {"object":"error","message":"System message must be at the beginning.","type":"BadRequest","param":null,"code":400}
```

Reproduction: run a session with a 228K-context OpenAI-compatible model behind bili, trigger
`compress` twice (two blocks), then send any follow-up request → 400. Every subsequent request
fails the same way; the session cannot recover without deleting blocks.

### Suggested fix

Before forwarding, merge **all** `system`/`developer` messages (head or mid-position) into a
single leading system message. This is behavior-preserving for Anthropic (their API tolerates
multiple system messages, and concatenation is a documented equivalence) and required for
SGLang/vLLM-class upstreams.

Minimal patch applied locally (in `dist/index.js`, `hoistMidSystemMessages`):

```js
// PATCH 2026-08-31: SGLang rejects multiple system messages ("System message must be
// at the beginning"). bili renders each compressed block as its own system summary;
// merge ALL system/developer messages (head or mid) into one leading message.
const sys = [], rest = [];
for (const m2 of messages) {
  if (m2.role === "system" || m2.role === "developer") sys.push(m2);
  else rest.push(m2);
}
if (sys.length === 0) return messages;
if (sys.length === 1 && messages[0] === sys[0]) return messages;
const merged = sys.map((m2) => typeof m2.content === "string" ? m2.content : JSON.stringify(m2.content))
  .join("\n\n---\n\n");
return [{ role: "system", content: merged }, ...rest];
```

Verified: after the patch, a session with 4 compressed blocks folds and recalls correctly
(118K tokens shrunk 92%, 5/6 verbatim retrieval via decompress+read on a 120K-context session).

### Optional hardening

- Per-provider capability flag (e.g. `systemMessageLimit: "leading-only"`) instead of always
  merging, in case some upstreams do distinguish multiple system messages.
- An integration test with a mock OpenAI-compatible server that enforces the index-0 rule.

### Environment

- billion-context 0.1.65 (npm global)
- Node v24.16.0, Linux x86_64
- Upstream: SGLang server, model Qwen/Qwen3.8-27B-FP8, 228K context, OpenAI-compatible API

ranxianglei:
## Summary

Fixes #377 — with **2+ active ACP compressed blocks**, bili rendered each block's summary as its own `role:"system"` message, so the OpenAI wire carried 2+ system messages. SGLang (and vLLM with the Qwen3-family "system-first" chat template) reject that with a hard 400:

```
{"object":"error","message":"System message must be at the beginning.","type":"BadRequest","code":400}
```

The earlier #355 fix (`hoistMidSystemMessages`) hoisted mid-conversation systems to a *contiguous leading prefix* but kept them **separate** — which SGLang still rejects, because its Qwen3 template guard is `if message.role == 'system' and not loop.first: raise`, i.e. it requires **exactly one** system message at index 0, not merely a leading prefix.

## Change

Replace `hoistMidSystemMessages` with `mergeSystemMessages` (`src/util.ts`): it collapses **all** system/developer messages into a **single** leading `system` message, preserving the relative order of every other message. `sysParts` (the client's original system + the static compress prompt) is placed **first** so the stable prefix-cache anchor stays at the head across compress turns; the folded block summaries follow in their original anchor order.

- `src/server.ts` (`prepareOpenai`) and `src/loop/adapter-openai.ts` (`buildRequest`) now call `coreToOpenai(...)` then `mergeSystemMessages(msgs, sysParts)`; the now-unused `injectOpenaiSystem` import is dropped from both.
- The Responses API path is unaffected (it already injects a single joined developer message).

## Tests

- `tests/hoist-mid-system.test.ts` → `tests/merge-system.test.ts`: unit tests rewritten for the merge (no-op, single/mid/multiple systems, sysParts-first ordering, the #377 wire shape collapsing to exactly one system, non-string content JSON-stringified).
- `tests/e2e-mid-system.test.ts`: the real-proxy e2e now asserts **exactly one** system/developer message at index 0 in every upstream request (the #377 requirement), in addition to the existing summary-marker checks.

## Pre-flight

- `npm run typecheck` — clean
- `npm test` — 863 pass, 0 fail (incl. the e2e #355 real-proxy run)
- `npm run build` — success


ranxianglei:

[bot] Reviewing this PR now. Diff reviewed: `mergeSystemMessages` in `src/util.ts` replaces `hoistMidSystemMessages`, both call sites (`prepareOpenai` in `src/server.ts`, `buildRequest` in `src/loop/adapter-openai.ts`) updated, unused `injectOpenaiSystem` imports dropped, tests rewritten + e2e assertion tightened to "exactly one system at index 0". Running the full pre-flight (typecheck / test / build) on the branch now — will report results here.

ranxianglei:

[bot] Review complete — verified independently on branch `2026-08-31_merge-system-messages` (57cd966). All pre-flight claims check out:

**Pre-flight (re-run locally)**
- `npm run typecheck` — clean
- `npm test` — **863 pass, 0 fail** (matches the PR claim)
- `npm run build` — success
- `tests/e2e-mid-system.test.ts` (real-proxy e2e) — passes standalone with the new "exactly one system/developer at index 0" assertion on every upstream request

**Code review**
- `mergeSystemMessages` (`src/util.ts:194`): correct — collects all system/developer messages, `sysParts` first (stable prefix-cache anchor), block summaries in original anchor order, non-system relative order preserved, no-op (same array ref) when there's nothing to merge, non-string content JSON-stringified. The `"\n\n---\n\n"` separator and the sysParts-first ordering match the stated cache-stability rationale.
- Both call sites updated (`prepareOpenai` in `src/server.ts:1442`, `buildRequest` in `src/loop/adapter-openai.ts:176`); unused `injectOpenaiSystem` imports dropped from both. No lingering references to `hoistMidSystemMessages` anywhere. Remaining `injectOpenaiSystem` hits are only kernel-API unit tests in `tests/basic.test.ts` / `tests/proxy-basic.test.ts`, which are legitimate.
- Tests: `tests/merge-system.test.ts` covers no-op, single/mid/multiple systems, sysParts-first, the #377 wire shape collapsing to exactly one system, and non-string content. E2e assertion correctly tightened from "contiguous leading prefix" to "exactly one at index 0" — that is sglang's actual guard.
- Responses API path untouched, as claimed.

**One caveat**
- The full codex e2e regression (`tests/e2e/e2e-codex.test.ts`) could not be run in this environment: no `codex` binary and no reachable upstream (`127.0.0.1:8199` refused, `E2E_UPSTREAM_URL` unset). Per AGENTS.md it should be exercised before merging request-pipeline changes — the PR's own real-proxy e2e covers the OpenAI wire path through the actual server code, but if you have a codex-capable box, a run of the 4-phase core there would be the last mile.

LGTM from my side — merge is yours to make (I can't merge PRs).

PR #384, opened 2026-08-31
docs: per-model context declaration now outranks the registry (match #345)

ranxianglei:
## What

One-paragraph fix to `CONFIGURATION.md` (the `providers.<route>.models.<model>` description).

## Why

PR #345 (shipped in v0.1.65) reordered the window-resolution chain in `src/server.ts` so the per-model `context` declaration (`configuredWindow`) is now checked **before** the warm models.dev registry (`peekWindow`):

```js
let native = betaWindow
    ?? pluginWindow
    ?? launcherWindow
    ?? configuredWindow   // per-model declaration  (2)
    ?? peekWindow         // models.dev registry    (3)
    ?? lookupContextLimit(model);
```

But the docs paragraph I added in #352 (also merged in v0.1.65) still documents the **old** order and explicitly states "the registry value **outranks this declaration**" — the exact opposite of what the code now does. This is the same confusion at the heart of issue #351, so leaving it stale would keep misdirecting relay/private-deployment users.

This PR swaps (2)/(3) and rewrites the guidance: the per-model `context` declaration now **outranks the registry** (set it to the window your deployment actually serves and it wins even when models.dev lists a larger window); `compress.modelContextLimit` remains the highest-priority source for pinning the window across every route.

## Scope

- Docs only, behavior-preserving. One paragraph in `CONFIGURATION.md`.
- No other doc statement references the old order (grep-verified).

## Pre-flight

- No code changed; typecheck/test/build unaffected (docs-only commit).


ranxianglei:

[bot] 🏷 Done — PR opened: https://github.com/ranxianglei/billion-context/pull/385

**What I changed** (one paragraph, `CONFIGURATION.md:151`):
- Swapped the resolution-order entries: **(2)** is now this per-model `context` declaration, **(3)** is the **warm** models.dev registry cache.
- Rewrote the guidance to match the code: the per-model `context` declaration now **outranks the registry** — "set it to the window your deployment actually serves and it wins even when models.dev lists a larger window."
- Kept `compress.modelContextLimit` as the highest-priority source for pinning the window across every route.

**Verification**
- Confirmed the actual code order in `src/server.ts:837-842`: `betaWindow ?? pluginWindow ?? launcherWindow ?? configuredWindow ?? peekWindow ?? lookupContextLimit(model)` — `configuredWindow` is checked **before** `peekWindow`. The inline comment at `src/server.ts:820-828` states it explicitly ("an explicit declaration always outranks the auto-fetched registry (#344)").
- Grep-verified no other doc statement references the old order. The zh-CN version (`CONFIGURATION.zh-CN.md:149`) already documents the new order ("when the model is not declared, the proxy falls back to the built-in table or registry"), so no change was needed there.

Docs-only, behavior-preserving; typecheck/test/build unaffected.

PR #383, opened 2026-08-31
release v0.1.67

ranxianglei:
## v0.1.67

Version bump only (isolated release commit). CI will build, test, publish to npm, tag `v0.1.67`, and create the GitHub Release on merge.

### Changes since v0.1.66
- **fix: Windows overlay symlink EPERM fallback + fail-loudly + SQLite set merge** (#381, PR #382)
  - `bili omp`/`pi`/`hermes`/`dsh` overlay now works on unprivileged Windows (no Developer Mode): `symlinkSync` `EPERM` falls back to junction (dir) / hardlink / copy (file).
  - Hollow overlay (no entry resolvable) now fails loudly and aborts the launch instead of silently pointing the client at an empty home — the Windows root cause of #289.
  - SQLite three-piece set (`.db`/`.db-wal`/`.db-shm`) moves as a unit with rollback; write-through hardlinks are re-pointed, never merged back.

### Pre-flight (matches CI)
- `npm run typecheck` ✅
- `npm test` — 868 pass, 0 fail ✅
- `npm run build` ✅

PR #382, opened 2026-08-31
fix: Windows overlay symlink EPERM fallback + fail-loudly + SQLite set merge

ranxianglei:
## Fixes #381 (the Windows root cause of #289)

Unprivileged Windows processes lack `SeCreateSymbolicLinkPrivilege`, so `fs.symlinkSync` throws `EPERM`. `refreshOverlayHome` wrapped every link in a bare `catch {}` and **returned `true` unconditionally** — so the `bili omp`/`pi`/`hermes`/`dsh` overlay came out hollow (only the generated config + pid marker, zero links) and the client read an empty home, launching fresh/unconfigured every time.

### Changes (`src/launcher.ts`)
- **`linkOverlayEntry`** — on `EPERM`/`EACCES`/`EINVAL`, fall back per entry type: dir → `symlink 'junction'` (no privilege needed), file → hardlink (same volume) → copy. The overlay now works without Developer Mode.
- **`refreshOverlayHome`** — track link success/failure per entry. When the overlay comes out hollow (no entry resolvable), log a loud warning and `return false` so the launcher aborts instead of pointing the client at an empty home. Partial failures are reported, not silently dropped. (Also fixed an `EEXIST` bug: entries already correctly linked by the per-entry loop / left by a SQLite rollback are counted as present, not re-linked.)
- **`mergeSqliteSet`** — move the SQLite three-piece set (`.db`/`.db-wal`/`.db-shm`) as a unit with two-phase rollback, so a locked/half-migrated set can no longer corrupt the real home db.
- A write-through hardlink (`nlink>1`, same `dev`+`ino` as the real file) is re-pointed on refresh, never merged back (data-loss guard).

### Tests (`tests/launcher.test.ts`, +6)
Simulate the Windows `EPERM` condition on Linux by denying `fs.symlinkSync` for non-junction links: dir→junction, file→hardlink, file→copy, all-fail→hollow→`undefined`, SQLite set unit-move, SQLite set rollback, write-through hardlink re-point.

### Pre-flight
- `npm run typecheck` ✅
- `npm test` — 868 pass, 0 fail ✅
- `npm run build` ✅

Note: I could not exercise a real Windows host from this environment; the fallback paths are unit-tested by simulating the `EPERM`/`EACCES` behavior on Linux. A real-Windows smoke test of `bili omp` is recommended before release.

PR #396, opened 2026-08-31
fix: #381 review — SQLite set adjudication, -journal, conflict rename, link kind

ranxianglei:
Follow-up to #382 (the Windows overlay symlink fix), addressing the code review in #381 (FLOOR 10).

## What changed

**1. `mergeSqliteSet` — adjudicate by the main db, not per-member mtime**

The original per-member mtime adjudication could splice a newer main db with a newer WAL from the *other* side — e.g. real `x.db` newer than overlay `x.db`, but overlay `x.db-wal` newer than real `x.db-wal`. SQLite validates the WAL's salt/page against the main db; a mismatched pair is silently discarded, so the overlay session's uncheckpointed writes were lost.

Now the whole set is adjudicated **once** by the main db's mtime; the winner's set moves as a unit and every loser member is preserved as a `.bili-conflict`. Orphan sets (no main db, only sidecars) are treated as stale residue and discarded — never moved individually.

**2. `sqliteSetMembers` — include `-journal`**

Rollback-journal mode uses `x.db-journal`, which was missing from the set and would have been merged separately (the same cross-side hazard).

**3. Conflict renames — non-colliding names**

`renameSync(src, dst.bili-conflict)` silently overwrites an existing same-name target, so a retry round (e.g. after a partial merge) clobbered the previous round's preserved loser. Both `mergeSqliteSet` and `mergeOverlayEntry` now use `freeConflictName` (`.bili-conflict`, `.1`, `.2`, …).

**4. `linkOverlayEntry` — explicit symlink kind**

The first attempt used `fs.symlinkSync(target, link)` with no type, which on Windows creates a *file*-tag reparse point for a directory target (libuv's type-omitted default is `'file'`) — something `readdir(withFileTypes)` and some backup/indexers may not follow. Now the kind is explicit: win32 dir → `'junction'` (no privilege, unambiguous), file → `'file'`.

## Tests

3 new tests (main-db adjudication no-splice, `-journal` unit, non-overwriting conflict rename) + the 6 existing #381 tests. 871 pass, typecheck + build green.

## Release note

Please merge this **before** the v0.1.67 release PR (#383) so the release ships the review fixes. If v0.1.67 has already shipped, this can go out as v0.1.68.


ranxianglei:

Verified this PR end-to-end on `2026-08-31_windows-overlay-review-fixes` (a2b2822). All pre-flight checks are green and it merges cleanly into `master`.

**Pre-flight (re-run locally):**
- ✅ `npm run typecheck` — clean
- ✅ `npm test` — **871 pass / 0 fail / 0 skipped**
- ✅ `npm run build` — success (self-contained `dist/index.js`)
- ✅ `git merge-tree` vs `master` — no conflicts (the branch doesn't touch `package.json`, so the already-landed v0.1.67 bump merges cleanly)

**Code review — all 4 fixes from the #381 FLOOR 10 review are correctly implemented:**
1. `mergeSqliteSet` (`src/launcher.ts:750`) — the whole set is now adjudicated **once** by the main db's mtime (`src/launcher.ts:763`), and the winner's members move as a unit via `movePreserving`. A newer WAL/journal from the losing side is preserved as `.bili-conflict`, never spliced in against the winning db. Orphan sets (no main db on either side) are preserved wholesale as conflicts, never moved in as an active db. I hand-traced the cross-generational splice case (newer main db + newer WAL from the other side) — it's no longer reachable.
2. `sqliteSetMembers` (`src/launcher.ts:716`) — now includes `${base}-journal`.
3. `freeConflictName` (`src/launcher.ts:724`) — non-colliding `.bili-conflict`, `.1`, `.2`, …; used by both `mergeSqliteSet` and `mergeOverlayEntry`, so a retry round can't clobber the previous round's preserved loser.
4. `linkOverlayEntry` (`src/launcher.ts:659`) — explicit reparse kind: win32 dir → `junction`, file → `file`; non-win32 → plain symlink.

The 3 new tests (main-db adjudication no-splice, `-journal` unit, non-overwriting conflict rename) plus the 6 existing #381 tests all pass.

**Release:** v0.1.67 has already shipped (master is at the #383 merge), so per the release note this goes out as **v0.1.68**. Merging PRs and cutting releases are human-only on my side — please merge #396, then cut the `v0.1.68` release branch.

**Optional nit (non-blocking):** the `if (st.isDirectory()) { symlinkSync(target, link, "junction") }` fallback at `src/launcher.ts:678-685` is now dead code — the first attempt already uses `junction` for win32 dirs, so that block is unreachable on every path. Harmless; can stay or be trimmed in a follow-up.

PR #423, opened 2026-08-31
fix: drive task-continuity across the compression boundary (#422)

ranxianglei:
## Problem

#422: an ACP compression round ends the model turn with `finishReason=stop` and **zero toolCalls**, interrupting an in-progress task — the user must manually type "continue" to resume the tool loop. Clean stop, no fragment leak, upstream API normal (distinct from #361's torn-state fragment leak).

The compression boundary is where the model's task-continuity awareness is weakest: the context was just replaced and the model just produced a natural "deliverable" (the compress summary), so without an explicit drive it stops and treats the summary as the turn's end.

## Fix

In `runCompressLoop` (src/loop/core.ts), when the loop re-requests after a **context-mutating** proxy tool (`compress`/`decompress`, i.e. `MUTATING_PROXY_TOOLS`), push a one-shot task-continuity driver into the re-request's `coreMessages`:

> `[ACP] A context-management step (compress or decompress) just ran — that was housekeeping, not your task. Continue with your next planned action (the next tool call in your plan). If your task is in fact already complete, stop normally.`

Design notes:
- **Proactive** (user suggestion 1) rather than reactive: no wasted round, directly addresses the root cause (lack of drive at the compression boundary).
- **Outcome-neutral text**: true for compress-success, compress-failure, and decompress alike (the gate is `MUTATING_PROXY_TOOLS`, which includes decompress and failed compress — so the text must not assert "compression completed").
- **Right-to-terminate**: grants the model the option to stop normally if the task is actually complete — guards the reverse risk (over-driving a finished turn) without a fragile proxy-side "is the task in progress?" heuristic.
- **Scoped to mutating tools**: read-only queries (`acp_status`/`search_context`) don't replace the context, so the model continues naturally without a driver.
- **Transient**: the driver lives in the loop-local `coreMessages` (never persisted to session state), deduped by a fixed id (`acp_loop_continue_driver`), injected once per request, survives per-round `hideConsumedCompressCalls` (which only touches `compress` tool-call records). Sent to the model in the re-request, never emitted to the client.
- `role: "user"` + `contentType: "text"` is the safe role across all three wire protocols.

## Tests

- `loop #422a`: compress round → re-request body carries the continue-driver (asserts the outcome-neutral text + the right-to-terminate clause).
- `loop #422b`: read-only `acp_status` round → re-request body does NOT carry the driver (scoping control).

## Pre-flight

- `npm run typecheck` — clean
- `npm test` — 898 pass, 0 fail (in this environment; includes 49 loop tests + 2 new)
- `npm run build` — success
- E2E (`tests/e2e/e2e-codex.test.ts`): could not run in this environment — the `codex` binary is unavailable. Recommend running the E2E 4-phase core on a machine with `codex` + a Responses-compatible upstream before merge.

## Review iteration

Addressed @ranxianglei's review: (1) softened the driver text to be outcome-neutral (no longer asserts "compression completed", which was false for decompress + failed-compress paths); (2) replaced the forceful "do not stop" with a right-to-terminate clause (reverse-risk guard); (3) rebased onto current master (`5fe15a2`, includes #431/#433).

### DIFF
--- a/src/loop/core.ts
+++ b/src/loop/core.ts
@@ -13,6 +13,7 @@ import type { BiliMessage } from "acp-kernel/wire";
 import {
     parseCompressInput,
     PROXY_TOOL_NAMES,
+    MUTATING_PROXY_TOOLS,
 } from "../compress-tool.js";
 import { applyRanges } from "../stream.js";
 import { resolveDecompress } from "../decompress-shared.js";
@@ -23,6 +24,17 @@ import { log as loggerLog } from "../logger.js";
 import type { WireProtocol } from "../util.js";
 
 export const MAX_LOOP_ROUNDS = 10;
+
+// #422: one-shot task-continuity driver for the re-request after a
+// context-mutating proxy tool — the compression boundary is where the model's
+// task-continuity is weakest (it stops, interrupting the in-progress task).
+// Injected into the re-request only; never persisted to session state. Text is
+// outcome-neutral (true for compress-success, compress-failure, and decompress
+// alike) and grants the model the right to stop if the task is actually done
+// (guards the reverse risk: over-driving a finished turn).
+export const CONTINUE_DRIVER_TEXT =
+    "[ACP] A context-management step (compress or decompress) just ran — that was housekeeping, not your task. " +
+    "Continue with your next planned action (the next tool call in your plan). If your task is in fact already complete, stop normally.";
 
 function isLoopThinking(m: CoreMessage): boolean {
     return m.contentType === "reasoning" && typeof m.id === "string" && m.id.startsWith("acp_loop_");
@@ -423,6 +435,21 @@ export async function* runCompressLoop(
                 loggerLog("warn", `[acp-loop] loop limit (${MAX_LOOP_ROUNDS}) reached; completing gracefully`);
                 yield adapter.emitCompletion({ finishReason: "length", usage });
                 return;
+            }
+
+            // #422: drive task-continuity across the compression boundary
+            // (mutating tools only — read-only queries don't replace context).
+            if (proxyResults.some((pr) => MUTATING_PROXY_TOOLS.has(pr.name))) {
+                const driverId = "acp_loop_continue_driver";
+                if (!coreMessages.some((m) => m.id === driverId)) {
+                    coreMessages.push({
+                        id: driverId,
+                        role: "user",
+                        contentType: "text",
+                        text: CONTINUE_DRIVER_TEXT,
+                    });
+                    ctx.log(`[acp-loop] round ${round}: injected continue-driver after context-mutating proxy tool (#422)`);
+                }
             }
 
             ctx.log(`[acp-loop] round ${round}: proxy tool executed; re-requesting so the model sees the result`);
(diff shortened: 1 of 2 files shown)


ranxianglei:

[bot] 🏷 Reviewed PR #422 (`2026-09-01_compress-continue-driver`) and re-ran the full pre-flight in a clean checkout. **All claims verified — ready for human merge.**

## Pre-flight (re-run here, not taken on faith)
- `npm run typecheck` — clean
- `npm test` — **893 pass, 0 fail** (includes both new tests: `loop #422a` + `loop #422b`)
- `npm run build` — success
- E2E (`tests/e2e/e2e-codex.test.ts`) — **could not run here**: `codex` binary is not on PATH (`spawn codex ENOENT`). This is an environment constraint, not a code issue. Agree with the recommendation: run the 4-phase core on a machine with `codex` + a Responses-compatible upstream before merge. (It stays gated by `ACP_TEST_E2E`, so the default `npm test` is unaffected.)

## Design claims — verified against the code
1. **Scoped to mutating tools** — `MUTATING_PROXY_TOOLS = {compress, decompress}`, `READONLY_PROXY_TOOLS = {acp_status, search_context}` (re-exported from `acp-kernel`). The guard `proxyResults.some(pr => MUTATING_PROXY_TOOLS.has(pr.name))` fires only for compress/decompress. `#422b` confirms `acp_status` does NOT carry the driver.
2. **Injection point** — the driver is pushed into `coreMessages` after the per-round hygiene block and immediately before `adapter.buildRequest(coreMessages, …)`, so it is in the re-request body.
3. **Survives `hideConsumedCompressCalls`** — that helper only filters records with `toolName === "compress" && contentType === "tool-call"` (plus their matching tool-results). The driver is `contentType: "text"` with no `toolName`, so it is untouched.
4. **Transient / never persisted** — `coreMessages` is a loop-local `const` seeded from `ctx.messages`; it is never written back to `ctx.session.state` or `ctx.messages`, so the driver is discarded at end of turn.
5. **Once per request** — deduped by `id === "acp_loop_continue_driver"`; fresh `coreMessages` per request means it re-injects only when a mutating tool fires again.
6. **Never emitted to the client** — the driver only lands in `coreMessages` (the `buildRequest` input); no `yield` path emits it.
7. **`role:"user"` + `contentType:"text"` safe across all three wires** — I fed the driver through all three converters directly:
   - Anthropic (`coreToAnthropic`): merges into the trailing user message as a second `text` block alongside the `tool_result` — valid.
   - OpenAI (`coreToOpenai`): a natural standalone `user` message — valid.
   - Responses (`coreToResponses`): a `message` item with `role:"user"` — valid.

## One minor (non-blocking) observation
The driver text hardcodes *"Context compression completed."* Two edge cases make that wording slightly imprecise, though the core instruction ("your task is NOT finished — continue, don't stop") stays correct in both:
- When the mutating tool is `decompress` (it's in `MUTATING_PROXY_TOOLS`), it's a decompression, not a compression.
- When a `compress` call **fails** (bad range / 0-char), the re-request still happens and the driver is still injected, so "completed" is a false claim — though the model already sees the failure as the tool result, and "continue" is still the right nudge.

No functional impact (the goal — don't stop at the boundary — is met in all cases), so I'd treat this as an optional polish: either soften the wording to not assert completion unconditionally, or gate the injection on a successful mutating op. Reporting only, per the minor-problem rule — no separate issue.

No `as any` / `@ts-ignore`, no `console.error` in server modules (uses `ctx.log`), `version` untouched. Good, minimal, well-scoped change.

PR #431, opened 2026-09-01
fix(#429): openai chat — write merged terminal chunk once, don't leak finish_reason

ranxianglei:
Fixes #429 — `OpenAI Chat received content after the finish reason`.

## What was wrong

Two independent defects in the openai adapter's `parseStream` (`src/loop/adapter-openai.ts`), both reachable **only** when the upstream merges `delta.content` and `finish_reason` into the **same** terminal chunk (MiniMax-M3, kimi-k3, and other providers that do this). The client's check is unconditional: any chunk carrying `content`/`refusal`/`reasoning_details`/`tool_calls` after `finish_reason` is set throws.

**Defect 1 — single round, real tool call.** The `if (finishReason)` block yielded the raw chunk as `meta` (verbatim) but did **not** `continue`, so control fell through to the text branch and the *same bytes* were written a second time. The client set `finishReason` on copy 1 and saw content on copy 2.

**Defect 2 — multi round, proxy tool.** A round that only called proxy tools (`compress`, …) is not terminal (`core.ts` `reRequest`) and emits no completion. But its merged chunk still reached the wire carrying `finish_reason`, so round 2's text landed *after* the finish reason. This is the high-frequency path (every compress on a long session with a merging provider).

## The fix

Invariant enforced: **a verbatim passthrough chunk must never carry `finish_reason`, and must be written at most once.** The sole authoritative completion is the final round's `emitCompletion` — *except* a real-tool-call round (`suppressCompletion`), where the verbatim terminal chunk **is** the completion and keeps `finish_reason`, written exactly once (the #209/#224 passthrough-consistency guarantee).

- `stripFinishReasonChunk(buf)` — blanks `finish_reason` on a passthrough chunk (safe fallbacks: non-`data:` line, parse error, or no `choices` → returned unchanged).
- `reasoning` / `text` / `meta` branches strip when the chunk carries `finish_reason`.
- `sawRealToolCall` branch keeps the chunk verbatim and adds `continue` (no double-write).

## Tests

`tests/openai-finish-reason-dup.test.ts` — 4 cases:
1. merged content+finish chunk with a real tool call is written **once**, keeping `finish_reason` (defect 1).
2. passthrough chunk from a proxy-tool round leaks **no** `finish_reason` (defect 2).
3. regression: text-only merged terminal round strips the passthrough finish but the `done` event is intact (not suppressed).
4. regression: reasoning chunk merged with `finish_reason` leaks nothing.

All 4 fail on the pre-fix baseline and pass with the fix.

[...]

PR #433, opened 2026-09-01
fix(#432): openai chat — settle a merged terminal chunk's tool_calls before the finish reason

### LINKED ISSUE #432: [bug] openai chat: tool_call in a merged terminal chunk (tool_calls + finish_reason) is silently dropped
Related to #429 (the two `finish_reason` defects fixed in #431). This is a **third**, distinct defect in the same area, reported in the #429 thread and deliberately left out of that fix.

## Symptom

When the upstream puts `delta.tool_calls` and `finish_reason` in the **same** terminal chunk, the tool call(s) in that chunk are lost. With two calls in one merged terminal chunk, only the earlier one (from a prior chunk) survives.

## Root cause

In `src/loop/adapter-openai.ts` `parseStream`, the `if (finishReason)` block runs `settleToolCalls()` **before** the current chunk's own `tool_calls` are accumulated into `pending`:

1. `if (finishReason)` → `settleToolCalls()` settles the *previously* accumulated `pending` and calls `pending.clear()`.
2. Control then falls through to the `if (delta.tool_calls)` branch, which pushes this chunk's `tool_calls` into `pending`/`rawToolChunks` — but `settleToolCalls()` has already run, so **no one settles them again**.
3. The chunk is either written verbatim (real-tool round) or the buffered calls are never replayed as structured events → the call is dropped.

## Impact

A tool call the model actually issued is silently discarded — the client never sees it and it is never executed. Rarer than the #429 defects (needs `tool_calls` in the *same* chunk as `finish_reason`, not just `content`), but it is a real functional loss.

## Suggested direction

Accumulate the current chunk's `delta.tool_calls` into `pending`/`rawToolChunks` **before** the `if (finishReason)` block's `settleToolCalls()`, so the settle sees every call in the round. Needs care around replay order (`rawToolChunks`) and the content-yield ordering, so it should land as its own change with its own tests.

## Repro sketch

```
sseChunk({ tool_calls: [{ index: 0, id: "a", function: { name: "bash", arguments: "{}" } }] })
sseChunk({ tool_calls: [{ index: 1, id: "b", function: { name: "grep", arguments: "{}" } }], }, "tool_calls")
// only call "a" survives; call "b" (in the merged terminal chunk) is dropped
```

Reported by @0ver1aw in the #429 thread.

ranxianglei:
Fixes #432.

## Root cause
In `src/loop/adapter-openai.ts` `parseStream`, the `if (finishReason)` block calls `settleToolCalls()` **before** the current chunk's own `delta.tool_calls` are accumulated into `pending`/`rawToolChunks`. So when the upstream puts `tool_calls` and `finish_reason` in the **same** terminal chunk, `settleToolCalls()` settles only the *previously* accumulated calls, then the post-settle `if (delta.tool_calls)` branch pushes this chunk's calls into `pending` — but nothing settles them again. The call is dropped from the structured events core sees; when it is the deciding (or only) real call it is never replayed to the client at all.

## Fix
Accumulate the current chunk's `delta.tool_calls` **before** the `if (finishReason)` block's `settleToolCalls()`, so the settle sees every call in the round:

1. Extracted the existing per-index accumulation into a helper `accumulateToolCalls(tcs, jsonStr, parsed, bufferRaw)`.
2. Call it before the `finishReason` block with `bufferRaw = !finishReason`:
   - non-terminal chunk → buffered into `rawToolChunks` (verbatim replay, as before);
   - terminal chunk → accumulated into `pending` but **not** re-buffered (it is forwarded once via `rawBuf` in the settle branch, so it must not be double-forwarded).
3. The post-settle `if (delta.tool_calls)` block now only yields any content merged into the same chunk; it no longer accumulates.

`hadToolCalls` is now computed after the pre-settle accumulation, so it correctly reflects the terminal chunk's own calls for the `finish_reason: "stop" → "tool_calls"` rewrite.

## Why this PR is based on the #429 fix branch (#431)
This change is in the same code region as #429 and **depends on it**: the #429 fix adds the `continue` in the `sawRealToolCall` branch. Without that `continue`, accumulating the terminal chunk's calls makes `sawRealToolCall` true, which yields the chunk via `meta(rawBuf)` and then falls through to the `delta.tool_calls` content branch — double-writing the content for a terminal chunk that carries both content and tool_calls. So this fix is only correct on top of #431. **Please merge #431 first, then rebase/merge this.**

## Tests
New `tests/openai-merged-terminal-toolcall.test.ts` (5 cases):
- single real call in a merged terminal chunk → forwarded once + a single authoritative `done` (was a double-`done`);
- two calls (prior + merged-terminal, the issue repro) → both survive in the structured events (was: only the earlier one);
- proxy tool (`compress`) in a merged terminal chunk → still server-side: no wire leak, no `finish_reason` leak, `passthrough` not set;
- name-split where the last fragment completes in the merged terminal chunk → full name in the structured event (was truncated);
- content + tool_call + finish in one chunk → written once.

The first, second, and fourth fail before this change and pass after; the other two are regression guards.

## Pre-flight
- `npm run typecheck` — clean
- `npm test` — 901 tests: 899 pass / 2 fail (see note)
- `npm run build` — success

**Test-suite note:** the 2 failures are `tests/plugin-agent.test.ts` "before_provider_headers stays silent when the manifest fetch keeps failing" and "pi extension is inert without a proxy" (`pi.tools.length` 4 vs 0). They reproduce on `origin/master` in the same checkout (892 tests: 890 pass / the same 2 fail) and in isolation, so they are pre-existing and unrelated to this change. This branch adds 5 passing tests and regresses nothing.

## Files (2)

- src/loop/adapter-openai.ts [MODIFIED] (+38 -19)
- tests/openai-merged-terminal-toolcall.test.ts [ADDED] (+165 -0)

### DIFF
--- a/src/loop/adapter-openai.ts
+++ b/src/loop/adapter-openai.ts
@@ -216,6 +216,34 @@ export function createOpenaiAdapter(requestBody: Record<string, unknown>, client
             // So: buffer EVERYTHING, decide once at finish.
             const rawToolChunks: { json: string; parsed: Record<string, unknown> }[] = [];
             let sawRealToolCall = false;
+            // bufferRaw=false for the terminal chunk: it is forwarded once via
+            // rawBuf in the settle branch, so its fragments must not also be
+            // buffered into rawToolChunks (double-forward).
+            const accumulateToolCalls = (
+                tcs: Array<Record<string, unknown>>,
+                jsonStr: string,
+                parsed: Record<string, unknown>,
+                bufferRaw: boolean,
+            ): void => {
+                if (bufferRaw) rawToolChunks.push({ json: jsonStr, parsed });
+                for (const tc of tcs) {
+                    const idx = typeof tc.index === "number" ? tc.index : 0;
+                    const fn = tc.function as Record<string, unknown> | undefined;
+                    const name = typeof fn?.name === "string" ? fn.name : "";
+                    const id = typeof tc.id === "string" ? tc.id : "";
+                    const rawArgs = fn?.arguments;
+                    const args = typeof rawArgs === "string" ? rawArgs : (rawArgs !== null && typeof rawArgs === "object" ? JSON.stringify(rawArgs) : "");
+                    let buf = pending.get(idx);
+                    if (!buf) {
+                        buf = { index: idx, id, name, arguments: args };
+                        pending.set(idx, buf);
+                    } else {
+                        if (id) buf.id = id;
+                        buf.name += name;
+                        buf.arguments += args;
+                    }
+                }
+            };
             const flushPendingAsStructured = function* (): Generator<ParsedStreamEvent> {
                 for (const [, tc] of pending) {
                     if (tc.name.length > 0 || tc.id.length > 0) {
@@ -316,6 +344,14 @@ export function createOpenaiAdapter(requestBody: Record<string, unknown>, client
                 const delta = choice.delta as Record<string, unknown> | undefined;
                 const finishReason = typeof choice.finish_reason === "string" ? choice.finish_reason : undefined;
 
+                // #431: accumulate this chunk's tool_calls BEFORE the
+                // finishReason block's settleToolCalls() so a merged terminal
+                // chunk (tool_calls + finish_reason in the same chunk) is seen
+                // by the settle and not dropped.
+                if (delta && delta.tool_calls) {
+                    accumulateToolCalls(delta.tool_calls as Array<Record<string, unknown>>, jsonStr, parsed, !finishReason);
+                }
+
                 if (finishReason) {
                     yield* flushFilter();
                     const hadToolCalls = [...pending.values()].some((tc) => tc.name.length > 0 || tc.id.length > 0);
@@ -351,25 +387,8 @@ export function createOpenaiAdapter(requestBody: Record<string, unknown>, client
                 }
 
                 if (delta.tool_calls) {
-                    const tcs = delta.tool_calls as Array<Record<string, unknown>>;
-                    rawToolChunks.push({ json: jsonStr, parsed });
-                    for (const tc of tcs) {
-                        const idx = typeof tc.index === "number" ? tc.index : 0;
-                        const fn = tc.function as Record<string, unknown> | undefined;
-                        const name = typeof fn?.name === "string" ? fn.name : "";
-                        const id = typeof tc.id === "string" ? tc.id : "";
-                        const rawArgs = fn?.arguments;
-                        const args = typeof rawArgs === "string" ? rawArgs : (rawArgs !== null && typeof rawArgs === "object" ? JSON.stringify(rawArgs) : "");
-                        let buf = pending.get(idx);
-                        if (!buf) {
-                            buf = { index: idx, id, name, arguments: args };
-                            pending.set(idx, buf);
-                        } else {
-                            if (id) buf.id = id;
-                            buf.name += name;
-                            buf.arguments += args;
-                        }
-                    }
+                    // tool_calls already accumulated above (before settle); only
+                    // yield any content merged into the same chunk now.
                     if (typeof delta.content === "string" && delta.content.length > 0) {
                         const clean = tagFilter.push(delta.content);
                         if (clean.length > 0) {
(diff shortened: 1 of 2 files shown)


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 On it — reviewing PR #433 (branch `2026-09-01_fix-merged-terminal-toolcall`). I'll verify the root-cause claim in `src/loop/adapter-openai.ts`, check the #429 dependency, run the new tests + full pre-flight (typecheck / test / build), and confirm the "fails-before / passes-after" claims. Will report back with results.

ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 **Review complete — LGTM.** I checked out the branch and independently verified every claim in the PR body. Details below.

## Verified

**Root cause (confirmed).** In `src/loop/adapter-openai.ts` `parseStream`, the `if (finishReason)` block called `settleToolCalls()` *before* the current chunk's own `delta.tool_calls` were accumulated. So a call whose first/only fragment lands in the merged terminal chunk was pushed into `pending` only *after* the settle ran — and nothing settled it again. Dropped from the structured events; when it's the deciding/only real call, never replayed to the client.

**Fix logic (traced all 5 cases).** The pre-settle `accumulateToolCalls(tcs, jsonStr, parsed, bufferRaw = !finishReason)` is correct:
- non-terminal → buffered into `rawToolChunks` (verbatim replay, unchanged from before);
- terminal → accumulated into `pending` but **not** re-buffered, since it's forwarded once via `rawBuf` in the `sawRealToolCall` settle branch (no double-forward);
- `hadToolCalls` is now computed after accumulation, so the `stop → tool_calls` rewrite is correct;
- the post-settle `if (delta.tool_calls)` block now only yields merged content (no longer accumulates).

**#429 dependency (confirmed real).** The `continue` in the `sawRealToolCall` branch (from `8b06952`) is load-bearing here. Without it, a terminal chunk carrying both content and tool_calls would be yielded via `meta(rawBuf)` and then fall through to the content branch — double-writing the content. I confirmed the `continue` is present at the settle branch and that the mixed case is covered.

**Fails-before / passes-after (confirmed by running it).** I swapped in the pre-#431 adapter (the `8b06952` state) and ran the new test file:
- ❌ test 1 (single real call, merged terminal) — fails before
- ❌ test 2 (two calls, issue repro) — fails before
- ✅ test 3 (proxy `compress`, merged terminal) — passes before (regression guard)
- ❌ test 4 (name-split completing in terminal) — fails before
- ✅ test 5 (content + tool_call + finish, written once) — passes before (regression guard)

Exactly matches "first, second, fourth fail before; the other two are regression guards." All 5 pass on the PR branch.

**Pre-flight (re-ran all three):**
- `npm run typecheck` — clean
- `npm test` — **901 pass / 0 fail** (896 baseline + 5 new)
- `npm run build` — success

**Both compression modes.** The change reorders accumulation only; the proxy-vs-real decision (`PROXY_TOOL_SET`) is untouched. Proxy mode is covered by test 3 (compress stays server-side: no wire leak, no `finish_reason` leak, `passthrough` not set). Plugin mode is unaffected — the agent's own tools aren't in `PROXY_TOOL_SET`, and the reorder doesn't change classification.

## Merge ordering (the one thing to sequence)

The #429 fix is **not on master yet** — it lives on its own branch `2026-09-01_fix-finish-reason-dup` (tip = `8b06952`). This PR's branch is exactly that branch + one commit (`bf0f91c`). So:

- **Recommended:** merge the #429 PR first, then rebase #433 onto the updated master (the redundant `8b06952` drops out cleanly), then merge #433. This is what the PR body asks for.
- **Also safe:** merge #433 directly — `8b06952` is a clean ancestor, so there's no conflict risk either way; you'd then close the standalone #429 PR as redundant (its work is fully contained here).

Either path is correct; I just want the two not to land as duplicate commits.

## Minor (non-blocking)

The #429 thread's floor-4 comment cross-references "PR #431" / "#432", but the tool-call-dropped bug is actually **#431** (fixed here by #433). Stale reference in an old comment only — the PR itself correctly says "Fixes #431". No action needed.

I won't merge (human-only). Ready for you to sequence the #429 → #433 merge.

PR #440, opened 2026-09-02
fix: acp-loop zero-side-effect truncation retry + well-formed anthropic error stream (#413)

ranxianglei:
fix: acp-loop zero-side-effect truncation retry + well-formed anthropic error stream (#413)

## Root cause (confirmed, all three)

When the upstream returns 200 and the SSE stream ends early (no `message_delta`/`message_stop`), the compress loop only discovers the truncation *after* the parse loop ends with `!sawDone` — `fetchWithRetry` inspects `response.ok` after headers and never reads the body, so mid-stream EOF is invisible to it. Consequences, each confirmed by code read and by red tests written against the unmodified code:

1. **No retry.** The `!sawDone` branch fails fast. A round that forwarded nothing (`0 text + 0 reasoning + 0 executed tools`) is a zero-side-effect retry candidate; instead the upstream flake landed in the host session (3 production events in 3 days, two of them 0+0).
2. **Orphan Anthropic stream.** The adapter's `emitError` is stateless w.r.t. what was already forwarded — it emits `content_block_start/delta/stop + message_delta + message_stop` directly. If the truncation hit before `message_start` reached the client, the client sees a block sequence with no starting frame (protocol-malformed); open `thinking` blocks are never closed. The responses branch already fixed its own copy of this class (`src/stream-error.ts`); the Anthropic adapter was missed.
3. **Double logging.** `ctx.log` (→ `log("info", "[sid] ...")`) plus a global `loggerLog("error", ...)` — both tee to the same sink, so every truncation produced two lines.

## Changes

**`src/loop/core.ts`**
- The per-round parse loop is wrapped in an inner attempt loop. On a qualifying truncation it retries **once** through the (hoisted) `fetchWithRetry` wrapper and re-parses the same round.
- Retry predicate — the zero-side-effect invariant: `!sawDone && !forwardedAny && calls.length === 0 && !(textProtocol && text accumulated) && !signal.aborted && !truncationRetried`. `forwardedAny` is a per-attempt flag set by the five parse-loop yield sites; the post-loop `tool_call`/marker/`emitText` yields are covered by the explicit `calls`/text checks. The check runs before `recordUsage` so a retried round's tokens are not double-counted.
- `roundBody` tracks the last body actually sent (updated on the re-request path before the `currentUpstream` swap), so a round-2+ retry re-sends the compressed body, not the original request.
- Truncation is logged once (via `ctx.log`); the duplicate global `loggerLog("error", ...)` line is removed.

**`src/loop/adapter-anthropic.ts`**
- The adapter closure now tracks `messageStartForwarded` and `openBlocks` (client indices of forwarded-but-unclosed blocks).
- `emitError` emits, in order: a synthetic `message_start` if none was forwarded (captured id or generated, zero usage), `content_block_stop` for every still-open block, the existing error text block, and the existing terminal. The stream is now always well-formed — starting frame first, every block closed, terminal last — regardless of where the truncation hit.

**`tests/loop-truncation.test.ts`** (new, 6 tests — all failed against the pre-fix code)
- T1: 0-event EOF + failed retry → exactly 1 fetch, schema-valid error stream, exactly 1 log line (acceptance criteria 1 + 3)
- T2: 0-event EOF + successful retry → one clean stream, no error surfaced
- T3/T4: truncation with an open thinking / text block → block closed before the error block, no retry (content already forwarded) (acceptance criterion 2)
- T5: the retry itself truncates → no second retry (one per request)
- T6: round-2 truncation after a proxy tool call → retry re-sends the round-2 body

## Pre-flight

- `npm run typecheck` — clean
- `npm test` — 874 pass / 0 fail (868 + 6 new)
- `npm run build` — success

## Out of scope (reported, not fixed)

- `adapter-responses.ts` `emitError`: a `response.failed` without a prior `response.created` is the same orphan class if the stream truncated before the created event.
- The re-request HTTP-failure path in `core.ts` also double-logs (`ctx.log` + `loggerLog("error")`) — same pattern; the issue only reported the truncation one.
- Truncation with `reRequest = true` (truncated after a complete proxy tool call, before the done event) is silently swallowed — the loop continues to the next round. Arguably benign; unchanged.

## Devlog

`devlog/2026-09-02_acp-loop-truncation-retry/` — REQ.md, WORKLOG.md, DESIGN.md.

### DIFF
--- a/devlog/2026-09-02_acp-loop-truncation-retry/DESIGN.md
+++ b/devlog/2026-09-02_acp-loop-truncation-retry/DESIGN.md
@@ -1,0 +1,109 @@
+# DESIGN: acp-loop truncation retry + well-formed Anthropic error stream
+
+## Problem shape
+
+`runCompressLoop` is an `AsyncGenerator<Buffer>`: it reads the upstream SSE
+stream, forwards client-facing chunks as it parses them, and only knows the
+stream was truncated *after* the parse loop ends without a done event
+(`!sawDone`). `fetchWithRetry` cannot see this failure — it inspects
+`response.ok` after headers and never reads the body. So recovery has to live
+at loop level, where the "did I already commit to the client?" state is known.
+
+## Retry predicate — the zero-side-effect invariant
+
+A truncated round can be transparently replayed **iff** nothing observable
+happened. All of these must hold at the end of the parse loop:
+
+- `!sawDone` — the stream actually truncated;
+- `!forwardedAny` — no chunk was yielded to the client this attempt. A
+  per-attempt flag set by `fwd()` wrapping the five yield sites in the parse
+  loop (raw text, synthetic text, raw reasoning, synthetic reasoning, meta).
+  Ping-only streams count as forwarded (conservative: the client saw
+  something, so a retry would duplicate it);
+- `calls.length === 0` — no tool call was collected. The `tool_call` /
+  marker / `emitText` yields happen *after* the parse loop, so they are not
+  yet covered by `forwardedAny` and need explicit checks:
+  - `calls.length === 0` rules out tool_call and marker yields (both are
+    derived from `calls`);
+  - `!(ctx.textProtocol && assistantText.length > 0)` rules out the
+    text-protocol `emitText` yield (only emitted when textProtocol and the
+    round produced text);
+- `!signal?.aborted` — the client is still connected;
+- `!truncationRetried` — function-level flag: at most ONE retry per request
+  (the issue says "走一次"). The flag is never reset, so a second truncated
+  round (round 2+) also fails fast.
+
+Placement: the check runs **after** the parse loop and **before**
+`recordUsage`. This matters: a round-2+ `message_start` sets `usage` without
+forwarding anything, so recording before the retry check would double-count
+the retried attempt's tokens.
+
+Mechanics: the parse loop is wrapped in an inner `for (;;)` attempt loop.
+On a qualifying truncation the loop calls `fetchUpstream(roundBody)` (the
+hoisted `fetchWithRetry` wrapper — same backoff, same `BILI_REPLAY_RETRY_*`
+env knobs, same #189 rejection logging) and `continue`s; the outer round
+number is unchanged, so a round-1 retry re-parses as round 1 and its
+`message_start` is forwarded exactly once (attempt 1 forwarded nothing).
+Per-attempt state (text/reasoning/calls/usage/finishReason/sawDone/
+suppressCompletion/forwardedAny) is reset at the top of the attempt loop.
+
+`roundBody` (new): the re-request path replaces the body each round
+(`compressMessages`); the old code re-fetched with `requestOptions` + a local
+`newBody`. The retry must re-send **the body that produced the truncated
+stream** — on round 1 that is the original `requestBody`, on round N it is
+the compressed body. `roundBody` is updated right before the
+`currentUpstream` swap, so it always holds the last body actually sent.
+
+Failure of the retry itself falls through to the existing truncation error
+path (empty body on the retry response is converted to
+`UpstreamHttpError` so it is caught by the same handler).
+
+## Adapter: tracking what was forwarded
+
+The Anthropic adapter closure gains two pieces of state:
+
+- `messageStartForwarded: boolean` — set where the round-1 `message_start`
+  is forwarded (always forwarded in round 1, so the flag is set there);
+- `openBlocks: number[]` — client indices of blocks whose
+  `content_block_start` was forwarded but whose `content_block_stop` was not.
+  Pushed in the non-tool `content_block_start` branch (forwarded in all
+  rounds); removed in the `content_block_stop` branches via an `indexOf`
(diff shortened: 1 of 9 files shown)


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 Reviewed PR #413 (acp-loop zero-side-effect truncation retry + well-formed anthropic error stream). I checked out the branch, re-ran the full pre-flight, and stress-tested the retry invariant. Verdict: **solid fix, but one real hole in the zero-side-effect invariant for the OpenAI adapter (round 2+)** — trivial to close (2 lines). Details below.

## Pre-flight — re-verified, all green
- `npm run typecheck` — clean
- `npm test` — **874 pass / 0 fail** (868 + 6 new)
- `npm run build` — success
- **Tests are not vacuous:** ran the 6 new `tests/loop-truncation.test.ts` against the merge-base (pre-fix) code → **0/6 pass**; against the fix → **6/6 pass**. Good red→green.
- Branch is 19 commits behind `master`, but I dry-ran a merge: **no conflicts** on `src/loop/core.ts` / `src/loop/adapter-anthropic.ts` (master hasn't touched them since the merge-base). Clean to merge as-is.

## What's correct
- **Retry predicate & placement** (`src/loop/core.ts`): check runs after the parse loop, before `recordUsage`, so a retried round's tokens aren't double-counted. Per-attempt state reset is complete. `roundBody` correctly tracks the last-sent body (round-1 original, round-N compressed), so a round-2+ retry re-sends the right body — T6 confirms.
- **`emitError` well-formedness** (`src/loop/adapter-anthropic.ts`): synthetic `message_start` only if none forwarded, `content_block_stop` for every still-open block, then error block + terminal. Correct for all four truncation positions (pre-start / mid-thinking / mid-text / post-content). `openBlocks` push/pop is right (tool_use never pushed — buffered in `pending`).
- **Double-logging fix**: the truncation path now logs once via `ctx.log`; the global `loggerLog("error")` duplicate is removed. The re-request HTTP-error path's double-log is correctly left out of scope (documented).
- **Both compression modes**: the retry is protocol-neutral (0-event truncation is safe to replay in plugin and proxy mode alike) and the `emitError` fix applies to both. No mode-specific breakage.

## Finding — the invariant has a hole for OpenAI round-2+ reasoning
The PR's core claim is "retry **iff** nothing was forwarded." DESIGN.md says `forwardedAny` is "set by `fwd()` wrapping the **five** yield sites ... (raw text, synthetic text, raw reasoning, synthetic reasoning, meta)." But only **3 of the 5** are actually wrapped — the two **reasoning** sites are not:

- `src/loop/core.ts:279` → `yield ev.raw;` (reasoning raw) — **not** `fwd()`
- `src/loop/core.ts:281` → `yield adapter.emitReasoning(ev.delta);` — **not** `fwd()`

For **Anthropic** this is harmless: a thinking block is always preceded by its `content_block_start`, which *is* forwarded via `fwd()` (round 1 by `round===1`, round 2+ by `!firstRoundOnly`), so `forwardedAny` is already true before any reasoning delta.

For the **OpenAI adapter** it is **not** harmless. In round 2+ the leading role chunk is emitted as `meta` with `firstRoundOnly: true` → **not forwarded** (adapter-openai.ts:368), and `reasoning_content` is forwarded via `raw` (adapter-openai.ts:328) → hits the unwrapped `yield ev.raw`. So a round-2 response that starts with reasoning and then truncates (no `finish_reason`/`[DONE]`) leaves `forwardedAny === false` even though bytes already reached the client.

**Repro (I ran this against the branch):** round 1 = `acp_status` proxy tool call (→ re-request), round 2 = role chunk + `reasoning_content:"let me think"` + EOF. Result on current code:
- `fetch` calls = **2** (the re-request **plus a spurious truncation retry**)
- `"let me think"` appears **2×** in the client stream — the already-forwarded reasoning is **duplicated** to the client

This is a regression vs pre-fix (which had no retry at all) and directly contradicts the "zero-side-effect" invariant the PR is built on. The Responses adapter is unaffected (it emits no `reasoning` events, only `text`/`meta`).

**Fix (verified):** wrap both reasoning yields in `fwd()`:
```ts
if (ev.raw) {
    yield fwd(ev.raw);
} else if (round > 1 && ev.delta.length > 0 && adapter.emitReasoning) {
    yield fwd(adapter.emitReasoning(ev.delta));
}
```
With that change my repro gives `fetch` calls = **1**, reasoning appears **1×**, and the full suite stays green (874/0). It also makes the code match DESIGN.md's "five yield sites" claim.

## Recommendation
The primary production case (Anthropic 0-event truncation) is handled correctly and the fix is otherwise clean — I'd be comfortable merging. But since the whole PR is *about* the zero-side-effect invariant, I'd recommend closing the 2-line OpenAI-reasoning gap before merge (it currently lets a side-effecting retry through, and DESIGN.md already asserts it's covered). Happy to add a regression test for the OpenAI round-2+ reasoning-truncation case alongside the fix if you want it.

Out-of-scope items (responses `emitError` orphan, re-request double-log, `reRequest=true` silent swallow) are correctly scoped and documented — no action needed here.

PR #448, opened 2026-09-02
fix: bili pi launcher disables pi's native auto-compaction (#447)

ranxianglei:
## Problem (#447)

`bili pi` never disabled pi's own context compaction, so pi's native summarizer (`shouldCompact`, `enabled: true` by default) could fire ahead of / alongside bili's ACP compression and lossily rewrite the context bili is managing — losing the detailed history.

## Fix

The pi overlay (`<piHome>-bili`) is now **always** built when the real pi home exists, and always carries a `settings.json` with `compaction.enabled=false` (merged over the user's real `settings.json`, which is never mutated). This makes bili's ACP compression the sole compression path under `bili pi`, mirroring how codex/claude already get their native auto-compact aligned (codex `model_auto_compact_token_limit`, claude `CLAUDE_CODE_AUTO_COMPACT_WINDOW`).

Changes:
- `refreshOverlayHome` / `mergeOverlayEntry` now accept a **set** of generated files (`models.json` + `settings.json`) so both stay out of the real-home symlink and merge-back sets (the #410 guarantee now holds for both).
- `preparePiHttpRewrite` returns `undefined` only for a missing real home; `models.json` is still rewritten for HTTP/wrapped-HTTPS upstreams when present, and copied verbatim otherwise.
- New `writePiCompactionDisabledSettings` merges the real `settings.json` with `compaction.enabled=false` (other keys — `packages`, `theme`, `reserveTokens`, … — preserved).

## Verification

- `npm run typecheck` ✅
- `npm test` — 899 pass, 0 fail ✅
- `npm run build` ✅
- Integration check: realistic pi home (HTTP vllm provider + user `settings.json`) → overlay `settings.json` = `{packages, theme, compaction:{reserveTokens:16384, enabled:false}}`, `models.json` baseUrl rewritten to `/bili/`, real `settings.json` untouched ✅

Launcher-only change (no request-pipeline change), so the codex E2E suite is not exercised by this diff.

### DIFF
--- a/src/launcher.ts
+++ b/src/launcher.ts
@@ -782,7 +782,10 @@ function mergeSqliteSet(overlay: string, realHome: string, base: string): boolea
     }
 }
 
-function refreshOverlayHome(realHome: string, overlay: string, generatedFile: string): boolean {
+function refreshOverlayHome(realHome: string, overlay: string, generatedFile: string | string[]): boolean {
+    const generatedFiles = new Set(Array.isArray(generatedFile) ? generatedFile : [generatedFile]);
+    const isGeneratedDraft = (name: string): boolean =>
+        [...generatedFiles].some((g) => name.startsWith(`.${g}.`) && name.endsWith(".tmp"));
     try {
         fs.mkdirSync(overlay, { recursive: true });
     } catch {
@@ -814,7 +817,7 @@ function refreshOverlayHome(realHome: string, overlay: string, generatedFile: st
         // re-points the db; any other set moves wholesale.
         const dbSets: { base: string; keepSidecars: boolean }[] = [];
         for (const entry of overlayEntries) {
-            if (!entry.endsWith(".db") || entry === generatedFile) continue;
+            if (!entry.endsWith(".db") || generatedFiles.has(entry)) continue;
             const members = sqliteSetMembers(entry);
             if (!members.some((m) => m !== entry && overlayEntries.includes(m))) continue;
             let mainSt: fs.Stats | undefined;
@@ -832,9 +835,9 @@ function refreshOverlayHome(realHome: string, overlay: string, generatedFile: st
             }
         }
         for (const entry of overlayEntries) {
-            if (entry === generatedFile) continue;
+            if (generatedFiles.has(entry)) continue;
             const overlayPath = path.join(overlay, entry);
-            if (entry.startsWith(`.${generatedFile}.`) && entry.endsWith(".tmp")) {
+            if (isGeneratedDraft(entry)) {
                 try {
                     fs.unlinkSync(overlayPath);
                 } catch {}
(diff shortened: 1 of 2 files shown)


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 Reviewed PR #448 (fixes #447). Verified and analyzed in detail — assessment below.

**Verification (all green):**
- `npm run typecheck` ✅
- `npm test` — 899 pass, 0 fail ✅
- `npm run build` ✅

**The root cause is real and the fix targets it correctly.** For pi, raw-HTTPS providers (the common case) are classified into `httpsDomains` (cert-MITM), *not* into `httpRewrites`/`httpsRewrites` (`discoverRoutes`, `src/launcher.ts:267-270`). So for a typical HTTPS-only pi setup, `preparePiHttpRewrite` returned `undefined`, the `<piHome>-bili` overlay was never built, `PI_CODING_AGENT_DIR` was never set, and pi ran from its real home with native compaction left at its default (`enabled: true`). That's exactly the gap #447 describes. Making the overlay always build (when the real home exists) closes it. I reproduced both scenarios manually:
- HTTPS-only providers (no rewrites) → overlay now built, `settings.json = {"theme":"dark","compaction":{"enabled":false}}`.
- Realistic home (HTTP vllm provider + user `settings.json`) → `settings.json = {packages, theme, compaction:{reserveTokens:16384, enabled:false}}`, vllm baseUrl → `/bili/…`, openai (HTTPS) untouched, real `settings.json` byte-for-byte unchanged, `auth.json` symlinked.

**The #410 guarantee now holds for both generated files.** `refreshOverlayHome`/`mergeOverlayEntry` take a *set* (`settings.json` + `models.json`), so both stay out of the real-home symlink set and the merge-back set at any depth. `isGeneratedDraft` correctly matches `.settings.json.<pid>.tmp` / `.models.json.<pid>.tmp`. Clean generalization of the prior single-file logic; the omp/hermes/dsh call sites still pass a single string, which the `Array.isArray` guard handles.

**`writePiCompactionDisabledSettings` is correct.** Reads the real `settings.json`, shallow-copies the `compaction` object, sets `enabled=false`, writes to the overlay — preserving `packages`/`theme`/`reserveTokens`/etc. Real file is never mutated. Missing/malformed/non-object `settings.json` and non-object `compaction` are all handled gracefully (falls back to `{}` then forces `enabled=false`).

**Consistent with codex/claude, with an intentional mechanism difference.** Same goal (ACP is the sole compression path). codex's local compaction is benign (same-session truncation the kernel deactivates by message id), so codex gets threshold *alignment* (`model_auto_compact_token_limit`). pi's native summarizer is lossy — it rewrites the context bili manages — so pi gets full *disable*. That's the right call for pi.

**Two small notes (non-blocking):**
1. The description says `models.json` is "copied verbatim otherwise," but the code re-serializes it via `JSON.stringify(modelsRoot)` whenever it's a present, valid object — even when no rewrite matched. For plain-JSON `models.json` the parsed content is identical (only whitespace/formatting differs), so it's harmless in practice. Either do a true byte-copy in the no-rewrite case or soften the description.
2. The whole fix hinges on pi reading `compaction.enabled` (boolean) from `settings.json` to gate its native summarizer. That's pi-internal (separate repo) and isn't referenced anywhere in this codebase, so I'm taking the author's word for it — worth a one-line pointer to the pi source that confirms the exact key/shape before merge.

**Bottom line: LGTM to merge.** Correct fix, well-tested (the new/updated `preparePiHttpRewrite` tests cover the no-rewrite overlay build, the non-existent-home `undefined`, compaction disable, and key preservation), clean verification. Launcher-only, so no codex E2E needed. I'd only want the `compaction.enabled` key-shape confirmation (note 2) before merge; note 1 is optional polish.

PR #452, opened 2026-09-02
fix: strip plugin-mode render-tag echoes on chat-completions/anthropic passthrough (#206 parity)

ranxianglei:
## What

Plugin-mode passthrough now strips model-emitted ACP render-tag echoes on the **OpenAI chat-completions** and **Anthropic** wires — the same protection the Responses wire already had since v0.1.56 (`7e95b48`, `pipePluginResponsesWithStrip`).

## Why (issue #14 / dog#14 "bili总是循环")

In plugin mode, `server.ts` routed streams by wire: Responses → strip pipe; **everything else → `pipeThroughWithUsage`, i.e. verbatim**. Non-stream JSON also only stripped for Responses. A client on chat-completions (pi) or /v1/messages (claude-code via MCP register) therefore had model-imitated `<acp …>` tags land in the agent's replayed history, where they amplify: model echoes tags → agent stores & replays them → model imitates more. Observed with pi + qwen3.8-27b: one assistant message carried **77 echoed tags + ~3,300 blank lines**; the tags render invisibly in the pi UI, so the user just sees endless blank output ("输出循环").

## How

- `src/loop/tag-echo-filter.ts`: add `stripOpenaiChatText` (choices[].delta/message .{content,reasoning_content,reasoning}) and `stripAnthropicText` (delta.{text,thinking}, content[].{text,thinking}).
- `src/plugin.ts`: new `pipePluginChatWithStrip` — SSE passthrough, byte-identical except managed text fields, which stream through the tag-echo state machine (one filter per field × choice/block index; held tails flushed before [DONE]/message_delta/finish/non-text events as synthetic deltas; echo-only chunks dropped). Replaces `pipeThroughWithUsage` (now unused; usage sniffing folded in).
- `src/plugin.ts` `pipePluginJson`: non-stream strip extended from responses-only to all three wires.
- `src/server.ts`: plugin-mode routing sends chat/anthropic streams through the new pipe.

Tool calls stay untouched (the agent's native loop must see them verbatim — same contract as the Responses pipe).

## Verification

- `npm run typecheck` ✅
- `npm test` — 910/910 ✅ (new `tests/plugin-passthrough-tag-strip-chat.test.ts`, 11 cases: openai delta strip, split-tag swallow, reasoning_content, echo-only chunk drop, tool_calls byte-identity, tail-at-end, anthropic text/thinking deltas + split + flush order, non-stream openai/anthropic JSON strip, tag-free bodies byte-identical; updated the old test that had pinned the verbatim openai behavior)
- `npm run build` ✅

Follow-up (separate issue): plugin mode has no hard backstop when the agent ignores the compress nudge — the session in #14 grew past the window and 400'd (input 131745 + max_completion 131072 > 262144) before pi's native compaction kicked in.


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

**Review + verification of the PR** (branch `2026-09-02_plugin-chat-tag-echo-strip`, `bc57a09`)

**Verification — re-ran locally on the PR head:**
- `npm run typecheck` ✅
- `npm test` — 910/910 ✅ (matches the claim; includes the 11 new cases in `tests/plugin-passthrough-tag-strip-chat.test.ts`)
- `npm run build` ✅
- No leftover references to the removed `pipeThroughWithUsage`; the routing change sits entirely inside the `prepared?.pluginMode` branch of `src/server.ts`, so proxy mode is untouched (both-modes check per dev spec).

**Triage verdict:**
- Problem confirmed on master: the plugin branch sent every non-Responses stream to `pipeThroughWithUsage` (verbatim) and `pipePluginJson` stripped only Responses — so pi (chat-completions) and claude-code-via-MCP (`/v1/messages`) received model-emitted render tags straight into the replayed history, where they amplify exactly as in #14/#206. Right layer: strip at the wire source in the proxy (parity with the battle-tested `pipePluginResponsesWithStrip`), not client-side cleanup.
- Contracts checked: `tool_calls` / `input_json_delta` deltas pass byte-identical (agent's native loop intact); usage sniffing preserved (`lastInputTokens` still feeds the nudge decision); held tails flush as synthetic deltas *before* `[DONE]`/finish/non-delta events (order-correct on both wires, incl. Anthropic `content_block_stop`); EOF flush passes legitimate prose tails through and drops truncated tag fragments; tag-free bodies stay byte-identical (tested).

**One minor finding (report-only):** mixed-field chunk drop in `processOpenai` (`src/plugin.ts:620` + `:632`). `droppedText` latches true when *any* managed field strips to empty, and `hadTextOtherThanTextFields()` inspects key *names*, not values — so a single chunk carrying `content: "real answer"` **plus** `reasoning_content: <pure tag echo>` is dropped wholesale, losing the real answer. Reproduced empirically with a temp probe (sibling field lost; the echo-only control case drops cleanly as designed). Trigger is narrow — needs two text-family fields in one chunk (some reasoning-model providers emit this at the reasoning→content transition) *and* one of them being a pure echo — but that's precisely the heavy-echo regime #14 lived in. Suggested fix: drop the chunk only when *all* rebuilt text-field values are empty (keep the existing non-text-payload check).

Recommendation: worth fixing pre-merge — it's a mechanical ~3-line change plus one regression test, and it closes a genuine (if rare) output-loss path in the very failure mode this PR targets. Say the word and I'll push the fix + test to the branch. It's not a blocker either way: probability is low and the core protection stands regardless.

**Pending follow-up:** the hard-backstop gap (agent ignored the compress nudge → session grew past the window and 400'd: input 131745 + max_completion 131072 > 262144, before pi's native compaction kicked in) — I searched the tracker and no separate issue exists for it yet. It needs its own issue per the PR note; happy to draft it from the #14 numbers if you want.

PR #465, opened 2026-09-02
fix(launcher): isolate launches on a random free port; drop proxy attach (#446)

ranxianglei:
Closing as superseded by #526 (plan 1: port isolation only).

This PR over-interpreted 「启动器模式都是隔离」 as process isolation — dropping the #417 attach-or-spawn share path and adding per-token handshake files. Per the issue thread, the intent was **port isolation only**: launcher proxies shouldn't squat on 8787, but a healthy compatible instance (including an explicitly-started `bili start`) should still be attached, keeping the single-writer property #417 was built for.

The crash-root-cause narrative here also predates part of #417's guards; the residual owner-exit cascade was accepted in the issue thread. Thanks for the groundwork — the ephemeral-port portion lives on in #526.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -103,7 +103,7 @@ Launcher (bili pi / bili codex / bili claude / bili omp / bili opencode / bili h
 Options (override config file / env):
   -F <url>                         upstream proxy to forward through (gost-style;
                                    http://host:port; must precede the client name)
-  --port <N>                       listen port (default 8787)
+   --port <N>                       listen port (bili start: 8787; launcher: random free port)
   --host <ADDR>                    listen host (default 127.0.0.1)
   --mitm-domain <domain>           extra MITM domain (repeatable; launcher only)
   --config <FILE>                  path to config JSON (default: XDG location)
--- a/src/instance.ts
+++ b/src/instance.ts
@@ -70,6 +70,14 @@ export function instanceFilePath(): string {
     return path.join(stateDir(), "proxy-origin");
 }
 
+/** Per-launch handshake report (#446): keyed by launchToken so concurrent
+ *  launcher-spawned proxies don't clobber each other's single proxy-origin
+ *  record. The launcher reads only its own child's report (matched by
+ *  construction — it generated the token), so N isolated launchers coexist. */
+export function launchTokenFilePath(token: string): string {
+    return path.join(stateDir(), `proxy-origin-${token}`);
+}
+
 /** tmp+fsync+rename (same shape as web/api.ts atomicWriteConfig) — a torn
  *  write must never leave a half-origin behind (#406 family). */
 export function atomicWriteInstanceFile(info: ProxyInstanceFile, file?: string): void {
--- a/src/launcher.ts
+++ b/src/launcher.ts
@@ -33,7 +33,7 @@ import path from "node:path";
 import { pathToFileURL } from "node:url";
 import { spawn, type StdioOptions } from "node:child_process";
 import { DEFAULT_MITM_DOMAINS } from "./mitm.js";
-import { isProxyInstanceFile, isPidAlive, readProxyInstanceFile, type ProxyInstanceFile } from "./instance.js";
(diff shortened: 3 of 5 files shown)


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 **Review of #465 — LGTM, no blockers.** Two minor notes at the end.

**Root cause & approach — verified correct.** The old attach-or-spawn path decided share-vs-spawn by reading the shared `proxy-origin` instance file, not the requested port — so a second launcher attached to the first one's proxy, and when the first client exited its `stopProxy()` killed the shared proxy and took down every attached launcher. Changing only the default port would indeed not have fixed this; removing attach entirely matches the decision (launcher mode = always isolated; sharing only via a separately-started server).

**Verified on `c0a3c25`:**
- **Isolation**: `ensureProxyRunning` always spawns (`src/launcher.ts:1560`); `pickEphemeralPort()` (bind :0) is the launcher default, explicit `--port` still honored, `bili start` unchanged (8787 default lives in `src/config.ts:282`, untouched). `stopProxy` now always kills its own child (win32 #414 parent-gone-watcher path untouched).
- **Concurrency**: per-`launchToken` report file `proxy-origin-<token>` (`src/instance.ts:77`) is written by the server only when `BILI_LAUNCH_TOKEN` is set (`src/server.ts:396`), read only by the launcher that generated the uuid (`src/launcher.ts:1575`), matched on `inst.launchToken === launchToken` (`src/launcher.ts:1615`), unlinked on graceful shutdown (`src/server.ts:471-475`). "Read my own child's real port" is correct by construction — no cross-process matching possible.
- **#407 retained**: self-bind EADDRINUSE retry (17 attempts, then ephemeral) stays, which also covers the pick→close→rebind TOCTOU window between `pickEphemeralPort` and the child's bind.
- **Shared-file consumers intact**: the shared `proxy-origin` write is kept and still needed by MCP shell discovery fallback (`src/mcp.ts:40`), `bili plugin install` (`src/plugin-install.ts:29`), and the config-repair liveness check (`liveProxyPorts`, `src/launcher.ts:1107`). Launcher-spawned MCP shells get an explicit `BILI_MCP_PROXY` env (`src/launcher.ts:563`, `:598`), so concurrent isolated launches can't cross-wire each other's shells — the shared file degrades cleanly to a best-effort "latest proxy" hint.
- **Dead code removed cleanly**: `LAUNCHER_DEFAULT_PORT`, `attached`, `fetchHealthInfo`, `probeExistingInstance`, `instanceCompatible` — grep confirms zero remaining references repo-wide.
- **Side benefit**: the tmp log `bili-proxy-<port>.log` is now unique per launch (it name-collided before when multiple launchers shared 8787).
- **Tests**: new #446 "always spawns, never attaches" test + `pickEphemeralPort` smoke test; the #407 handshake test (child reports a different real port → handle follows it) is retained.

**Pre-flights I ran:**
- `npm run typecheck` — clean; `npm run build` — OK
- `npm test` — 913/914. The single failure (`resolveClientCommand: codex/claude resolve to themselves`) also fails on master in this environment (verified via worktree) → the "pre-existing environment-only" claim is accurate.
- Full E2E could not run here: the default upstream `http://127.0.0.1:8199/v1` refuses connections in this sandbox (warmup dies ECONNREFUSED→502 before any PR-changed code runs). Note the suite doesn't set `BILI_LAUNCH_TOKEN` anyway, so the changed handshake path isn't exercised by it; the `server.ts` delta is instance-file bookkeeping guarded by the token. Worth one standard 4-phase E2E run against a local upstream before merge, per AGENTS.md §3.

1. Cosmetic: the `--port <N>` help line in `src/cli.ts` is indented 3 spaces while all sibling options use 2.
2. Orphaned `proxy-origin-<token>` files linger in the state dir if the proxy dies without graceful shutdown (crash/SIGKILL — `finishShutdown` is the only unlink site). No correctness risk (only the owning launcher ever reads a given token file), but they accumulate ~200B each with no GC. Optional follow-up: prune stale `proxy-origin-*` entries at startup.

Happy to push the one-line indent fix onto this branch if you want it included.

PR #459, opened 2026-09-02
fix(#457): strip model-emitted render tags from plugin-mode openai/anthropic wires

ranxianglei:
## What

Closes #457. Plugin mode stripped model-emitted ACP render tags on the responses wire only (`pipePluginResponsesWithStrip`, #206); the openai/anthropic plugin paths passed model output through verbatim, so tags echoed in prose leaked into the client TUI and were replayed from client history next turn (amplification loop).

## Changes

- `src/plugin.ts` — `pipeThroughWithUsage` (openai/anthropic SSE): event-level processing now routes prose text deltas — openai `choices[].delta.content`, anthropic `text_delta` — through the same `createTagEchoFilter` state machine as the responses path, with the held tail flushed as one synthetic text delta before terminal events (openai `finish_reason`/`[DONE]`, anthropic `message_delta`/`message_stop`) and at stream end. Structural events (tool_calls, tool_use `input_json_delta`, usage frames) and tag-free streams stay byte-identical (fast path + verbatim fallback). Usage sniffing unchanged.
- `src/plugin.ts` — `pipePluginJson`: the #206 strip branch generalized to all three protocols; openai/anthropic bodies go through the new strippers.
- `src/loop/tag-echo-filter.ts` — new `stripOpenAIText` / `stripAnthropicText` (mutate-in-place, same convention as `stripResponsesText`); structural fields never touched.
- `src/server.ts` — updated the stale "VERBATIM" contract comment for the plugin passthrough branch.
- `tests/plugin-passthrough-tag-strip.test.ts` — the old test asserting a tagged openai JSON body stays byte-identical codified the bug; inverted. Added 8 tests: openai/anthropic SSE strip (single event, split across deltas with terminal flush ordering, tool_calls/input_json_delta untouched, tag-free byte-identity), openai/anthropic JSON strip (prose stripped, tool_calls/tool_use untouched, usage still sampled).

## Deliberate scope

Only `delta.content` / `text_delta` are filtered — `reasoning_content` (deepseek/qwen) and `thinking_delta` are not covered (matches issue scope; flagged in-thread as residual). When a delta becomes empty after stripping, an empty-content event is emitted rather than dropped (safer than the responses implementation's drop for role-chunk framing).

## Pre-flight

- `npm run typecheck` — clean
- `npm test` — 911/911 pass (tag-strip file: 18/18)
- `npm run build` — success
[...]

PR #466, opened 2026-09-02
fix(#460): strip render-tag echoes from non-injected proxy-mode chat SSE

ranxianglei:
## What

Issue #460 (ework #459), as rescoped by @ranxianglei: requests that skip compress injection — OpenAI title-gen exclusion (`max_tokens <= 200`), `ACP_NO_INJECT_TOOL` sessions, #353 stop_sequences classifier bypass — piped upstream SSE straight through to the client with no render-tag handling. A model echoing history-borne ACP render tags into prose leaks them to the client TUI, where they re-enter the conversation via replay and amplify (same mechanism as #457/#459).

This routes those non-injected streaming responses through the same tag-strip pipes plugin mode already uses (#452): `pipePluginChatWithStrip` (openai/anthropic) and `pipePluginResponsesWithStrip` (responses). Byte-identical for tag-free streams.

## Why no session is passed

The new call sites pass `undefined` for session on purpose: `pipePluginChatWithStrip` accumulates usage samples into `session.stats.lastInputTokens` (last-wins merge). A title-gen call's tiny `input_tokens` would clobber the main conversation's value and break compression triggering. Usage accounting is now guarded by session presence in both pipes.

## Changes

- `src/plugin.ts`: `pipePluginChatWithStrip` signature reordered to `(stream, res, protocol, session?, log?)`, session optional; usage sample applied only when a session is present. Same optionality for `pipePluginResponsesWithStrip`.
- `src/server.ts` `!useRewriter` branch: streaming requests whose upstream response is `text/event-stream` go through the strip pipes; everything else (non-SSE bodies, non-streaming) stays raw `pipeThrough` exactly as before. The content-type gate also protects against the strip pipes' residual-buffer behavior on malformed/undelimited bodies.
- Also replaces a stale comment that referenced `rewriteSseStream` "below" (removed long ago) — the misleading comment that contributed to the original mis-triage of this issue.
- `tests/proxy-noinject-tag-strip.test.ts` (new): full-server tests — anthropic `injectTool:false` stream with tags split across deltas → stripped, prose intact; tag-free stream → byte-identical; openai title-gen stream → stripped, `[DONE]` + `finish_reason` intact.
- `tests/plugin-passthrough-tag-strip-chat.test.ts`: call sites updated for the reordered signature (behavior unchanged).

## Residuals (reported in issue thread, out of scope here)

- Non-streaming JSON responses on the same non-injected paths are still not stripped (owner's scope decision).
- Dead legacy rewriters `rewriteSseStream` / `rewriteOpenaiSseStream` still exported with zero call sites — removal awaiting owner sign-off.

## Pre-flight

- typecheck: clean
- npm test: 917/918 pass; the 1 failure (`launcher.test.ts` resolveClientCommand codex path) fails identically on pristine master in this environment (codex installed at /usr/bin/codex) — pre-existing, unrelated
- tsup build: OK
- E2E codex suite not run locally (requires upstream at http://127.0.0.1:8199, unavailable here); change is additive on a branch covered by the new unit tests, plugin-mode call sites unchanged behaviorally

### DIFF
--- a/src/plugin.ts
+++ b/src/plugin.ts
@@ -536,12 +536,18 @@ function mergeUsageSample(acc: UsageSample, sample: UsageSample): void {
  *  model prose through the tag-echo state machine — #206 parity with
  *  pipePluginResponsesWithStrip. The verbatim variant let a model-emitted
  *  render tag echo land in the agent's replayed history and amplify into the
- *  "endless blank output" loop observed with pi + qwen (issue #14). */
+ *  "endless blank output" loop observed with pi + qwen (issue #14).
+ *
+ *  Also serves proxy-mode chat SSE that skipped compress injection (#460:
+ *  title-gen exclusion / ACP_NO_INJECT_TOOL / classifier bypass). Pass no
+ *  session there — usage accounting must be skipped or a title-gen call's
+ *  tiny input_tokens would clobber lastInputTokens and break compression
+ *  triggering for the main conversation. */
 export async function pipePluginChatWithStrip(
     stream: ReadableStream<Uint8Array>,
     res: import("node:http").ServerResponse,
-    session: Session,
     protocol: WireProtocol,
+    session?: Session,
     log?: (msg: string) => void,
 ): Promise<void> {
     const reader = stream.getReader();
@@ -691,7 +697,7 @@ export async function pipePluginChatWithStrip(
         }
         const rest = flushTails();
         if (rest.length > 0 && !res.destroyed && !res.writableEnded) await write(rest);
-        if (acc.inputTokens !== undefined || acc.outputTokens !== undefined || acc.cachedTokens !== undefined) {
+        if (session && (acc.inputTokens !== undefined || acc.outputTokens !== undefined || acc.cachedTokens !== undefined)) {
             applyUsageSample(session, acc, protocol);
             markDirty(session);
         }
@@ -727,11 +733,14 @@ function hadTextOtherThanTextFields(choices: unknown): boolean {
  * tag-echo state machine the compress loop uses. A held tail is flushed as a
  * final delta before the first done/completed event so the client's assembled
  * text never loses content.
- */
+ *
+ *  Also serves proxy-mode Responses SSE that skipped compress injection
+ *  (#460, e.g. ACP_NO_INJECT_TOOL). Pass no session there — see
+ *  pipePluginChatWithStrip for why usage accounting must be skipped. */
 export async function pipePluginResponsesWithStrip(
     stream: ReadableStream<Uint8Array>,
     res: import("node:http").ServerResponse,
-    session: Session,
+    session?: Session,
     log?: (msg: string) => void,
 ): Promise<void> {
     const reader = stream.getReader();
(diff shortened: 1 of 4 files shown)


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 **Review: verified, LGTM** (merge is yours — I don't merge PRs)

**Triage — bug verified, layering is right.** On master the `!useRewriter` branch is a bare `pipeThrough(upstream.body, res)` for every non-injected request — zero render-tag handling. The non-injection paths are exactly as described: title-gen `max_tokens <= 200` (`src/server.ts:1516`), the #353 classifier early-return (`src/server.ts:1406-1408`), and `ACP_NO_INJECT_TOOL` sessions. So a model echoing history-borne ACP tags into prose on those paths does leak to the client TUI verbatim, with the same replay-amplification mechanism as #457/#459. Root cause = the tag-strip step was only wired into plugin mode (#452) and the injected rewriter path; this fix closes the gap by routing the affected streams through the same pipes rather than adding new strip logic. Scope call (SSE-streaming only, non-streaming JSON left as residual) matches your rescoping of #460.

**Design points checked out:**
- `session?: undefined` at the new call sites is correct and necessary: `applyUsageSample` sets `lastInputTokens = Math.max(0, total - compressCreditTokens)` (`src/plugin.ts:516`), so passing the session for a title-gen call would clobber the main conversation's value and skew compression triggering. With `undefined`, usage accounting on these paths stays exactly as master (master never sniffed usage here either).
- Branch placement is sound: plugin-mode branch (earlier) untouched; `prepared === null` falls through to the raw `pipeThrough`; the native-compact `resetAfterSuccess` path runs before the new branch; the content-type gate keeps undelimited non-SSE bodies away from the pipes' residual-buffer behavior.
- Byte-identity claim holds for LF-terminated tag-free streams: events are emitted verbatim per event boundary, and `normalizeSseLineEndings` is a literal no-op when the buffer contains no CR. One nuance: a CRLF-terminated upstream would be normalized to LF — pre-existing pipe behavior (plugin mode already does this today), not a regression.

**Pre-flight (ran on cf959cb):**
- typecheck: clean
- npm test: 917/918 — the single failure (`launcher.test.ts` resolveClientCommand codex/claude subtest; env resolves codex to an absolute path) reproduces identically on pristine master in a worktree → pre-existing, unrelated. Confirmed the PR touches only the 4 listed files.
- tsup build: OK
- E2E codex suite (warmup / load / ACP compress / purity): **pass** against a live Responses-compatible upstream (~4 min; forge phase not run — needs `E2E_FORGE=1`)

**Non-blocking findings (report-only, no separate issues filed):**
1. Pre-existing edge in both strip pipes, exposure widened by this change: if an SSE-labeled body ends without the trailing blank line after its last event, the unterminated buffer remnant is dropped at stream end — only the tag-filter tails are flushed, the leftover `buf` remnant is never emitted (`src/plugin.ts:698-699` chat / `835-838` responses). Worst case the client loses the final chunk (e.g. `finish_reason`/usage); the connection still closes cleanly. Suggest flushing any non-empty remnant as a final event on stream end — fine as a follow-up.
2. Minor: the content-type gate `.includes("text/event-stream")` is case-sensitive; an oddly-cased Content-Type skips the strip (no regression vs master, just a missed strip).
3. Note for completeness: the new branch has no try/catch around the pipes — consistent with the existing plugin-mode branch (only the rewriter path carries the P1.1 try/catch + `emitStreamError`).

Declared residuals (non-streaming JSON unstripped; dead `rewriteSseStream`/`rewriteOpenaiSseStream` exports awaiting sign-off) acknowledged and tracked in-thread.

ranxianglei:
The merge was textually clean but left master's `npm test` red: 4 test call sites added by #469/#464 (after this branch's merge-base `ac4fbcb`) still use the old argument order `(stream, res, session, protocol)`. `typecheck` doesn't catch it because `tsconfig.build.json` only includes `src/**`. At runtime the mis-binding leaks the tag in the #468 anthropic micro-fragment test.

Fix: #472 (test-only, 4 call-site swaps).

PR #464, opened 2026-09-02
fix: keep sibling text when a mixed chunk's other field is a pure tag echo (#462)

ranxianglei:
## What

plugin-mode openai-chat SSE strip (`pipePluginChatWithStrip` → `processOpenai`, `src/plugin.ts`) latched `droppedText` per-field: when a chunk's delta carried multiple managed text fields (`content` / `reasoning_content` / `reasoning`) and ANY one of them stripped to empty (pure render-tag echo), the WHOLE chunk was dropped — including real content in the sibling fields.

Repro (verified failing before the fix): `delta = { content: "real answer", reasoning_content: "<acp …>m00044</acp>" }` → chunk returned `""`, `real answer` lost. This is the #14/#206 heavy-echo scenario at the reasoning→content transition on reasoning providers (deepseek/qwen), shipped in v0.1.71.

## Fix (~3 lines, `src/plugin.ts`)

- Track `keptText`: set when any managed field retains non-empty visible text after stripping — either passed through unfiltered (skip branch) or returned non-empty by the tag-echo filter.
- Drop condition is now `droppedText && !keptText && !hadTextOtherThanTextFields(...)` — a chunk is dropped only when EVERY managed text field emptied out. The existing non-text payload check (finish_reason / tool_calls / etc.) is unchanged, as suggested in the issue.

## Tests (`tests/plugin-passthrough-tag-strip-chat.test.ts`)

- New regression: mixed chunk `content="real answer"` + pure-echo `reasoning_content` → echo stripped, `real answer` survives (failed before the fix).
- New control: chunk where BOTH managed fields are pure echoes → still dropped (guards against over-correction; matches the existing single-field drop test).

## Pre-flight

- `npm run typecheck` clean
- `npm test`: 916/917 pass; the one failure (`launcher.test.ts` resolveClientCommand codex) fails identically on clean master in this sandbox (codex installed at an absolute path here) — pre-existing, unrelated.
- `npm run build` OK.
- E2E codex suite not run: change is in the plugin passthrough pipe (`src/plugin.ts`), outside the listed pipeline scope (server.ts / src/loop/ / adapters / preflight).

Fixes #462.

PR #475, opened 2026-09-02
fix: strip render-tag echoes on the no-inject outbound paths (#460 residual); drop dead SSE rewriters

ranxianglei:
## What

Strip model-emitted render tags on the remaining outbound paths that skip compress injection (#460 residual), repair a test that makes pristine master RED, and delete the dead SSE rewriters that kept causing mis-triage.

### 1. fix(#460 residual) — tag strip on the no-inject outbound paths (`9895a8b`)

Three request paths never reach the tag-echo filter, because they are exactly the paths that skip compress injection:

- **G1'** `src/server.ts` `forward()` — the `prepared.resetAfterSuccess` branch (Responses native-compaction interception) ended in a bare `pipeThrough(toClient, res)`. A native compaction turn is by definition the compression-triggered one, so its context necessarily carries render tags: the likeliest leak of the three. Now an upstream `text/event-stream` body on that branch goes through `pipePluginResponsesWithStrip(toClient, res, undefined, tagLog)`; a non-SSE body keeps the old `pipeThrough`. `observeResponsesTerminalState(toObserve, prepared.stream)` still runs before the await, so the terminal-state observation is unchanged.
- **G2** `src/stream-responses.ts` `rewriteResponsesJsonResponse` — the non-streaming Responses JSON path had no strip at all. Added a `containsRenderTagText(JSON.stringify(b.output))` probe that logs a warn and calls `stripResponsesText(b)`, placed *before* the compress loop so the synthesized note is never edited and the `!converted` early return is covered too.
- **G3** `src/server.ts` ladder — the final bare `pipeThrough` swallowed non-injected, non-streaming JSON. Added a branch ahead of it: `application/json` → `pipePluginJson(upstream.body, res, undefined, prepared.protocol)`. `pipePluginJson`'s `session` parameter is now optional (`if (session && usage)` guard) so proxy mode can reuse it without a plugin session.
- tests: `tests/proxy-noinject-tag-strip.test.ts` +194 lines (Responses compaction SSE, non-streaming openai/responses JSON, byte-identical tag-free passthrough, split-opener `chunkText(full, 5)`), `tests/tag-echo.test.ts` +2 (the G2 rewriter driven directly).

### 2. fix(test) — pristine master is RED (`9883c6c`)

The #469 merge left two `pipePluginChatWithStrip` call sites with swapped arguments: `tests/plugin-passthrough-tag-strip-chat.test.ts:182/:210` passed `(stream, res, session, "anthropic")` against the signature `(stream, res, protocol: WireProtocol, session?: Session, log?)`. `protocol` received the Session object, so `protocol === "anthropic"` was always false and the anthropic fixture ran `processOpenai`. Reproduced on pristine `origin/master`:

```
✖ plugin anthropic passthrough strips a tag streamed in tokenizer-sized fragments (#468)
  actual: 'ok <acp tokens="247" type="text">m00042</acp> tail'
  expected: 'ok  tail'
```

The openai sibling passed only by accident (the false branch IS `processOpenai`). Tests are not in `tsconfig.build.json`, so `typecheck` cannot catch this — filed as a separate issue.

### 3. refactor — dead code removal (`6e9092b`)

`rewriteSseStream` (`src/stream.ts`) and `rewriteOpenaiSseStream` (`src/stream-openai.ts`) have **zero production call sites**: the live SSE path is `runCompressLoop` + `pickAdapter`, and every adapter already pushes its stream through `createTagEchoFilter`. They survived only because five test files drove them directly — which is precisely why #460/#462 mis-triaged the outbound strip. Removed both plus their private helpers (`routeEvent`, `emitToolReplacement`, `buildStopReasonRewrite`, `parseSseEvent`, `emitEvent`, `BlockState`, `NOOP`, `SseEvent`, `StreamState`, `newState`, `routeOpenaiEvent`, `buildOpenaiTail`, `extractDataLine`) and everything bound to them: `tests/proxy-anthropic-stream.test.ts`, `tests/proxy-sse.test.ts`, `tests/sse-fixtures.test.ts`, `tests/fixtures/glm-*.sse`, and the two `OpenAI SSE rewriter …` cases in `tests/basic.test.ts` / `tests/proxy-basic.test.ts`. Kept: `RewriteCtx`, `executeAnthropicProxyTool` (used by `rewriteJsonResponse`), `refNum`, `applyRanges`, both JSON rewriters. Net −963 lines (`src/stream.ts` 379→172, `src/stream-openai.ts` 227→51).

Heads-up: fixture-level OpenAI passthrough-integrity coverage was bound to the removed implementation. The live adapters are covered by `tests/loop-adapters.test.ts`, but an adapter-level fixture test for that specific property would be new work.

### 4. docs (`ab7cbe7`)

The sibling-text fix lives on the plugin path, but its refs pointed at #462 (proxy-mode SSE) instead of #463 (plugin-mode openai SSE). Retargeted `src/plugin.ts:644` and the two test names.

## Verification

- `npm run typecheck` — 0 errors
- `npm test` — 911 tests / 909 pass / 2 fail; both failures are the pre-existing environment-dependent ones (`before_provider_headers stays silent when the manifest fetch keeps failing`, `pi extension is inert without a proxy`). Pristine master fails the same two *plus* the swapped-args test above.
- `npm run build` — `dist/index.js` 2.59 MB
- E2E `tests/e2e/e2e-codex.test.ts` against the local sglang (`E2E_UPSTREAM_URL=http://127.0.0.1:8199/v1`): core ✔ 280 s, `E2E_FORGE=1` ✔ 162 s. Caveat observed while validating: this suite drives a real 27B sampler and is flaky at the load/recall phases — one run failed with `load4 ack wrong:` (empty codex reply) and passed on a retry of the very same commit; pristine master ✔ 181 s.

## Notes

- Supersedes nothing: #464/#466/#469 are already merged (their content is in master), and #462/#463 were auto-closed by those merges. This PR closes the remaining #460 residual.
- No `version` bump — the release flow owns that field (AGENTS.md §5).


ranxianglei:
**Review: verified correct overall; found and fixed 1 merge-interaction bug (commit 8cbeebb, now in this PR).** Full suite 944/942 (2 known permanent env failures), typecheck clean.

## Per-commit verdict

| Commit | Verdict |
|---|---|
| `ab7cbe7` devlog | — |
| `9883c6c` call-order fix | Correct but **duplicates #472** (both swap the same 2 lines; #472 fixed all 4 call sites on master earlier). Harmless — identical content, auto-resolves. |
| `9895a8b` #460 residual (G1'/G2/G3) | All three verified correct (see below) |
| `6e9092b` dead-code removal | Verified: `rewriteSseStream` / `rewriteOpenaiSseStream` have zero production call sites (typecheck green over the deletions). Net −963 lines. The PR's own note stands: fixture-level OpenAI passthrough-integrity coverage was removed with them — adapter-level coverage would be new work. |

## G1' / G2 / G3 verification

- **G1'** (resetAfterSuccess / native Responses compaction): `pipePluginResponsesWithStrip(toClient, res, undefined, tagLog)` — signature checked, responses variant takes no protocol arg. Correct, and usage accounting stays off as intended.
- **G2** (`stream-responses.ts` `rewriteResponsesJsonResponse`): probe on `b.output` before note synthesis and before the `!converted` early-return. Correct placement, strips in place.
- **G3** (non-injected `application/json` answers): `pipePluginJson(..., undefined, prepared.protocol)` with session optional. Correct gating (needs `prepared !== null`).

## Bug found (only manifests after merging with master)

G3 originally read `upstream.body` (src/server.ts:2650). With `BILI_FAKE_COMPLETION_RETRIES>0` (**default 0**) and a streaming request, the #378 fake-completion buffering block consumes `upstream.body` *before* the ladder — so G3 piped a **drained stream and the client got an empty body**. Same pattern as the SSE strip branch bug fixed earlier in 892ffa2.

Fix (8cbeebb): read `responseBody` — matching the SSE strip branch and the bare `pipeThrough` fallback. When fake-completion is off, `responseBody === upstream.body`, so behavior is unchanged for defaults. Includes a red-then-green regression test through the real proxy harness (`BILI_FAKE_COMPLETION_RETRIES=1` + `stream:true` + JSON answer → client must receive the stripped JSON, previously received empty).

## Merge state

I merged current master (135f832: #440/#444 truncation-retry family) into this branch and re-ran everything — conflict-free by content, mergeable, CI should be green. After this merges, the diff is clean against master.

Related: #476 is stale (its failing tests are the #472 fix sites, already green on master) — but its systemic suggestion (typecheck covering `tests/**` or options-object signatures) is worth a follow-up PR.

PR #480, opened 2026-09-03
fix(launcher): eliminate port TOCTOU via child-reported handshake bind; stop forcing --debug (#401)

ranxianglei:
## Problem

The launcher used a two-step probe-release-rebind for proxy ports: `findFreePort` bound the preferred port in the **parent** process, closed it, then passed the number to the child which bound it again — classic TOCTOU:

- another bili grabbing the port inside the window → `probeHealth` passed against the *foreign* instance and client + MCP config silently attached to an instance with different routes/windows/compress settings (undetectable to the user);
- any other program grabbing it (worse on Windows where Hyper-V reserves port ranges) → the health probe burned the full 20s timeout before erroring;
- no rebind retry in either case.

Additionally the launcher forced `debug: true` on every spawned proxy (introduced in 422fc6c for diagnosability), so every `bili omp` / `bili codex` / ... ran at max log level (~12MB/day on one dev machine).

## Fix

**Child-reported bind (kills the race at its root).** The parent no longer probes any port. Each launch spawns `bili start --port <preferred> --handshake-file <per-launch tmp file>` (new flag; also settable via `BILI_HANDSHAKE_FILE`). After binding, the child atomically publishes `{origin, pid}` (tmp file + rename) to that handshake file. Only our own child knows the per-launch path, so the reported origin is ours **by construction** — attaching to a foreign instance becomes impossible. The parent cross-checks `pid` against the spawned child before attaching as defense-in-depth (full `instanceId` handshake tracked separately in #403).

**Stolen preferred port → atomic fallback instead of hang.** In handshake mode, an `EADDRINUSE` on the preferred port makes the server re-listen on port 0 (OS-assigned, atomic — zero probe window), log a clear warn, and report that origin through the handshake file. Plain `bili start` conflict behavior (hint + exit 1) is unchanged.

**Fast failure.** The parent now watches for child exit during both wait phases (handshake + health). A dead child fails in well under 5s with a clear error pointing at the child log (`/tmp/bili-proxy-<launchId>.log`) instead of hanging 20s.

**Log-level parity.** Removed the forced `debug: true`; spawned proxies inherit their log level exactly like a plain `bili start` (`--debug` flag / `ACP_DEBUG` env / config file). The per-launch `/tmp/bili-proxy-<launchId>.log` redirect (the actual intent of 422fc6c) is kept.

## Tests

[...]

PR #481, opened 2026-09-03
fix(launcher): fail fast when the spawned proxy child exits pre-bind

### LINKED ISSUE #596: launcher: spawned proxy child exiting before bind leaves the client launch hanging
### Problem
`bili <client>` spawns the proxy child and waits for it to bind before launching the client. If the child dies pre-bind (bad flag, bad config, port issue, crash), the launcher keeps waiting — the user sees a silent hang with no actionable error.

### Impact
Startup failures of the proxy look like launcher bugs (hang) instead of surfacing the child's stderr/exit code.

### Suggested fix
Fail fast: watch for child exit before the bind handshake and surface the child's stderr + exit code. PR #481 implements this with tests (+53); merges clean.

ranxianglei:
Closes #596 — tracking issue for this PR.

## Context

Salvages the one non-duplicated improvement from **PR #480** ("child-reported handshake bind; stop forcing --debug").

While reviewing #480 I found that two of its three goals are **already merged into master** via an independent, more complete implementation — commit `c2567cc` "instance governance … (#417)":

- The port TOCTOU is already fixed: the child self-binds and retries on `EADDRINUSE` (≤17 attempts), reporting its real origin through the instance file matched by `BILI_LAUNCH_TOKEN` (`src/server.ts`, and `ensureProxyRunning` in `src/launcher.ts`).
- The forced `debug: true` is already removed (`proxyStartArgs` uses `opts.debug`).

Merging #480 as-is would add a second, competing handshake mechanism (a per-launch `--handshake-file`) and conflicts with master in `src/launcher.ts`, `src/server.ts`, and `tests/launcher.test.ts` (all changed-in-both since #480's base). So #480 should be closed as superseded by #417.

The single piece of #480 that master still lacks is **fast-fail on child exit**: if the spawned proxy dies before becoming healthy (bad config, missing upstream, startup crash), master's spawn path waits the full `SPAWN_WAIT_MS = 20000` before erroring — there is no `child.on("exit")` watch inside `ensureProxyRunning` (the only exit handler lives in `runClient`).

## Change

- `src/launcher.ts` `ensureProxyRunning`: register an `exit` listener on the spawned child; break the health-poll loop as soon as it fires; throw a specific error carrying the exit code/signal and pointing at the per-launch log instead of the generic 20s timeout message.
- `tests/launcher.test.ts`: new test — the child fires `exit(1)` right after spawn → `ensureProxyRunning` rejects in well under 5s with `exited before becoming healthy (code 1)` plus the log path.

The success path is untouched: a live child never emits `exit` during polling, so normal startup behavior is unchanged.

## Verification

- `npm run typecheck` — clean
- `npm test` — 947/948 pass; the single failure (`resolveClientCommand: codex/claude resolve to themselves`) is the pre-existing env-dependent one (this sandbox has `/usr/bin/codex`), reproduces identically on pristine master, unrelated to this change
- `npm run build` — clean

Partially supersedes #480; see that thread for the full duplicate analysis.

### DIFF
--- a/src/launcher.ts
+++ b/src/launcher.ts
@@ -1689,8 +1689,20 @@ export async function ensureProxyRunning(
         child.unref?.();
     } catch {}
 
+    // #401/#480: fail fast when OUR spawned child dies before becoming
+    // healthy — otherwise a startup crash (bad config, missing upstream, …)
+    // burns the whole SPAWN_WAIT_MS poll window before erroring.
+    let childExit: { code: number | null; signal: string | null } | undefined;
+    child.on?.("exit", (...rest: unknown[]) => {
+        childExit = {
+            code: typeof rest[0] === "number" ? rest[0] : null,
+            signal: typeof rest[1] === "string" ? rest[1] : null,
+        };
+    });
+
     const deadline = now() + SPAWN_WAIT_MS;
     while (now() < deadline) {
+        if (childExit) break;
         await sleepImpl(HEALTH_POLL_INTERVAL_MS);
         const inst = readInstance();
         if (isProxyInstanceFile(inst) && inst.launchToken === launchToken) {
@@ -1708,6 +1720,12 @@ export async function ensureProxyRunning(
         if (stale && (await probeHealth(proxyOrigin(opts.host, port), fetchImpl))) {
             return { origin: proxyOrigin(opts.host, port), port, child, logPath };
         }
+    }
+    if (childExit) {
+        const detail = childExit.code !== null
+            ? `code ${childExit.code}`
+            : childExit.signal ? `signal ${childExit.signal}` : "unknown reason";
+        throw new Error(`bili: proxy child exited before becoming healthy (${detail}) (log: ${logPath})`);
     }
     throw new Error(`bili: proxy did not become healthy within ${SPAWN_WAIT_MS}ms (log: ${logPath})`);
 }
--- a/tests/launcher.test.ts
+++ b/tests/launcher.test.ts
@@ -681,6 +681,41 @@ test("ensureProxyRunning: launchToken handshake returns the child's real port (#
     assert.equal(handle.origin, "http://127.0.0.1:8799");
 });
 
+test("ensureProxyRunning: spawned child exits pre-bind → fails fast (<5s) with log path (#401/#480)", async () => {
+    const child: SpawnChild = {
+        pid: 42441,
+        unref() {},
+        kill() {
+            return true;
+        },
+        on(event, listener) {
+            if (event === "exit") setImmediate(() => listener(1, null));
+            return undefined;
+        },
+    };
+    const spawnImpl: SpawnFn = () => child;
+    const t0 = Date.now();
+    await assert.rejects(
+        ensureProxyRunning(
+            { host: "127.0.0.1", port: 8787, passthrough: false, debug: false },
+            {
+                spawnImpl,
+                fetchImpl: async () => ({ ok: false }),
+                fetchHealthInfo: async () => undefined,
+                readInstanceFile: () => undefined,
+                sleep: () => new Promise((r) => setTimeout(r, 1)),
+            },
+        ),
+        (err: Error) => {
+            assert.match(err.message, /exited before becoming healthy/);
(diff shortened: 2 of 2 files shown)


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 Picked up. Verified the premise on master: `ensureProxyRunning` (src/launcher.ts:1633) has no `exit` watch — the only exit handler in the file is in `runClient` (src/launcher.ts:1747), so a pre-bind child death burns the full `SPAWN_WAIT_MS = 20000` and reports the generic timeout. Implementing the scoped fix now (exit listener + early loop break + specific error with code/signal + log path, plus the fast-fail test). Will follow up with the PR link.

ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 Done — **PR: https://github.com/ranxianglei/billion-context/pull/482**

**What changed** (branch `2026-09-03_launcher-fail-fast-child-exit`, off master `09ef208`):
- `src/launcher.ts` `ensureProxyRunning`: registers an `exit` listener on the spawned child right after spawn/unref; the health-poll loop condition becomes `!childExit && now() < deadline` so it breaks immediately once the child dies; post-loop it throws `bili: proxy child exited before becoming healthy (code N)` — or `(signal SIGTERM)` / `(code N, signal SIGTERM)` when applicable — with `(log: <per-launch log path>)`, instead of the generic 20s timeout message. The timeout path is preserved verbatim for the no-exit case.
- `tests/launcher.test.ts`: new test `ensureProxyRunning: fails fast when the child exits before becoming healthy (#481)` — fake child fires `exit(1, null)` via `setImmediate` right after spawn; asserts rejection matches `exited before becoming healthy (code 1)`, includes the `bili-proxy-8787.log` path, and lands in well under 5s of wall-clock (real timers, ~200ms actual).

**Verification** (matches your stated bar):
- `npm run typecheck` — clean
- `npm test` — 947/948 pass; sole failure is `resolveClientCommand: codex/claude resolve to themselves`, the pre-existing env-dependent one (this sandbox has `/usr/bin/codex`)
- `npm run build` — clean

Success path untouched: a live child never emits `exit` during polling. Ready for review — merge stays with you.

ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 Cross-reference: a second, independent implementation of this same fix landed as **PR #482** (branch `2026-09-03_launcher-fail-fast-child-exit`, `a027923`) ~5 minutes after this one. I reviewed and verified both branches locally (typecheck clean; 947/948 on each — the single failure is the pre-existing env-dependent `resolveClientCommand` test, reproduces on pristine master; build clean; fail-fast test rejects in ~205ms vs the old 20s window). The trees differ only in style (typed `{code,signal}` vs `[unknown,unknown]`; test placement/timing strategy).

Per duplicate screening I **closed #482 as `duplicate`** in favor of this PR — it's first-in-time and its body carries the fuller #480 salvage analysis. Full comparison table and reasoning are in the #482 thread. If you'd rather merge the #482 tree instead, reopen it and close this one — both are verified merge-ready. One suggested tweak before merge: the code comment cites `#401/#480`, but #401 is the unrelated sessions-storage bloat issue; `#480` is the correct source ref. Merge stays yours.

ranxianglei:
**Review vs master 3cd85b6:** ✅ merge-ready.

- Verified against master's `ensureProxyRunning` (`src/launcher.ts:1474`): the patch context (`child.unref?.()` → poll loop → timeout throw) is unchanged; applies clean.
- `SpawnChild.on?:` exists in the interface (`src/launcher.ts:100`), so the exit listener attach type-checks; attaching synchronously right after spawn is safe (Node emits `exit` asynchronously).
- `if (childExit) break` at loop top + post-loop detail throw is correct; the pre-existing "stale record / live record" fallback paths are untouched.
- Test: mock child fires `exit(1, null)` via `setImmediate` — deterministic, <5s assertion is meaningful against `SPAWN_WAIT_MS = 20000`.

One nit (non-blocking): the fast-fail error could also mention the child command (`process.execPath + script`) — the log path already points at the captured stderr, so current form is fine.

(tracking: #596)

Issue #453, opened 2026-09-02
plugin mode: no hard backstop when the agent ignores the compress nudge — context grows past the window and upstream 400s


ranxianglei:
Found while investigating #14 ("bili总是循环").

## Repro (real session)

`bili pi` (plugin mode, OpenAI chat-completions wire) with qwen3.8-27b on vLLM (262144 window), bili.log for session 01a06275:

- 14:14:00 `nudge INJECT T1` (pendingT1 52017 ≥ 50000) — model does not call `compress`
- 14:14–14:19 `nudge idle ... blocked: T1 (cadence)` while usage climbs 56% → 70%
- 14:19:34–14:20:49 `nudge INJECT T1 reason="OVER-LIMIT T1"` usage 77% → 85% — still no `compress`
- 14:27:39 upstream rejects: `400 Requested token count exceeds the model's maximum context length of 262144 tokens. You requested a total of 262817 tokens: 131745 tokens from the input messages and 131072 tokens for the completion.`

The client (pi) then falls back to its own native compaction → the model loses all working state and re-derives its analysis from scratch (compounding the perceived "loop" in #14). A second session (01a062a7, qwen3.8-flash) sat at 80–85% usage with `nudge INJECT T1` firing **every turn** — the model never called `compress` at all.

## Impact

In plugin mode the nudge is advisory: bili trusts the ACP-native agent to execute `compress`. When the model simply ignores the injected nudge (weaker instruction-following, e.g. qwen3.8-27b), nothing stops the payload from overflowing the window. Proxy mode has preflight/preflightCompressIfNeeded as the backstop; plugin mode has none.

## Suggested fix (discussion)

Options, roughly in ascending invasiveness:

1. **Clamp the outgoing output budget**: `max_tokens`/`max_completion_tokens` = min(requested, window − estimated input − margin) so input+output can never exceed the window on request-rebuilding upstreams (vLLM rejects instead of clamping). Cheap, no wire changes; doesn't fix input-overflow itself.
2. **Proxy-side emergency truncation**: when estimated input tokens exceed the window and the agent hasn't compressed after N over-limit nudges, drop/truncate the oldest inessential history server-side for that request (like proxy-mode preflight) even though the session is plugin-mode. Needs care to keep the agent's view consistent.
[...]

PR #485, opened 2026-09-03
fix: abort-path usage drop, upstream-timer leak, and bare AbortError logs (#411)

### LINKED ISSUE #411: 客户端取消 = 无主 AbortError ×260/天 + 已计费 usage 丢弃(lastInputTokens 冻结)+ 透传路径漏 10min timer(无 finally)
## 症状(0.1.66/0.1.67,HEAD 同)

客户端(宿主 agent)主动取消流式请求时:
1. 日志只有一条**无上下文的全局 `[error] AbortError`**——本机一天 **260 条**(448 条中 408 条紧跟 `forward POST → /v1/messages`,并发交错无从归因);真实来源是 `res.on("close") → clientAbort.abort()`(server.ts:2262-2265),属**正常行为却按异常记账**;
2. **已计费用量随 abort 丢弃**:`applyUsageSample(session, acc)` 在抛错路径被跳过(plugin.ts:517+,dist 53956)——`acc` 里已有 message_start 的 input_tokens,取消一轮后 `lastInputTokens` 冻结在旧值(取证窗口内 `nudge idle: usage=0%` 跨轮滞留与此吻合);
3. 非流式 `arrayBuffer()` 分支同样裸奔(server.ts:2545);纯透传路径 `clearUpstreamTimer()` 在 `await pipeThrough` 之后无 finally(server.ts:2270-2274)→ **每次 abort 漏一个 10 分钟 timer + listener**(fetch-util.ts:12)。

注:`stream aborted mid-response`(重写器自己的 marker,stream-error.ts:39)0 次——**没有一次 abort 走压缩路径**,全在 plugin pipes/server 级 catch(server.ts:317-327)。与 #388(侧请求污染)是两回事。

## 建议修复
plugin pipes 与 arrayBuffer 分支 `catch AbortError` → 照常 `applyUsageSample(acc)` + `markDirty`,日志降级为 info 且带 session id;timer 进 finally;全局 `[error]` 只留给真异常。

## 验收标准
- [ ] 假上游 SSE(message_start + 慢速 delta),t+1s 断客户端:session.stats.lastInputTokens 反映本轮 input;日志为 `[<sid>] client aborted mid-stream`(info),无裸 `[error] AbortError`
- [ ] abort 风暴压测:无 timer 累积(active handles 不单调增)

ranxianglei:
Fixes #411.

## Defects

1. **Usage drop on client abort** — `pipePluginChatWithStrip` / `pipePluginResponsesWithStrip` only ran `applyUsageSample` at the try-tail; a client disconnect mid-stream skipped it, so `lastInputTokens` froze at the pre-turn value and corrupted later nudge/budget decisions (same input-corruption family as #453).
2. **Upstream idle-timer leak** — `forward()` had `await pipe*(); clearUpstreamTimer();` with no try/finally on several paths; each client abort left one live `setTimeout` (up to 10 min) plus a detached abort listener. ~260 aborts/day per busy proxy.
3. **Bare `[error] AbortError` log spam** — the server-level catch logged `String(err)` with no classification or session context.
4. **Non-stream persistence gap** — `markDirty(prepared.session)` sat outside the try in the non-stream rewriter path; the adjacent comment promised persistence on throw but the code skipped it.

## Fix

- **src/plugin.ts**: both SSE pipes now `settleUsage()` (applyUsageSample + markDirty) at the try-tail **and** in the catch. Client-abort errors are swallowed once the response is destroyed/ended (`client aborted mid-stream`, info-level), everything else rethrows. `pipePluginJson` gets a try/catch around its read loop.
- **src/server.ts**: `forward()` wraps every upstream-body-consuming path in a single try/finally so `clearUpstreamTimer()` runs on every exit (the rewriter paths keep their own earlier clears; double-clear is idempotent). Both rewriter finallys now `markDirty(prepared.session)`. The server catch classifies `AbortError` + closed response as `info: client aborted mid-stream: <msg>` instead of `error`.
- **src/fetch-util.ts**: tracks live upstream timers in a module-level Set (self-removing), exposed via `_liveUpstreamTimersForTest()` for leak assertions.

Note: usage that only arrives in a stream's final chunk (openai chat `stream_options.include_usage`, Responses `response.completed`) is inherently lost on early abort — nothing to salvage there. Anthropic usage arrives at `message_start` and is now salvaged.

## Tests

New `tests/issue411-abort-usage.test.ts` (5 tests):

1. plugin stream abort keeps `lastInputTokens=1234` (from `message_start`) + logs info, not a bare error
2. 3-round abort storm leaves zero live upstream timers
3. upstream cut with client alive still lands usage and is NOT misclassified as a client abort
4. non-stream plugin abort mid-body: no crash, timer cleared
5. anonymous passthrough abort: info not error, timer cleared

Pre-flight: `npm run typecheck` ✅ · `npm test` 971/973 (2 fails are the known permanent `plugin-agent.test.ts` pair, reproduce on clean master) ✅ · `npm run build` ✅

## Design note

One subtlety worth flagging for reviewers: `clearUpstreamTimer()` also **removes the external-abort listener** (via fetch cleanup). Calling it early — e.g. right after response headers but before the body is consumed — would sever the client→upstream abort chain. An earlier draft of this PR did exactly that and broke `tests/vscode-copilot.test.ts:147` ("upstream fetch was not aborted after client cancel"). The final shape wraps to the end of `forward()` so clearing only happens at real exits.

### DIFF
--- a/src/fetch-util.ts
+++ b/src/fetch-util.ts
@@ -11,6 +11,15 @@
 export const MAX_REQUEST_BYTES = 100 * 1024 * 1024;
 export const UPSTREAM_TIMEOUT_MS = 10 * 60 * 1000;
 
+const liveUpstreamTimers = new Set<ReturnType<typeof setTimeout>>();
+/** Test hook: how many fetchWithTimeout idle-timers are currently armed.
+ *  #411: an aborted passthrough used to leak its 10-minute timer because
+ *  clearTimer was only called on the success path — tests assert this stays
+ *  at zero after a client abort. */
+export function _liveUpstreamTimersForTest(): number {
+    return liveUpstreamTimers.size;
+}
+
 /** undici's fetch accepts a `dispatcher` option (its own Dispatcher type) that
  *  @types/node's RequestInit already declares — but typed as the internal
  *  `Dispatcher` interface, which conflicts with the `undici` package's
@@ -43,10 +52,19 @@ export async function fetchWithTimeout(
     externalSignal?: AbortSignal,
 ): Promise<{ response: Response; clearTimer: () => void }> {
     const controller = new AbortController();
-    let timer = setTimeout(() => controller.abort(), timeoutMs);
+    const armTimer = () => {
+        const t = setTimeout(() => {
+            liveUpstreamTimers.delete(t);
+            controller.abort();
+        }, timeoutMs);
+        liveUpstreamTimers.add(t);
+        return t;
+    };
+    let timer = armTimer();
     const rearm = () => {
         clearTimeout(timer);
-        timer = setTimeout(() => controller.abort(), timeoutMs);
+        liveUpstreamTimers.delete(timer);
+        timer = armTimer();
     };
     let onExternalAbort: (() => void) | null = null;
     if (externalSignal) {
@@ -58,6 +76,7 @@ export async function fetchWithTimeout(
     }
     const cleanup = () => {
         clearTimeout(timer);
+        liveUpstreamTimers.delete(timer);
         if (onExternalAbort && externalSignal) externalSignal.removeEventListener("abort", onExternalAbort);
     };
     try {
--- a/src/plugin.ts
+++ b/src/plugin.ts
@@ -622,6 +622,16 @@ export async function pipePluginChatWithStrip(
             return new Promise<void>((r) => res.once("drain", () => r()));
         }
     };
+    // #411: an aborted read (client cancel / upstream cut) must still land the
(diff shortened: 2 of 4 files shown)


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 Acknowledged — starting review of this PR. Per project rules I'll verify each claimed defect against master (repro/evidence), assess root cause vs symptom, evaluate the fix shape (including both compression modes), and run typecheck/test/build locally before giving a verdict. Will report findings here shortly. I will not merge — that stays with you.

PR #486, opened 2026-09-03
fix: plugin mode gets the fake-completion detect+retry when enabled (#473)

### LINKED ISSUE #473: plugin mode is not covered by the fake-completion detect+retry (#371 / #378 follow-up)
Follow-up of #371 / PR #378: **plugin mode does not get the fake-completion detect+retry.**

## Background

#378 ships detection (tool-call XML structure + no real tool block) + bounded retry with a corrective hint, opt-in via `BILI_FAKE_COMPLETION_RETRIES` (default 0). But in `forward()` (`src/server.ts`) the plugin-mode branch **returns early** (streams into `pipePluginChatWithStrip` / `pipePluginResponsesWithStrip`), and the `resolveFakeCompletion` gate sits **after** that return:

```
if (pluginMode) { ...pipePlugin*...; return; }   // plugin: never buffered
...
let responseBody = upstream.body;
if (prepared !== null && prepared.stream && maxFakeCompletionRetries() > 0) {
    responseBody = bufferToStream(await resolveFakeCompletion(upstream.body, {...}));
}
```

So with retries enabled, proxy-mode clients get the backstop and plugin-mode clients (pi / omp via `bili pi` / `bili omp`) do not.

## Impact

The **original reporters** of #361 (fake completion + tool-fragment leak) and #371 were plugin-mode clients (omp). The scenario that motivated the fix is exactly the path it does not cover: in plugin mode the agent owns the tool loop, sees a text-only turn with no tool block, ends the turn — and shows the user a normal completion.

## Fix directions (discussion)

1. **Proxy-side, same mechanism**: move the `resolveFakeCompletion` gate **before** the plugin-mode early return, so plugin streams are buffered + retried identically when enabled. Same streaming-latency trade-off (whole-response buffering while enabled), same env knob. Lowest code cost; keeps the corrective-hint retry transparent to the agent.
2. **Agent-side retry**: implement the detect+retry in the agent plugins (`src/agent/`), since in plugin mode the agent owns the loop and can decide its own retry. Architecturally cleaner (no buffering, no proxy guessing) but N implementations (pi/omp/opencode) and more surface.

Either way the acceptance criteria from #371 (bounded retries, no false positives on prose, protocol adaptation, tests) apply unchanged.

## Notes

- Detected while verifying #378 against post-#466 master (892ffa2 comment: https://github.com/ranxianglei/billion-context/pull/378#issuecomment-5513915203).
- Not a regression of #378 — the gap predates the merge fix.

ranxianglei:
Fixes #473. Direction ① from the issue (proxy-side, same mechanism, same env knob).

## The gap

#378's `resolveFakeCompletion` gate sat in `forward()` **after** the plugin-mode early return — so with `BILI_FAKE_COMPLETION_RETRIES` enabled, proxy-mode clients got the backstop and plugin-mode clients (pi / omp, the original #361/#371 reporters) did not.

## Fix

Move the same gate into the plugin branch: when enabled and the request is streaming, buffer through `resolveFakeCompletion` (detect fake → bounded hinted retry), then pipe the resolved bytes through the existing plugin pipes instead of `upstream.body`:

- `src/fake-completion.ts` needed **zero changes** — `isFakeCompletion` / `injectFakeCompletionHint` already handle all three wire protocols (anthropic `tool_use`, openai `tool_calls`, responses `function_call`).
- The retry re-requests with the same rewritten forwarded body + hint, so the agent's context (compress tool, prior history) is preserved.
- The plugin pipes still run after the gate: render-tag strip (#452/#475) applies, and usage sniffing follows the **retry's** response (correct input accounting for the next nudge decision).
- Session streak bookkeeping (`fakeCompletionStreak`) applies to plugin sessions identically.
- Default (retries=0) behavior is byte-identical: no buffering, verbatim passthrough.
- The #411 try/finally around the pipes still wraps the gate — an abort during buffering clears the upstream timer and logs info, not a bare error.

## Tests

New `tests/issue473-plugin-fake-completion.test.ts` (3 tests):

1. plugin **openai chat**: fake completion (tool-call XML as text, no `tool_calls` block) → 2 upstream requests, hint merged into the retry's last user message, client receives the recovered real tool block, fake text never reaches the client, session streak resets
2. **default off**: 1 upstream request, fake text passes through verbatim (no behavior change without the env knob)
3. plugin **anthropic**: same recovery via `tool_use` block; hint lands as an appended `{type:"text"}` block (anthropic wire normalizes content to a block array)

Pre-flight: `npm run typecheck` ✅ · `npm test` 974/976 (2 fails are the known permanent `plugin-agent.test.ts` pair, reproduce on clean master) ✅ · `npm run build` ✅

Direction ② (agent-side retry in `src/agent/`) intentionally not taken — N implementations, more surface, same latency trade-off.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -2651,20 +2651,36 @@ async function forward(
     // Plugin mode: the agent's native loop owns the tool surface — pass the
     // response through VERBATIM (a model-emitted compress call must reach the
     // plugin untouched) while sniffing usage so lastInputTokens (the input to
-    // the next nudge decision) keeps tracking reality.
+    // the next nudge decision) keeps tracking reality. The one exception: the
+    // opt-in #371 fake-completion backstop buffers + retries first, same as
+    // proxy mode (#473).
     if (prepared?.pluginMode) {
         // #411: clear the idle timer on the abort path too — the pipes rethrow
         // when the client is still connected, and a client cancel throws from
         // inside them; without a finally each abort leaked a 10-minute timer.
         try {
+            let pluginBody = upstream.body as ReadableStream<Uint8Array>;
+            if (prepared.stream && maxFakeCompletionRetries() > 0) {
+                const resolvedBuf = await resolveFakeCompletion(pluginBody, {
+                    protocol: prepared.protocol,
+                    body,
+                    upstreamUrl,
+                    reqHeaders: buildForwardHeaders(headers),
+                    proxyUrl,
+                    signal: clientAbort.signal,
+                    session: prepared.session,
+                    log,
+                });
+                pluginBody = bufferToStream(resolvedBuf);
+            }
             if (prepared.stream) {
                 if (prepared.protocol === "responses") {
-                    await pipePluginResponsesWithStrip(upstream.body as ReadableStream<Uint8Array>, res, prepared.session, (msg) => log("info", `[${prepared.session.id}] ${msg}`));
+                    await pipePluginResponsesWithStrip(pluginBody, res, prepared.session, (msg) => log("info", `[${prepared.session.id}] ${msg}`));
                 } else {
-                    await pipePluginChatWithStrip(upstream.body as ReadableStream<Uint8Array>, res, prepared.protocol, prepared.session, (msg) => log("info", `[${prepared.session.id}] ${msg}`));
+                    await pipePluginChatWithStrip(pluginBody, res, prepared.protocol, prepared.session, (msg) => log("info", `[${prepared.session.id}] ${msg}`));
                 }
             } else {
-                await pipePluginJson(upstream.body as ReadableStream<Uint8Array>, res, prepared.session, prepared.protocol);
+                await pipePluginJson(pluginBody, res, prepared.session, prepared.protocol);
             }
         } finally {
             clearUpstreamTimer();
--- a/tests/issue473-plugin-fake-completion.test.ts
+++ b/tests/issue473-plugin-fake-completion.test.ts
@@ -1,0 +1,220 @@
+import assert from "node:assert";
+import http from "node:http";
+import { once } from "node:events";
+import test from "node:test";
+
+process.env.NODE_ENV = "test";
+
+import { defaultConfig } from "acp-kernel";
+import { startServer, type ProxyOptions } from "../src/server.ts";
+import { SessionStore, _setStoreForTest } from "../src/persist.ts";
+import { _setForTest as setRegistryForTest } from "../src/registry.ts";
+import { _resetPluginStateForTest } from "../src/plugin.ts";
+import { listSessions } from "../src/session.ts";
+
+// #473: the #371 fake-completion detect+retry (opt-in via
+// BILI_FAKE_COMPLETION_RETRIES) sat AFTER the plugin-mode early return, so
+// plugin-mode clients (pi / omp — the original #361/#371 reporters) never got
+// the backstop. These tests drive a plugin session through the gate.
+
+const FAKE_MARK = "FAKE-COMPLETION-MARKER";
+// antml-style tool-call XML written as TEXT (hex escapes; never a real block)
+const FAKE_XML = `\x3cinvoke name="get_weather"\x3e\x3cparameter name="city"\x3eSF\x3c/parameter\x3e\x3c/invoke\x3e`;
+const HINT_MARK = "[billion-context]";
+
+function listen(server: http.Server): Promise<void> {
+    if (server.listening) return Promise.resolve();
+    return once(server, "listening").then(() => void 0);
+}
+
+function close(server: http.Server): Promise<void> {
+    const { promise, resolve, reject } = Promise.withResolvers<void>();
+    server.close((error) => (error ? reject(error) : resolve()));
+    return promise;
+}
+
+function openaiSse(obj: unknown): string {
+    return `data: ${JSON.stringify(obj)}\n\n`;
+}
+
+function anthropicSse(event: string, data: unknown): string {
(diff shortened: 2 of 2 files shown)


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 Reviewed #486 end-to-end (checked out the head, ran the full pre-flight independently, and diffed the branch against master). Two-part verdict:

## ✅ The #473 fix itself is correct — I approve the code

- **Root cause confirmed on master.** In `forward()`, the plugin branch (`if (prepared?.pluginMode)`) pipes `upstream.body` straight through and returns, and the `resolveFakeCompletion` gate sits *after* that return — so with `BILI_FAKE_COMPLETION_RETRIES` set, plugin clients (pi/omp) never hit the backstop. The gap is real, exactly as #473 describes.
- **Minimal + symmetric.** The change reuses the *same* gate, *same* env knob, and the *same* `prepared.stream` condition proxy mode uses; zero changes to `src/fake-completion.ts`. Resolved bytes feed the unchanged plugin pipes (render-tag strip + usage-sniff still run downstream; usage follows the retry's response), and the existing `try/finally` still wraps it. Clean.
- **Pre-flight (reproduced):** typecheck ✅ · build ✅ · the 3 new tests in `tests/issue473-plugin-fake-completion.test.ts` all pass (log shows the gate firing: `fake completion (tool-call XML, no tool block); retry 1/1 with corrective hint`). The default-off test correctly proves byte-identical verbatim passthrough when the knob is unset.
- **Full suite:** 976 tests, **1 failure here** — `tests/launcher.test.ts … resolveClientCommand: codex/claude resolve to themselves`. I confirmed that exact test **fails identically on clean master**, and this PR touches neither `launcher.ts` nor `launcher.test.ts`, so it's a pre-existing environment sensitivity (depends on which client CLIs happen to be installed), **not a regression from this change**. (The 2 `plugin-agent.test.ts` failures you saw are likewise environment-specific — they pass in my run.) So effectively green.

## ⛔ Blocker before merge: the branch bundles three unrelated changes

Diffing `origin/master..head` shows the branch carries **three** logical changes, but the PR describes only one:

1. `release v0.1.75` — a version bump (`package.json`/lock). Master is currently at `0.1.74`.
2. The separate **#411 abort-path fix** (`src/fetch-util.ts`, `src/plugin.ts`, `src/server.ts` + `tests/issue411-abort-usage.test.ts`).
3. The **#473 fix** (the actual subject).

Why this needs splitting:

- **The version bump violates the release flow (§4/§5).** Version changes belong *only* on a dedicated `*_release-v*` branch as an isolated `release v{VERSION}` commit, delivered via its own release PR. Because #486's branch name (`…_issue-473-plugin-fake-completion`) and its merge-commit message don't match the pattern `.github/workflows/release.yml` keys off, **CI will NOT publish** — so merging would leave master sitting at `version: 0.1.75` with no npm publish/tag/release. That's an inconsistent state and forces the next real release to jump to `0.1.76`.
- **#411 gets merged as a side effect** of #473 rather than reviewed/merged on its own merits (it has its own scope and a 240-line test file). A defect in either would drag the other down with it.
- **There's real coupling, though:** #473's `server.ts` edit lives *inside* the `try/finally` that #411 introduced, so #473 depends on #411 landing first. That argues for **sequencing** (#411 PR merged first), not for shipping both in one PR.

### Recommended path
1. Split **#411 into its own PR**; review + merge it first.
2. **Rebase #486** onto post-#411 master so it contains *only* the #473 change (`src/server.ts` + `tests/issue473-plugin-fake-completion.test.ts`). Drop the `release v0.1.75` commit from the branch.
3. If `v0.1.75` is still wanted, ship it through the standard §5 flow **after** the fixes land (dedicated `YYYY-MM-DD_release-v0.1.75` branch + isolated `release v0.1.75` commit → CI publishes).

(I won't merge anything — that's human-only per the repo rules.)

## Minor notes (non-blocking)
- **Streaming-only scope:** the gate requires `prepared.stream`, so non-streaming plugin JSON responses don't get the retry. That's consistent with proxy mode (identical condition), so it's fine — just worth stating explicitly so it's a conscious decision (a non-stream fake completion wouldn't be caught).
- **Test parity:** the new tests cover the openai-chat and anthropic plugin paths but not the **responses** wire for the plugin path specifically. Detection/hint-injection is shared with proxy mode (which you note already handles responses), so risk is low — optional to add one responses-wire case for parity.
- **Buffering latency** while enabled (whole response buffered before first byte reaches the client) is inherent to direction ① and matches what proxy mode already accepts; opt-in + default-off keeps it safe. Flagging it as the known trade-off, not a bug.

**Duplicate screening:** searched open issues/PRs for plugin-mode fake-completion coverage — no duplicate. Lineage is #361/#371/#378 (original chain) → #473 (this follow-up); no other open PR addresses it.

Bottom line: good fix, well-tested, correct root cause — **split out the #411 fix and the version bump**, then this is ready to merge.

ranxianglei:
The windows-22 failure (`scheduleSave debounces and eventually writes`, `ENOENT ... rename .tmp- → .json; SPILLOVER ALSO FAILED`) is a known Windows flake, not from this diff:

- the **same test** failed earlier today on the unrelated launcher PR eb6ad47 (windows-24) and passed on its follow-up push a027923;
- this diff touches only the plugin branch of `forward()` + a new test file — `proxy-persist.test.ts` doesn't import the server;
- the shape (transient ENOENT/EPERM on the tmp→rename write pattern under Windows Defender) is the documented #362/#367 family.

Re-ran the failed job — green (`completed success`). No code change needed here. A follow-up hardening of `withTempStore` cleanup (settle before `rmSync`) + a write-retry for ENOENT/EPERM would despeckle CI; tracked under #367.

PR #483, opened 2026-09-03
fix: land #454 (omp native compaction disable) on master — it was merged into its already-merged base branch

ranxianglei:
## What happened

#454 was a **stacked PR**: its base was `2026-09-02_pi-disable-native-compaction` (#448's branch), not `master`. The merge sequence was:

1. #421 → master (1000c12, 04:26:32Z) ✅
2. #448 → master (cdbf2bc, 04:27:14Z) ✅
3. #454 → **merged 04:27:49Z into the pi branch** (b36d418), which had already been merged 35s earlier — so the omp fix never reached master

Evidence:
- `git merge-base --is-ancestor b36d4183 master` → **NO**; b36d4183 lives only on `origin/2026-09-02_pi-disable-native-compaction`
- master's `prepareOmpHttpRewrite` is still the old version (models.yml-only, no generated config.yml, no `compaction.enabled=false` for omp)
- The stranded content is exactly one commit: 01212cc `fix: bili omp launcher disables omp's native auto-compaction (#449)`
- #449 was closed as completed, but the fix is absent from master

## This PR

Merges the pi-branch tip into master, bringing in exactly the missing #454 content (`cdbf2bc..b36d4183` = 01212cc only). Clean merge, no conflicts.

## Pre-flight

- `npm run typecheck` ✅
- Full suite: **956/958** — the 2 failures are the known permanent `plugin-agent.test.ts` ones (reproduce on clean master)
- `tests/launcher.test.ts` + `tests/launcher-plugin-mode.test.ts`: **130/130** ✅

## After merge

- Re-close comment on #449 pointing here
- omp launcher compaction disable finally effective; pairs with #421 (archive path) already on master

Human merges, as always.

PR #489, opened 2026-09-03
fix: multimodal token accounting + store:false in Responses preflight summary (#488)

ranxianglei:
## What
Issue #488: codex session with 7 screenshots → payload ~1.49M tokens vs 1.05M window → every request 502-stuck. Two root causes, both verified in code:

**A. Image bytes were invisible to every payload-size decision.** The wire codecs move images out of `CoreMessage.text` into sidecars (`imageBase64`/`rawResponsesItem`/…), so `estimateCoreMessages()` (text-only) undercounts image-bearing payloads up to 15× — yet the images ARE forwarded verbatim. Consequences: the preflight fit gate (`result.payloadEstimate < limit`) declared "fits" for an over-window multimodal payload and forwarded it anyway (upstream 400 → retry → identical → permanent loop); self-heal window learning recorded text-only sizes.

**B. Preflight summary request omitted `store:false` for the Responses protocol.** Codex relays reject it with `HTTP 400 {"detail":"Store must be set to false"}`, killing the rescue exactly when it is needed. Only the synthesized summary call was affected: the main forward path passes the client body through (codex sends its own `store:false`) and compress-loop re-requests spread `{...requestBody}`.

## Changes
- `src/image-tokens.ts` (new): host-side image share of a RAW request body. Known base64 payloads charged `ceil(b64len/4)` — the same chars/4 rule `defaultCountTokens` uses, matching how byte-counting relays actually bill (#488 evidence: 7 screenshots ≈ 1.4M tokens rules out pixel-tile billing, which would cap at ~11K). Remote URLs we cannot size get a flat `REMOTE_IMAGE_TOKENS = 4096`. `BILI_IMAGE_TOKEN_CAP` env clamps the per-image cost for pixel-tile upstreams that bill by tiles (official Anthropic/OpenAI), where the default intentionally overestimates to trigger compression earlier rather than undercount.
- `src/preflight.ts`: `summaryPayload` responses branch now sends `store: false`; new optional `PreflightDeps.imageFloor` added to every size decision inside the preflight loop (initial estimate, per-round floor-max, per-round result) so folds can never declare success while image bytes alone exceed the window.
- `src/server.ts`: preflight trigger + fail-fast detail + self-heal window learning all add the image share of the actual forwarded body; when images alone exceed the window the fail-fast error says so explicitly ("Images alone account for ~N tokens … compression cannot remove them").
- `tests/fix-488-multimodal.test.ts`: unit tests for the estimator (all three protocols, data URLs, remote URLs, multi-image sums, raw-body parse gate, cap clamp) + two e2e regressions through the real proxy: (B) preflight summary requests carry `store:false` and the rescued session forwards 200; (A) 7 × 15K-token images vs a 10K window → withheld 502 `preflight_compress_failed` with the actionable image note, nothing forwarded upstream.

## Scope notes
- Both compression modes covered per AGENTS.md §6: preflight runs unconditionally (plugin mode included); the `store:false` fix applies to the proxy-synthesized summary call in either mode; image accounting is mode-independent (it measures the forwarded body).
- No acp-kernel change — counting happens host-side from the raw wire body, which also avoids the kernel's pre-existing lossiness for remote-URL images (its `allImageParts` only captures data: URLs).
- Text-only sessions are byte-for-byte unaffected (estimator returns 0 without image markers; no JSON.parse even runs — string sniff gate).

## Pre-flight (on rebased branch f7deccc)
- `npm run typecheck` ✅
- `npm test` 984/986 — the 2 fails (`plugin-agent.test.ts`) fail identically on pristine master in this sandbox; environment-dependent, unrelated to this change
- `npm run build` ✅

Open question for @Rika-xie before merge: the default billing model (b64 length / 4) matches the byte-counting relay in the report; if your upstream bills by pixel tiles, set `BILI_IMAGE_TOKEN_CAP` accordingly.

### DIFF
--- a/src/image-tokens.ts
+++ b/src/image-tokens.ts
@@ -1,0 +1,94 @@
+// #488: images are invisible to the kernel's token model (wire codecs move them
+// out of CoreMessage.text into sidecars), yet they ARE forwarded verbatim — so
+// every fit decision built on the text-only estimate undercounts image-bearing
+// payloads up to 15×. This module estimates the image share of a RAW request
+// body so callers can add it to the text estimate. Known base64 payloads are
+// charged ceil(base64Length/4) — the same chars/4 rule defaultCountTokens uses,
+// matching how byte-counting relays actually bill (#488: 7 screenshots ≈ 1.4M
+// tokens); remote URLs we cannot size get a flat conservative cost. Pixel-tile
+// upstreams (official Anthropic/OpenAI) will see an overestimate — set
+// BILI_IMAGE_TOKEN_CAP to clamp the per-image cost.
+
+export const REMOTE_IMAGE_TOKENS = 4096;
+
+function imageTokenCap(): number {
+    const v = Number(process.env.BILI_IMAGE_TOKEN_CAP ?? "");
+    return Number.isInteger(v) && v > 0 ? v : 0;
+}
+
+function applyCap(cost: number): number {
+    const cap = imageTokenCap();
+    return cap > 0 ? Math.min(cost, cap) : cost;
+}
+
+function costForUrl(url: string): number {
+    if (url.startsWith("data:")) {
+        const i = url.indexOf("base64,");
+        if (i >= 0) return applyCap(Math.ceil((url.length - i - "base64,".length) / 4));
+    }
+    return applyCap(REMOTE_IMAGE_TOKENS);
+}
+
+function isObj(v: unknown): v is Record<string, unknown> {
+    return typeof v === "object" && v !== null;
+}
+
+function urlOf(v: unknown): string | undefined {
+    if (typeof v === "string") return v;
+    if (isObj(v) && typeof v.url === "string") return v.url;
+    return undefined;
+}
+
+export function imageTokensInParsedBody(protocol: "anthropic" | "openai" | "responses", body: unknown): number {
+    if (!isObj(body)) return 0;
+    let total = 0;
+    if (protocol === "responses") {
+        const input = body.input;
+        if (!Array.isArray(input)) return 0;
+        for (const item of input) {
+            if (!isObj(item) || !Array.isArray(item.content)) continue;
+            for (const part of item.content) {
+                if (!isObj(part) || part.type !== "input_image") continue;
+                const url = urlOf(part.image_url);
+                if (url) total += costForUrl(url);
+            }
+        }
+        return total;
+    }
+    const messages = body.messages;
+    if (!Array.isArray(messages)) return 0;
+    for (const m of messages) {
+        if (!isObj(m) || !Array.isArray(m.content)) continue;
+        for (const part of m.content) {
+            if (!isObj(part)) continue;
+            if (protocol === "openai") {
+                if (part.type !== "image_url") continue;
+                const url = urlOf(part.image_url);
+                if (url) total += costForUrl(url);
+            } else {
+                if (part.type !== "image") continue;
+                const src = part.source;
+                if (isObj(src) && src.type === "base64" && typeof src.data === "string") total += applyCap(Math.ceil(src.data.length / 4));
+                else if (isObj(src) && src.type === "url" && typeof src.url === "string") total += costForUrl(src.url);
+            }
+        }
+    }
+    return total;
+}
+
+// Cheap gate: most bodies carry no images — skip the JSON parse entirely then.
+// prepared.body is bili's own compact JSON.stringify, but client raw buffers
+// may carry spaces, so probe both forms.
+export function imageTokensInRawBody(protocol: "anthropic" | "openai" | "responses", raw: string | Buffer): number {
+    const s = typeof raw === "string" ? raw : raw.toString("utf8");
+    const probe =
+        protocol === "responses" ? s.includes("input_image")
+        : protocol === "openai" ? s.includes("image_url")
+        : s.includes('"type":"image"') || s.includes('"type": "image"');
+    if (!probe) return 0;
+    try {
+        return imageTokensInParsedBody(protocol, JSON.parse(s));
+    } catch {
+        return 0;
+    }
+}
--- a/src/preflight.ts
+++ b/src/preflight.ts
@@ -42,6 +42,8 @@ export interface PreflightDeps {
     proxyUrl?: string;
     signal?: AbortSignal;
     log: (level: string, msg: string) => void;
+    /** Constant floor on the forwarded-payload size for this request (image bytes, #488). Folding only ever removes images, so adding this to every text estimate keeps the fit decision sound for multimodal payloads. */
+    imageFloor?: number;
 }
 
 export type PreflightFailureKind = "upstream" | "exhausted" | "aborted";
@@ -145,7 +147,8 @@ function summaryPayload(protocol: PreflightProtocol, model: string, system: stri
     if (protocol === "openai") {
         return { model, max_tokens: MAX_SUMMARY_OUTPUT_TOKENS, messages: [{ role: "system", content: system }, { role: "user", content }], stream: false };
     }
-    return { model, max_output_tokens: MAX_SUMMARY_OUTPUT_TOKENS, instructions: system, input: [{ role: "user", content }], stream: false };
+    // #488: codex relays reject Responses calls without store:false ("Store must be set to false").
+    return { model, max_output_tokens: MAX_SUMMARY_OUTPUT_TOKENS, instructions: system, input: [{ role: "user", content }], stream: false, store: false };
 }
 
 function extractSummaryText(protocol: PreflightProtocol, json: Record<string, unknown>): string {
@@ -223,7 +226,7 @@ const ABORTED_FAILURE: PreflightFailure = { kind: "aborted", detail: "the client
 
 export async function preflightCompress(deps: PreflightDeps, messages: CoreMessage[]): Promise<PreflightResult> {
     const limit = deps.config.modelContextLimit;
-    const result: PreflightResult = { compressedRanges: 0, savedTokens: 0, payloadEstimate: estimateCoreMessages(messages) };
+    const result: PreflightResult = { compressedRanges: 0, savedTokens: 0, payloadEstimate: estimateCoreMessages(messages) + (deps.imageFloor ?? 0) };
     if (limit <= 0) return result;
     const budget = Math.max(MIN_CHUNK_TOKENS, Math.floor(limit * CHUNK_FRACTION));
     // applyCompression rejects ranges below config.compress.minCompressRange
@@ -254,10 +257,10 @@ export async function preflightCompress(deps: PreflightDeps, messages: CoreMessa
         // Floor on the session's measured input baseline: the upstream's
         // input_tokens also covers the system prompt + tool definitions, which
         // are not in turn.messages, so the direct estimate can undershoot.
-        currentTokens = Math.max(deps.session.stats.lastInputTokens, estimateCoreMessages(turn.messages));
+        currentTokens = Math.max(deps.session.stats.lastInputTokens, estimateCoreMessages(turn.messages) + (deps.imageFloor ?? 0));
         // The caller's forward/fail-fast gate uses the payload's own estimate
         // (the floor can be stale — see PreflightResult.payloadEstimate).
-        result.payloadEstimate = estimateCoreMessages(turn.messages);
+        result.payloadEstimate = estimateCoreMessages(turn.messages) + (deps.imageFloor ?? 0);
         if (startTokens < 0) startTokens = currentTokens;
         if (currentTokens < limit) break;
         const ranges = viableRanges(turn.nudge?.compressibleRanges ?? []);
--- a/src/server.ts
+++ b/src/server.ts
@@ -46,6 +46,7 @@ import { COMPRESS_TOOL, ACP_TOOLS_ANTHROPIC, ACP_TOOLS_OPENAI, ACP_TOOLS_RESPONS
 import { rewriteJsonResponse, type RewriteCtx } from "./stream.js";
 import { applyRanges } from "./stream.js";
 import { preflightCompress, estimateCoreMessages } from "./preflight.js";
+import { imageTokensInRawBody, imageTokensInParsedBody } from "./image-tokens.js";
 import { renderUI, handleConfigGet, handleConfigPut } from "./web/index.js";
 import { reapOrphanBlocks } from "./orphan-gc.js";
 import { getStore } from "./persist.js";
@@ -1559,13 +1560,15 @@ export function emergencyNudge(nudge: NudgeDecision | null | undefined, escalati
 function clampOutgoingOutput(
     rebuilt: Record<string, unknown>,
     field: "max_tokens" | "max_completion_tokens" | "max_output_tokens",
-    ctx: { systemText: string; tools: unknown; processedMessages: CoreMessage[]; lastInputTokens: number; nativeWindow: number },
+    ctx: { systemText: string; tools: unknown; processedMessages: CoreMessage[]; lastInputTokens: number; nativeWindow: number; imageTokens: number },
     sessionId: string,
     log: (level: string, msg: string) => void,
 ): void {
     const raw = rebuilt[field];
     if (typeof raw !== "number") return;
-    const inputEstimate = estimateInputTokens(ctx.processedMessages, ctx.systemText, ctx.tools, ctx.lastInputTokens);
+    // #488: images ride along in the rebuilt body but are invisible to the text model —
+    // without them the cap is too generous and input+output can still overflow.
+    const inputEstimate = estimateInputTokens(ctx.processedMessages, ctx.systemText, ctx.tools, ctx.lastInputTokens) + ctx.imageTokens;
     const capped = clampOutputBudget(raw, inputEstimate, ctx.nativeWindow);
     if (capped !== undefined) {
         rebuilt[field] = capped;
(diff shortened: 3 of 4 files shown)


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 Review of PR #488 in progress. Per AGENTS.md triage rules I'm doing: (1) verify both root causes actually exist on master with code evidence, (2) layer analysis (root cause vs symptom), (3) evaluate whether the fix hits the core, including both compression modes (§6). I'll run typecheck/test/build myself and post a full verdict here.

ranxianglei:
[bot] 🏷 Review complete — verdict: **fix is correct; branch was stale, now rebased + extended**.

**What the review found**

1. **Branch was stale.** It was cut before #483/#485/#486/#487 landed on master. Rebased cleanly onto current master (zero conflicts); force-pushed as `f7deccc`.
2. **Same-class gap in newer code.** Master's #453 output-budget clamp (`estimateInputTokens` / `clampOutgoingOutput`, added after this branch was cut) had the exact same blind spot: image bytes invisible to the input estimate → cap too generous → input+output could still overflow on multimodal sessions. Extended the fix: `clampOutgoingOutput` now takes `imageTokens` (counted from the rebuilt outgoing body — the same source preflight uses) at both call sites (OpenAI + Responses). New e2e regression A2 pins it: text-only request keeps its 5k output budget; the same request plus one 4k-token screenshot gets clamped below 5k.
3. **Separate pre-existing gap noted (not fixed, out of scope):** the Responses output clamp does not count the injected compress developer prompt (~1.9K tokens) — `processedMessages` is computed before the dev-message injection and `ctx.systemText` excludes it. Affects turn-1 estimates of all compress-injected Responses sessions (`lastInputTokens` dominates from turn 2 onward). Bounded impact; flagging for a follow-up decision rather than bundling into this fix.

**Verification (on the rebased branch)**

- `npm run typecheck` ✅
- `npm test` 984/986 — the 2 fails (`plugin-agent.test.ts`: "pi extension is inert without a proxy", "before_provider_headers stays silent when the manifest fetch keeps failing") fail identically on pristine master in this sandbox; environment-dependent, unrelated to this change
- `npm run build` ✅
- `tests/fix-488-multimodal.test.ts` 7/7 ✅ (was 6; +A2)

The open question from the description stands: default billing = b64 length/4 (matches the byte-counting relay in the report); pixel-tile upstreams should set `BILI_IMAGE_TOKEN_CAP`. Merge stays human-only per repo rules.

ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 Independent review of PR #488 (branch `2026-09-03_multimodal-image-tokens-store-false`, commit `acd9dce`) complete. I re-derived both root causes from master code myself, audited the fix design, ran the checks locally, and screened for duplicates. **Verdict: both root causes verified, the fix hits the core, both compression modes covered. Two minor doc gaps to close before merge.**

## Root-cause verification (independent of the PR's own claims)

**A. Image bytes invisible to size decisions — CONFIRMED, and the layering is right.** On master, `estimateCoreMessages()` (`src/preflight.ts`) sums `defaultCountTokens(m.text)` only. I traced all three wire codecs in the bundled acp-kernel: decode moves images out of text into sidecars (anthropic → `[image]` placeholder + `rawAnthropicBlock`; openai → `rawOpenaiContent`/`imageBase64`/`rawOpenaiContentParts`; responses → `rawResponsesItem` + `imageBase64` for data URLs only), and encode re-emits them verbatim (`coreToAnthropic` pushes the raw block; `coreToOpenai` pushes raw parts / reconstructs the data URL; `coreToResponses` pushes the raw item when text is unchanged). So the forwarded payload carries the full image bytes while every `estimateCoreMessages`-based decision (preflight trigger, fit gate, self-heal learning) sees ~text-only size. The PR's diagnosis is accurate: the failure mode is a *false success* ("fits") on an over-window multimodal payload, not merely a missed nudge.

**B. Missing `store:false` — CONFIRMED, scope is exact.** Master's `summaryPayload()` responses branch builds `{model, max_output_tokens, instructions, input, stream:false}` with no `store` field → codex relays reject the synthesized summary call with `HTTP 400 {"detail":"Store must be set to false"}`, killing the rescue precisely when needed. Main forward path (client body passthrough, codex sends its own `store:false`) and compress-loop re-requests (`{...requestBody}` spread) are correctly identified as unaffected.

## Fix-design audit

- **Constant `imageFloor` is sound by construction**: folds only ever *remove* images (summaries are text), so the original forwarded body's image share is an upper bound on any post-fold payload. It can never declare "fits" while image bytes alone exceed the window. The pessimistic edge case (old images folded away, floor still high → fail-fast when post-fold would actually fit) self-heals because preflight mutates session state every round even on failure — the next request sees the smaller payload. Correct trade against #488's permanent 400 loop.
- **Estimator placement on the rebuilt wire body** (not kernel sidecars) is the right call: it measures exactly what gets forwarded, and sidesteps two pre-existing kernel codec gaps (below). Sniff gate means text-only bodies skip `JSON.parse` entirely — zero behavior change for them.
- **`store:false` is spec-valid** for any Responses implementation (official OpenAI supports `store`), so low regression risk beyond codex relays.
- **Both compression modes (§6)**: preflight runs unconditionally (plugin mode included); the `store:false` change touches only the proxy-synthesized summary call, which exists in either mode; image accounting measures the forwarded body, mode-independent. ✓

## Verification run (my sandbox, on the PR head)

- `npm run typecheck` ✅
- `npm test` ✅ 952/953 — single failure is `launcher.test.ts: resolveClientCommand codex/claude`. The diffstat touches no launcher file (`src/image-tokens.ts`, `src/preflight.ts`, `src/server.ts`, new test file only), so this is provably environment-dependent and unrelated.
- `npm run build` ✅
- New tests reviewed: 4 unit (all three protocols, data/remote URLs, multi-image sums, sniff gate incl. malformed JSON → 0, cap clamp) + 2 real-proxy e2e. Confirmed the asserted error code `preflight_compress_failed` is what `src/server.ts:1304` emits, and e2e-A's "nothing forwarded upstream" assertion actually guards the fail-fast path. Good coverage of both root causes.
- Duplicate screening: searched open+closed issues for multimodal/image-token/store-false topics — no prior or overlapping items.

## Findings from the review

**Minor (report-only, please close before merge):**
1. `BILI_IMAGE_TOKEN_CAP` is a user-facing knob but absent from the env tables in `CONFIGURATION.md` (~L340–365) and `CONFIGURATION.zh-CN.md` (~L345). Since the PR's open question hinges on it, the doc line should ship with the change.
2. No `CHANGELOG.md [Unreleased]` entry — repo convention is that content PRs carry theirs (see recent fixes bc57a09, 5e35adb, 380e323).
3. Known limitation, no action required now: the Anthropic scan covers top-level `messages[].content[]` parts only; images nested inside `tool_result.content` arrays are forwarded but not counted (rare shape; same conservative-direction argument applies since the fit gate also floors on the trigger-side count).

**Major problem found during analysis → filed separately:** the two acp-kernel codec gaps the PR works around are silent data-loss bugs in the kernel itself — `openaiToCore` drops a *single* non-data-URL image (no sidecar chosen → encode emits text only), and `responsesToCore` drops *image-only* user items (empty `effText` → no core message). Filed with repros and suggested fixes as **ranxianglei/acp-kernel#187** (source-tagged to this work). This is not a defect of PR #488 — host-side raw-body counting is deliberately robust to both (overcounts conservatively vs. what the kernel forwards) — but worth tracking.

## Open question (default billing model)

Recommendation stands: **keep chars/4 as the default**. The #488 evidence (7 screenshots ≈ 1.4M tokens; pixel-tile billing would cap at ~11K) proves byte-counting relays exist in the wild, and undercounting there reproduces exactly the permanent 400 loop this PR fixes. Overcounting on tile-billing upstreams (official Anthropic/OpenAI) only triggers compression earlier — the safe direction. The `BILI_IMAGE_TOKEN_CAP` escape hatch covers the rest; it just needs the doc line (finding 1).

Ready to merge once findings 1–2 land (or accepted as an immediate follow-up — your call). Per repo rules I won't merge it myself.

ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 **Re-review of `f7deccc` (rebased head) — verified independently.** Verdict: the rebase is clean, the output-clamp extension is sound, and the code is ready to merge once the two minor doc items below land.

**Rebase validated**
- Merge-base of the branch is exactly current `master` (`f037419`, the #487 merge) — no stale base remains.
- PR diff vs master is precisely the 4 files you listed: `src/image-tokens.ts` (+94), `src/preflight.ts` (+11/−3), `src/server.ts` (+33/−13), `tests/fix-488-multimodal.test.ts` (+268). Nothing extra rode in with the rebase.

**My checks at `f7deccc` (this sandbox)**
- `npm run typecheck` ✅ · `npm run build` ✅
- `npm test`: **985/986**. The single failure is `launcher.test.ts:22 resolveClientCommand codex/claude` — `/usr/bin/codex` exists in my sandbox so the resolver returns the absolute path instead of the bare name; it fails identically on pristine master here and the diff touches no launcher code. (Your 984/986 with two `plugin-agent.test.ts` fails is the same class of environment-dependency on your machine; neither file is touched by the diff.)
- `tests/fix-488-multimodal.test.ts`: **7/7 pass**, including the new A2.

**Clamp extension — soundness review**
- `clampOutgoingOutput`'s ctx gains `imageTokens`, added to `inputEstimate` before `clampOutputBudget` (`src/server.ts:1545-1577`). Both call sites count it from the **rebuilt outgoing body** (`imageTokensInParsedBody("openai", rebuilt)` at :1681, `"responses"` at :1890) — the same "count what is actually forwarded" principle as the preflight trigger and self-heal learning. Consistent source of truth across all three size decisions.
- Adding image tokens monotonically shrinks the cap (`cap = nativeWindow − inputEstimate − margin`, margin ≥ 2048 or 5%), so behavior is strictly more conservative than master; text-only sessions get `imageTokens = 0` and are byte-for-byte unaffected. Mode-independent (measures the forwarded body) — covers both compression modes per §6.
- One mechanism note (pre-existing, not introduced here): `nativeWindow` is captured at `src/server.ts:1256` **before** the output-headroom reservation block, so the clamp operates against the full/native window while the kernel nudge/truncate bands see the reserved one. That's current master behavior and A2's control assertion (text-only keeps its 5k budget against the 10k window) documents exactly that. No action needed — just flagging so the interaction isn't rediscovered later.
- A2 is well-designed: it captures `max_output_tokens` from the **actual upstream request** (not internal state), includes a text-only control, and asserts both directions of the clamp (`< 5000` and `≥ 1024` floor).

**Your out-of-scope flag (dev prompt ~1.9K) — verified accurate**
- `processedMessages` is computed at `src/server.ts:1805` (`stripKernelSummaries(turn.messages, …)`), which is **before** `injectResponsesDeveloperMessage` (:1819-1821); `ctx.systemText` at :1890 is `projection.systemParts` joined, which excludes the injected compress prompt. So the Responses clamp undercounts the injected dev message on turn 1 exactly as you described — bounded, because `lastInputTokens` (real upstream usage, which includes it) dominates from turn 2 onward, and any resulting overflow self-heals through the learned-window path. Agreeing with keeping it out of this PR's scope. Suggestion: file it as a separate sourced issue (来源: #488/#489 analysis) so it doesn't evaporate — happy to file it if you want; otherwise your call.

**Carried over from round 1 (still holds at `f7deccc`)**
- Both root causes independently re-verified; estimator design sound (constant `imageFloor` is an upper bound because folds only ever *remove* images; conservative direction is the safe one; per-round state mutation means a pessimistic fail-fast self-heals next request instead of looping).
- `store:false` correctly scoped to the proxy-synthesized summary call only.
- Recommendation stands: keep the `chars/4` default (byte-billing relays proven by the #488 evidence; overcounting only triggers earlier compression, undercounting reproduces the 400 loop), `BILI_IMAGE_TOKEN_CAP` as the escape hatch for tile-billing upstreams.
- Kernel codec gaps found during review were filed as **ranxianglei/acp-kernel#187** (sourced to #488); unaffected by this rebase, and the host-side counting decision sidesteps them correctly.

**Still open before merge (minor)**
1. `BILI_IMAGE_TOKEN_CAP` is absent from `CONFIGURATION.md` and `CONFIGURATION.zh-CN.md` (grep-verified at `f7deccc`) — add the env-var lines alongside the other `BILI_*` entries.
2. No `[Unreleased]` entry in `CHANGELOG.md`; repo convention (bc57a09, 5e35adb, 380e323) is that content PRs carry their own entry.
3. Known limitation (non-blocking, noted for follow-up): the anthropic scan misses base64 images nested inside `tool_result.content` parts — the estimator undercounts that rarer shape. Small follow-up candidate, not a merge blocker.

**One pre-merge recommendation:** the PR touches `server.ts`/`preflight.ts`, so per AGENTS.md §3 the codex e2e regression should run before merge. Zero-token preflight (`E2E_CHECK=1`) passes in my sandbox, but the full `ACP_TEST_E2E=1` run needs a live Responses-compatible upstream that this sandbox doesn't have — please run it locally or via the manual-dispatch `ci-e2e.yml` workflow before merging.

Merge stays human-only per repo rules. If items 1–2 land (or you explicitly defer them), I consider this reviewed and ready.

ranxianglei:
第三方独立 review(在 f7deccc 上全部本机复跑):

- 两个根因诊断与修法均确认正确(store:false 作用域精确;图片统计口径 = 重建后的转发 body,`prepared.body` 为 `JSON.stringify(rebuilt)`,已核实)。
- **全量 codex e2e 通过**(真实 codex 0.147.0 + 本地 sglang,ACP_TEST_E2E=1,452s)— AGENTS.md §3 对 server.ts/preflight.ts 改动的门槛,前两轮 review 沙箱均无法执行。typecheck / build ✅,单测 984/986(2 个 plugin-agent 失败在 master 上同样失败,环境依赖)。
- 合并前小项:`BILI_IMAGE_TOKEN_CAP` 补进 CONFIGURATION(.zh-CN).md、CHANGELOG [Unreleased] 条目、master 已前进需 rebase。
- **一个设计风险已立案为 #496**:默认 b64/4 无上限计费在像素 tile 计费的官方 Anthropic/OpenAI 上,图片单项估算 ≥ 窗口时是硬 502(非"提前压缩"),2 张 400KB JPEG 在 200K 窗口即触发误杀(真实成本 ~3K)。建议后续按 #496 的 forward-once-then-learn 方案收窄硬拦条件;不阻塞本 PR 合并。

ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 收到第三方 review 汇总,逐点确认:

**1. 全量 codex e2e 通过 — 补上了我这边缺的最后一道门槛。** 这正是我在两轮 verdict 里标记的 pre-merge 建议(我的沙箱只跑通了 `E2E_CHECK=1` 零 token 预检,完整 run 需要 live upstream)。真实 codex 0.147.0 + 本地 sglang、`ACP_TEST_E2E=1` 452s 通过,AGENTS.md §3 对 `server.ts`/`preflight.ts` 改动的回归要求即满足。单测差异互证无碍:你的 984/986(`plugin-agent.test.ts` ×2)vs 我的 985/986(`launcher.test.ts` ×1),都是沙箱本地二进制存在性导致的环境依赖失败,两个文件均不在 diff 内。

**2. #496 的定性成立,我修正第一轮的一个措辞。** 我当时说"高估只是提前触发压缩,方向安全"——这个论断只在 fit 门表现为"提前压缩"时成立;我低估了极端情形:`imageFloor` **单项 ≥ 窗口**时 fit 门是硬 fail-fast(502,绝不转发,`src/server.ts` preflight 的 `imageTokens >= limit` 分支),"高估"就从保守方向变成**保证误杀**。数字对照代码复核过:400KB JPEG → b64 ≈ 546K 字符 → 估算 ~136K tokens/张,2 张 ≈ 273K ≥ 200K 窗口 → 命中硬拦;而官方 tile 计费真实成本 ~3K,**同一场景在 master 上正常转发**。所以这是官方 API 多模态用户的回归向量,不是单纯的保守取舍;`BILI_IMAGE_TOKEN_CAP` opt-in + 错误信息引导用户删图/调窗口,进一步放大。单独立案(#496)、不阻塞本 PR——同意。

**3. forward-once-then-learn 方案的两个实现观察**(供 #496 实现参考,不阻塞):
- 门条件"图片是唯一超窗成分"可落为 `currentTokens − imageFloor < limit && currentTokens >= limit`(文本折叠后已 fit),再叠加"无上游实证"(`lastInputTokens < limit` 且该 model 无 `learnedContextLimits` 记录)即 issue 所述。
- **终止性依赖"拒绝被记录"**:字节计费 relay 首转发 400 → self-heal 写 `learnedContextLimits`(本 PR 已含 `rejectedImageTokens`,学窗路径齐备)→ 第二次请求"无实证"条件翻 false → 硬 fail-fast,一次额外转发后收敛,严格优于 master 的无限 400 循环。边界情形:若 relay 的 400 body 里没有可解析的窗口信息(`inspectContextOverflow` 学不到东西),"无实证"条件第二次仍为真,会反复转发反复被拒。建议 overflow 拒绝时**无条件**记一个 marker(哪怕没解析出窗口值),否则 "forward once" 会退化成 "forward every time"。
- tile 计费上游成功转发后无需学习(估算保持偏高但无害,反正总是放行),行为零变化——与 issue 预期一致。

**4. 合并前剩余项**(与你的清单一致,补一条 rebase 注意):
- `BILI_IMAGE_TOKEN_CAP` 补进 `CONFIGURATION.md` / `CONFIGURATION.zh-CN.md` env 表;`CHANGELOG.md [Unreleased]` 条目。
- master 已前进到 `4f5fe7e`(我 review 基线 `f037419` 之后新合入 `174006e` = #332/#335 "detect codex compaction requests before preflight")——与本 PR 的 preflight 改动**同一代码区域**,rebase 后请重跑 `tests/fix-488-multimodal.test.ts` + 全量单测确认无交互回归(尤其 codex compaction 请求现在在 preflight 之前就被识别并 verbatim passthrough,需确认不会绕过/干扰新的图片尺寸决策路径)。

结论不变:代码本身 ready;文档两项 + rebase 验证完成后即可由 owner 合并(按仓库惯例合并人工执行)。

Issue #436, opened 2026-09-01
上下文满了不压缩 bili omp


ranxianglei:
日志稍后到 

2026-09-01 23:33:24  ⤵ 18K  ⤴ 2.3K  💾 120K  ⏱ 2.4s  ⚡ 86.5/s                                                                                                                               

 ╭─────────────────────────────────────────────╮                                                                                                                                              
 │           ACP Context Analysis              │                                                                                                                                              
 ╰─────────────────────────────────────────────╯                                                                                                                                              
 billion-context@0.1.68                                                                                                                                                                       
                                                                                                                                                                                              
 Context (session accounting, host footer scale): 53% (138k / 262k) — never shrinks; includes compressed originals                                                                            
 Growth: +13k since last nudge                                                                                                                                                                
                                                                                                                                                                                              
 Sent to LLM (after compression, est.): 80k (31% of limit)                                                                                                                                    
                                                                                                                                                                                              
 Token Breakdown (sent view):                                                                                                                                                                 
   Tool       ████████████████░░░░  80%  64k                                                                                                                                                  
   Text       ████░░░░░░░░░░░░░░░░  20%  16k                                                                                                                                                  
   Code       ░░░░░░░░░░░░░░░░░░░░   0%  154                                                                                                                                                  
                                                                                                                                                                                              
 Nudge: idle — growth 13177 < floor 22500, ready: T1 64088, blocked: T1 (cadence)                                                                                                             
                                                                                                                                                                                              
 Compressible ranges (1, oldest first):                                                                                                                                                       
   m00001–m00150  149 msgs  64.1K [tool 52% | text 48%]                                                                                                                                       
                                                                                                                                                                                              
 Blocks: none (nothing compressed yet)                                                                                                                                                        
                                                                                                                                                                                              
 Tag visibility: tags injected to LLM only (deep copy), not persisted in session, not shown in terminal.                                                                                      

╭── 🗑 /tmp > 💾 97.08% > ◫ 100.5%/131K ▶───

---
_Mirrored from ework issue #436_

<!-- ework-mirror -->

ranxianglei:
[bili.log](/attachments/1358ce4c-95eb-4b2e-8f5f-9307a4417b91)

ranxianglei:
> 🤖 ework agent · qwen3.8-27b

收到，正在分析你贴的 `bili.log`。

初步定位（session `01a05d8f`，`bili omp`，proxy 模式，本地 Qwen 模型）：

[...]

PR #442, opened 2026-09-02
chore(#436): log context-window source per model (diagnostic)

ranxianglei:
Closes #597 — tracking issue for this PR.

## What
Added a one-line diagnostic log at the context-window resolution site (src/server.ts) that records, once per model, the chosen window source (anthropic-beta / plugin / launcher / configured / registry-peek / table-or-registry / default) plus the native + effective values.

## Why
The window-resolution path was invisible in logs. In #436 the user's omp model (contextWindow: 131072 in models.yml) showed 229376/261120 in the proxy — the variation implied a dynamic source (learned overflow), but without the source log it was impossible to confirm whether BILI_LAUNCHER_MODEL_WINDOWS was reaching the proxy. This log makes the source explicit so the next repro is one grep away.

## Pre-flight
- `npm run typecheck` — clean
- `npm test` — 897/897 pass
- `npm run build` — success

Related: #436

PR #493, opened 2026-09-03
fix: recognize sglang's overflow error format in context-window self-heal

ranxianglei:
## Problem

sglang's context-overflow error is not recognized by `inspectContextOverflow`, so the window self-heal never fires for sglang-backed sessions.

Captured verbatim from a running sglang server (both wire formats share the message):

```
HTTP 400
{"object":"error","message":"The input (400013 tokens) is longer than the model's context length (262144 tokens).","type":"BadRequestError","param":null,"code":400}
```

None of the existing `CONTEXT_OVERFLOW_PATTERNS` match this phrasing ("is longer than the model's context length"), so `isOverflow` stays `false`, the error is passed through as an ordinary 400, and bili never learns the real window. A session whose configured window exceeds sglang's actual limit then overflows on every retry — permanently stuck, because the one signal that could correct the window is ignored.

Found while investigating #436 (sglang-backed local model, window never self-heals).

## Fix (`src/util.ts`)

- Add pattern `/longer than the model'?s context length/i` to `CONTEXT_OVERFLOW_PATTERNS`.
- Add parser entry `/context length\s*\((\d[\d,]*)\s*token/i` to `parseOverflowWindow` — captures the LIMIT (the number after "context length"), never the input size that appears earlier in the same message.

## Test

New regression test in `tests/context-overflow.test.ts` using the verbatim captured bodies for both `/chat/completions` and `/responses` formats; asserts `isOverflow === true` and `window === 262144` (the limit, not the 400013 input size).

## Pre-flight

- `npm run typecheck` ✓
- `npm test`: `tests/context-overflow.test.ts` 20/20 ✓; full suite has 2 pre-existing environment-dependent failures in `tests/plugin-agent.test.ts` ("pi extension is inert without a proxy", "before_provider_headers stays silent when the manifest fetch keeps failing") — identical failures on clean master (verified via worktree), unrelated to this change
- `npm run build` ✓

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -111,6 +111,7 @@ const CONTEXT_OVERFLOW_PATTERNS: RegExp[] = [
     /maximum context length/i,
     /max context length/i,
     /maximum context size/i,
+    /longer than the model'?s context length/i,
     /exceeds the context window/i,
     /out of room in the model/i,
     /exceeded model token limit/i,
@@ -140,6 +141,7 @@ function parseOverflowWindow(text: string): number | undefined {
         text.match(/maximum context length of (\d[\d,]*)/i) ??
         text.match(/maximum context size (?:is|of) (\d[\d,]*)/i) ??
         text.match(/(?:maximum|max)\s+(?:context\s+)?length\s+(?:is\s+)?(\d[\d,]*)/i) ??
+        text.match(/context length\s*\((\d[\d,]*)\s*token/i) ??
         text.match(/limit of (\d[\d,]*)\s*token/i) ??
         text.match(/(\d[\d,]*)\s*maximum\b/i);
     if (m) return toTokenNumber(m[1]);
--- a/tests/context-overflow.test.ts
+++ b/tests/context-overflow.test.ts
@@ -38,6 +38,20 @@ test("OpenAI Responses: 'exceeds the model's maximum context size of N' → over
     assert.equal(info.isOverflow, true);
     assert.equal(info.window, 128000);
 });
+test("sglang: 'input (N tokens) is longer than the model's context length (M tokens)' → overflow + window M", () => {
+    // Captured verbatim from a real sglang server; both wire formats share the
+    // message. The learned window must be the LIMIT (M), not the input size (N)
+    // — learning N would self-heal to a window larger than the real one.
+    const message = "The input (400013 tokens) is longer than the model's context length (262144 tokens).";
+    const chatBody = JSON.stringify({ object: "error", message, type: "BadRequestError", param: null, code: 400 });
+    const info = inspectContextOverflow(400, chatBody);
+    assert.equal(info.isOverflow, true);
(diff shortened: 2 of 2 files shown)


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 Reviewed and independently verified PR #493 (`2026-09-03_sglang-overflow-format` @ 8e66f7a, single commit on top of current master 4f5fe7e). Verdict: **correct, minimal, ready to merge** — merge itself is human-only per repo rules.

**Triage (verified before touching anything)**
- Bug is real: none of the existing `CONTEXT_OVERFLOW_PATTERNS` (`src/util.ts:107-122`) matches "is longer than the model's context length", so `inspectContextOverflow` returns `isOverflow: false` for sglang's 400 and self-heal never fires. Root cause = missing recognition pattern (not the self-heal machinery), and the fix hits exactly that layer.
- Not a duplicate: distinct from #442 (diagnostic window-source logging only, open) and #281 (Codex phrasing, closed). This is the overflow-recognition gap found while investigating #436 — merging this unblocks the self-heal path for sglang-backed sessions but does not by itself close #436 (that one still needs the user's window-source data).

**Diff check (matches the issue spec exactly)**
- `src/util.ts:114` — `/longer than the model'?s context length/i` added to `CONTEXT_OVERFLOW_PATTERNS`.
- `src/util.ts:144` — `text.match(/context length\s*\((\d[\d,]*)\s*token/i) ??` inserted into the `parseOverflowWindow` chain after the max/maximum-length entries and before `limit of …`. Placement is safe: I traced every earlier chain entry against the sglang message — none can match it (no `>`, no "maximum/max" before "context length") — and the new entry requires the literal parenthesized form, so it cannot hijack another provider's phrasing. It captures the limit `(262144 tokens)`, never the input size `(400013 tokens)` which appears earlier without "context length" preceding it.
- Byte-checked the apostrophe: straight ASCII `'` in both the pattern and the test fixture, matching what sglang actually emits.
- No `version` bump, no unrelated changes — compliant with the release rules.

**Pre-flight re-run on the PR branch (clean checkout, fresh `npm ci`)**
- `npm run typecheck` ✓
- `npm run build` ✓
- `tests/context-overflow.test.ts`: 20/20 ✓ (both wire formats assert `isOverflow === true`, `window === 262144`)
- Full suite: 990/991. The single failure is `tests/launcher.test.ts` "resolveClientCommand: codex/claude resolve to themselves" — environmental (this sandbox has `/usr/bin/codex`, so resolution yields an absolute path); reproduced identically on clean master via a throwaway worktree, unrelated to this change. Note: the two `plugin-agent.test.ts` failures mentioned in the issue did not reproduce here (that file passes 27/27 standalone and in the full run) — different environment-dependent set, also unrelated.

Please merge when ready: https://github.com/ranxianglei/billion-context/pull/493

PR #504, opened 2026-09-03
fix: #498 weak-overflow learning + #499 P1a affinity persistence + P1b cache-collapse warn

### LINKED ISSUE #498: 高 usage 下的非 400 失败（流截断/超时）无任何 overflow 兜底 —— 会话直接死
来源：#351 线程拆分（https://github.com/ranxianglei/billion-context/issues/351 ，三个加固项之一）。

## 现象

#351 的致死组合：窗口被高估（解析 967232 vs 上游真实 ~477K）+ 上游以**非 400** 方式失败（03:55:26 `upstream stream truncated`）→ self-heal 学不到真实窗口、也没有强制压缩兜底 → 会话死亡。

## 现状

窗口自愈只认 HTTP 400（明确的 overflow 信号）；流截断 / 超时 / 连接重置一律不算。

## 建议方向（需谨慎设计）

高 usage（如 ≥90% 解析窗口）下的流截断/超时作为**弱 overflow 信号**：
- 触发一次性窗口自学习探测（下调一档试转发）或强制压缩兜底。
- 截断原因多样（网络抖动、上游 bug、真 overflow），**不能一律当 overflow**：建议仅「高 usage + 短时间窗内重复 ≥N 次」才降窗/强制压缩，且日志明确标注这是弱信号推断，避免把网络问题误诊为 overflow 把好窗口改小。

ranxianglei:
Closes #498. Implements the bili-side half of the #499 fix family (paired with acp-kernel PR acp-kernel#196).

## 1. #498 — learn from weak overflow signals

**Problem:** non-4xx overflow evidence (stream truncation without a completion event, upstream HTTP errors on oversized requests) never taught the proxy anything, so an oversized session could fail forever without a strong 400 signal.

**Fix — new `src/weak-overflow.ts` (`noteWeakOverflow`):**
- Fires on: `loop/core.ts` truncation without done-event (both adapters) + `UpstreamHttpError` catch, and both plugin passthrough pipes (chat: `[DONE]`/`message_stop` = terminal; responses: done-family events = terminal).
- Qualifies at input/window ≥ 90% where window = min(learned, effective); 3 events in 15 min confirm.
- On confirm: learns a shrink-only conservative window (per-model map when model known), arms the emergency shrink (`lastInputTokens = max(...)`), marks dirty, logs the decision.
- Strong 4xx learning (existing `!upstream.ok` path) remains the primary teacher; this is the fallback.

## 2. #499 P1a — prefix-affinity persistence

**Problem:** prefix-affinity was pure in-memory (LRU 256, 7-day TTL) → a proxy restart re-forked every anonymous session (#351: pfa-8588 one-shot ingest of 1746 msgs / 458 662 tok, raw 458K resend, cache hit 0%).

**Fix — `src/prefix-affinity.ts` gains `exportSnapshot()`/`importSnapshot()`; new `src/affinity-persist.ts`:** snapshot to `${XDG_STATE_HOME}/billion-context/prefix-affinity.json` (`{version:1, entries}`), atomic tmp+rename, 5 s debounced write after each `note()`, flush in `finishShutdown()`, hydrate on startup. Import validates entry shape, drops TTL-expired chains, re-imposes LRU/item caps.

## 3. #499 P1b — prompt-cache collapse warn (diagnostic)

**Problem:** cache hit 0% on the mirror is a *symptom* of re-forking (or upstream KV drop); nothing surfaced it.

**Fix — new `src/cache-warn.ts` (`warnCacheCollapse`):** after a session has seen ≥50% hit on a ≥8k-token request, 5 consecutive large requests at ≤10% hit log ONE warning pointing at the likely causes (restart re-fork / upstream KV drop) and what to check (upstream session id, `compress.modelContextLimit`). Hooks: `plugin.ts applyUsageSample` + `loop/core.ts recordUsage`. Diagnostics only — no behavior change.

## Tests

20 new tests across 4 files: `tests/weak-overflow.test.ts` (7), `tests/plugin-truncation-weak-overflow.test.ts` (5), `tests/cache-warn.test.ts` (4), `tests/affinity-persist.test.ts` (4).

Pre-flight: `npm run typecheck` ✅ · `npm test` 1018/1020 (2 known permanent env fails in `plugin-agent.test.ts`, unrelated) ✅ · `npm run build` ✅

## Out of scope (deliberate)
- acp-kernel first-sight mass nudge fix (kernel PR #196) — separate repo, human merge → kernel 0.0.51 → separate pin-bump release PR.
- Content-guessed session identity mapping (#286 rule stays).
- `prompt_cache_key` forwarding unchanged.

### DIFF
--- a/src/affinity-persist.ts
+++ b/src/affinity-persist.ts
@@ -1,0 +1,72 @@
+import fs from "node:fs";
+import path from "node:path";
+import { prefixAffinity } from "./prefix-affinity.js";
+import { stateDir } from "./paths.js";
+import { log } from "./logger.js";
+
+/**
+ * #499 P1a: prefix-affinity persistence. The anonymous affinity chains were
+ * pure in-memory (#309), so a proxy restart orphaned every anonymous session:
+ * the next replay forked a fresh session with zero compression state and
+ * resent the raw history (#351 — 458K tokens, 0% cache). The chains are
+ * small (≤256 sessions × ≤128 hashes); persist them to the state dir with a
+ * debounced atomic write and hydrate on boot.
+ */
+
+const PERSIST_DEBOUNCE_MS = 5_000;
+
+function affinityFile(): string {
+    return path.join(stateDir(), "prefix-affinity.json");
+}
+
+let timer: NodeJS.Timeout | null = null;
+let writing = false;
+
+function writeSnapshot(): void {
+    if (writing) return;
+    writing = true;
+    try {
+        const file = affinityFile();
+        const snapshot = { version: 1, entries: prefixAffinity.exportSnapshot() };
+        const tmp = `${file}.tmp`;
+        fs.mkdirSync(path.dirname(file), { recursive: true });
+        fs.writeFileSync(tmp, JSON.stringify(snapshot));
+        fs.renameSync(tmp, file);
+    } catch (e) {
+        log("warn", `[prefix-affinity] persist failed (${e instanceof Error ? e.message : String(e)}); affinity survives in memory`);
+    } finally {
+        writing = false;
+    }
+}
+
+/** Debounced snapshot write — call after every affinity mutation. */
+export function scheduleAffinityPersist(): void {
+    if (timer) clearTimeout(timer);
+    timer = setTimeout(() => {
+        timer = null;
+        writeSnapshot();
+    }, PERSIST_DEBOUNCE_MS);
+    timer.unref?.();
+}
+
+/** Immediate snapshot write — shutdown path. */
+export function flushPrefixAffinity(): void {
+    if (timer) {
+        clearTimeout(timer);
+        timer = null;
+    }
+    writeSnapshot();
+}
+
+/** Load the snapshot a previous process left behind. Call once on boot. */
+export function hydratePrefixAffinity(): void {
+    try {
+        const file = affinityFile();
+        if (!fs.existsSync(file)) return;
+        const parsed = JSON.parse(fs.readFileSync(file, "utf8")) as Record<string, unknown>;
+        const imported = prefixAffinity.importSnapshot(parsed.entries);
+        if (imported > 0) log("info", `[prefix-affinity] hydrated ${imported} chain(s) from ${path.basename(file)} — anonymous sessions reattach across restarts`);
+    } catch (e) {
+        log("warn", `[prefix-affinity] hydrate failed (${e instanceof Error ? e.message : String(e)}); starting with empty affinity`);
+    }
+}
--- a/src/cache-warn.ts
+++ b/src/cache-warn.ts
@@ -1,0 +1,56 @@
+import type { Session } from "./session.js";
+import { log } from "./logger.js";
+
+/**
+ * #499 P1b: upstream prompt-cache collapse warning. A warm session normally
+ * rides the provider's prefix cache (90%+ hit). When a restart or a session
+ * fork makes every request cold, the raw resend is expensive but INVISIBLE:
+ * requests succeed, so no error path fires — the only trace is a hit ratio
+ * that fell off a cliff. This watches each usage sample; once a session has
+ * demonstrated a healthy cache (≥HIGH_HIT on a large input) and then logs
+ * LOW_RUN consecutive large inputs with ~zero hits, it warns ONCE with the
+ * numbers and the likely causes. Diagnostics only — no state is changed.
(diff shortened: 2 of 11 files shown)


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 **PR #504 review** — pre-flight re-run on `2026-09-03_fix-498-weak-overflow` @ `64d605f`. Code matches the description; **one blocking coordination point** (P1a needs #502's guard to actually work end-to-end), plus minor nits below.

## Pre-flight (verified locally)

| Check | Result |
|---|---|
| `npm run typecheck` | ✅ clean |
| `npm test` | 1019/1020 — single failure is the known env-dependent `launcher.test.ts` codex-PATH test (`/usr/bin/codex` vs `codex`; unrelated to this diff). All **20 new tests pass**; `plugin-agent.test.ts` 27/27 here |
| `npm run build` | ✅ self-contained `dist/index.js` |
| E2E | ⚠️ not run here (needs a local Responses upstream). This touches `src/loop/core.ts` + both plugin pipes → per AGENTS.md, please run the 4-phase `e2e-codex` suite before merging |

## Blocking: P1a alone does not fix #499 — needs #502's migration guard

This PR persists the routing table (`prefix-affinity.json`, hydrated at `server.ts:390`), but on **every startup** `migrateLegacyIds()` (`src/persist.ts:211`) still destroys all `pfa-*` session *state* files: anonymous sessions carry label `"prefix-affinity"` ≠ id `pfa-xxxx`, violating the `label === id` invariant checked at `persist.ts:224` → all pfa files get re-keyed into one id and siblings deleted, *before* hydration runs. Net effect after restart: replay resolves to a `pfa-xxxx` id whose persisted blocks no longer exist → fresh empty session → raw resend continues. The exact #351 symptom (#499's motivating event) persists.

The fix already exists in **open PR #502** (skip `pfa-` ids in migration + regression test). Two options:
1. **Fold #502's guard + regression test into this PR**, then close #502 as superseded (avoids two parallel persistence mechanisms — per-session-metadata reseed vs snapshot file).
2. Merge #502 first, then this one.

Your call — flagging because P1a's stated goal ("restart resumes the same compressed context") is unreachable without whichever option ships.

## Minor (non-blocking)

- **`src/loop/core.ts:531`** — the truncation path (`if (!sawDone || truncatedDone)`) has no `signal?.aborted` guard. If the client aborts while the generator is executing *between* yields (inner stream loop breaks at line 279, no pending yield), control falls through to line 531 and records a false weak-overflow signal. Common case is safe (consumer break → generator `.return()` unwinds before reaching 531), but it contradicts the #498 floor-3 design ("客户端主动断开（IDE cancel）不计入"). One-line fix: add `&& !signal?.aborted`. (The plugin pipes do this right via `res.destroyed || res.writableEnded`.)
- **`src/plugin.ts:897`** — inner `if (type === completed || failed || incomplete)` duplicates its outer condition verbatim; harmless redundancy.
- **`resetWeakOverflow(sessionId)`** is exported but only called from tests — never wired to session GC/removal. Bounded leak (≤512 entries) so acceptable for v1, just noting.
- **Multi-instance**: the snapshot file lives under the shared XDG state dir, so two concurrently-running bili instances clobber each other's affinity snapshots. Degrades to today's behavior (no corruption — chains are content-derived); fine for v1, worth a README footnote.

## Verified correct

- Weak-overflow learning writes the **same metadata fields** the strong-signal path uses (`learnedContextLimit` scalar / `learnedContextLimits[model]` map) — read/write sites consistent, so nudge/preflight consume them identically.
- Hooks cover **both compression modes**: proxy mode via `recordUsage` + the re-request `UpstreamHttpError` catch in `loop/core.ts`; plugin mode via both passthrough pipes with proper terminal detection (`[DONE]`/`message_stop`; done-family events). Strong 4xx remains the primary teacher as designed.
- `importSnapshot` validates entry shape, drops TTL-expired chains, re-imposes LRU (256) / item (128) caps; writes are atomic (tmp+rename) with an unref'd 5 s debounce and a flush on shutdown.
- `cache-warn` is diagnostics-only (no state/behavior change), fires exactly once per session, and is hooked in both modes.

**Verdict:** implementation is solid and faithful to the spec. Mergeable once (a) the #502-guard question above is resolved and (b) the e2e suite passes. I can't merge PRs (human-only) — happy to apply the `signal?.aborted` one-liner or fold in #502's guard if you say go.

PR #507, opened 2026-09-04
release v0.1.81

ranxianglei:
## Changes since v0.1.80

- **#502**: stop `migrateLegacyIds` destroying anonymous pfa sessions — anonymous prefix-affinity sessions (label `"prefix-affinity"` ≠ id `pfa-<hash>`) were misclassified as legacy on every boot, deleting all but the newest sibling and orphaning the survivor; prerequisite for #504's affinity-table hydration.
- **#500** (#496): forward-once-then-learn for image-dominated payloads — the default `b64/4` image estimate overestimates pixel-tile upstreams (official Anthropic/OpenAI) up to ~200×, hard-failing multimodal requests with `502 preflight_compress_failed` on estimate alone; now forwarded once so the upstream arbitrates billing, with the existing self-heal learning from any rejection (no 400-loop regression).

Both independently reviewed on the PR threads (incl. combined-tree pre-flight on post-0.1.80 master).

## Pre-flight

- typecheck ✅ · tests 1020/1022 (2 known env fails in `plugin-agent.test.ts`) ✅ · build ✅
- Commit is version-bump only.

PR #508, opened 2026-09-04
feat: index billion-context on the pi.dev gallery

ranxianglei:
## Why

`billion-context` is missing from https://pi.dev/packages while `billion-context-pi` shows up.

Pi's gallery catalogues exactly the npm keyword `pi-package` (pi docs, `packages/coding-agent/docs/packages.md`: "The package gallery displays packages tagged with `pi-package`"). Verified against the live catalog: every listed package (`billion-context-pi`, `context-mode`, `bigpowers`, `pi-mcp-adapter`, `pi-lens`) carries it; `billion-context` (0.1.80) does not.

The detail page https://pi.dev/packages/billion-context does render on demand (it reads the `pi` manifest, so "Types: extension" is correct), but its Downloads row says "not available" — proof it is not in the gallery DB, hence no listing/search entry.

## Change

One keyword in `package.json`: `pi-package`. No code, no version bump.

Legitimate: the package really does ship a pi extension (`"pi": { "extensions": ["./dist/agent/pi.js"] }`, `src/agent/pi.ts`), and the plugin self-disables when there is no proxy (`src/agent/pi.ts:156-157`), so a pi user installing it standalone gets a no-op.

## Pre-flight

- `npm run typecheck` — clean
- `npm test` — 1022 pass / 0 fail
- `npm run build` — not run (metadata-only change, no code touched)

Note: `tests/plugin-agent.test.ts` initially reported 2 failures (`4 == 0` at :273 and :325). Reproduced on clean master, then traced to the harness exporting `BILLION_CONTEXT_PROXY` into the test process — `src/agent/shared.ts:40 proxyBaseFromEnv()` treats it as proxy detection, so the plugin registered 4 tools where the inert tests expect 0. With the var unset the file is 27/27 green. Environment contamination, not a repo defect; flagging it because `npm test` is not hermetic against that var.

## Follow-up

npm keywords are read from the latest published version, so this only takes effect on the next publish. After merge I will open a `*_release-v0.1.81` branch with an isolated `release v0.1.81` commit; CI publishes and pi.dev re-reads the keywords.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,8 @@
     "codex",
     "aider",
     "anthropic",
-    "openai"
+    "openai",
+    "pi-package"
   ],
   "license": "MIT",
   "author": "ranxianglei",


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 Reviewed #508. The one-keyword change itself is correct and legitimate — but **the branch behind this PR does not match its stated scope**. Merging it as-is would ship unrelated, unreviewed code. I've opened a clean, isolated PR for exactly the keyword: **#509**.

**Verified legitimate.** The package really ships a pi extension (`"pi": { "extensions": ["./dist/agent/pi.js"] }`, `src/agent/pi.ts`), and the plugin self-disables without a proxy (`src/agent/pi.ts:156-157`: `if (proxyBase === undefined) return;`), so a standalone install is a safe no-op. Adding `pi-package` is the right call.

**The problem — PR #508's branch bundles 5 extra commits that are not in master.** Its top commit is the clean keyword edit (`package.json` only), but its history also carries:
- `fix: forward-once-then-learn for image-dominated payloads (#496)` → `src/server.ts` + `tests/issue496-forward-once-learn.test.ts`, `tests/fix-488-multimodal.test.ts`
- `docs: BILI_IMAGE_TOKEN_CAP env entries + changelog` → `CHANGELOG.md`, `CONFIGURATION.md`, `CONFIGURATION.zh-CN.md`
- `fix: exclude anonymous pfa sessions from legacy-id migration (#499 prerequisite)` → `src/persist.ts` + `tests/proxy-persist.test.ts`
- two merge commits for the above (#500 / #502)

Confirmed with `git diff origin/master…<branch>` and `git merge-base --is-ancestor`: none of those are reachable from master (`v0.1.80` / `091d4da`). So merging #508 ships #496 + #499 + docs/tests under a "no code" label.

**What I did.** Opened **#509** (`2026-09-04_pi-package-keyword-clean`, off master `v0.1.80`) containing *only* the keyword:
```diff
     "anthropic",
-    "openai"
+    "openai",
+    "pi-package"
```
Pre-flight: `npm run typecheck` clean (exit 0); `npm test` 1019/1020. The single failure — `resolveClientCommand: codex/claude resolve to themselves` at `tests/launcher.test.ts:715` — is **environmental/pre-existing**, not caused by this metadata-only change: it calls `resolveClientCommand("codex", { PATH: "/usr/bin" })` expecting the bare `command: "codex"` but gets the resolved full path because the CLI happens to be present on PATH in the runner. The test is not hermetic against ambient binaries — same class as the `BILLION_CONTEXT_PROXY` non-hermeticity you flagged (that var was proxy detection; here it's the local CLIs). My commit touches only `package.json` keywords, which that test never reads. `npm run build` not run (metadata-only, no code touched).

**Recommendation.** Merge **#509** (or rebase #508 down to just the keyword commit) and close the contaminated one. If #496 / #499 are wanted, land them via their own reviewed PRs — they're preserved on the #508 branch, not lost by switching. The follow-up release is already open as #507 (`release v0.1.81`); the keyword takes effect once that publishes.

PR #514, opened 2026-09-04
fix: break the proxy-origin hijack chain (#405)

ranxianglei:
## What

Fixes #405 — four independent defects in the proxy-origin lifecycle, each with its own fix + regression tests:

1. **Read side had zero liveness** (`src/mcp.ts`) — new async `resolveLiveOrigin()` probes candidates `env → pointer file → default` via `GET /__bili/health`, caches after success, and invalidates the cache on fetch failure (incl. timeout). A dead ephemeral pointer self-heals on the *next* MCP call — no host-client restart. Rebinds are declared on stderr with the instance id.
2. **Write side clobbered unconditionally** (`src/server.ts` + new `src/proxy-origin.ts`) — the listen callback now *claims* instead of clobbers: a different existing origin that is healthy keeps the pointer (logged); otherwise it is reclaimed. Main file format unchanged (plain URL line, old shells keep working); a new sidecar lease (`proxy-origin.lease`: origin/pid/bootedAt/instanceId) carries ownership proof. Graceful shutdown releases the pointer only if it still points at this instance. `/__bili/health` now reports `instanceId`.
3. **Launcher spawned a second instance even when one was healthy** (`src/launcher.ts`) — `ensureProxyRunning` probes the preferred port first and reuses a healthy one (`ProxyHandle.child` now optional; `stopProxy` no-ops on reused handles). Warns if MITM domains were requested on a reused shared instance (domains only apply to instances started with them).
4. **plugin-install froze the transient origin into persistent configs** (`src/plugin-install.ts`) — all four bake sites (claude/codex/opencode) now bake the stable `DEFAULT_PROXY_ORIGIN`; explicit `BILI_MCP_PROXY`/`--origin` still wins. Custom-port instances keep working at runtime because the shell's env→file→default chain finds them (and dead pointers now heal per #1).
5. **Dual-instance snapshot rollback** (`src/persist.ts`, the concrete landing spot for #394's direction) — when two processes share `BILI_SESSIONS_DIR`, a stale full snapshot (monotonic signals `stats.requests` / `state.nextBlockId` regressing) is rejected with a throttled warn and the disk state is kept; `flushSync` on a stale copy reports success so eviction drops the stale in-memory copy (self-heal).

## Acceptance criteria

- [x] kill the launch-spawned ephemeral proxy (:8787 alive) → next MCP call recovers — `tests/mcp-live-origin.test.ts` (self-heal test); cache invalidation covers ECONNREFUSED/timeouts
[...]

PR #516, opened 2026-09-04
fix: skip dead-pid proxy-origin records in MCP shell origin resolution (#405)

ranxianglei:
Split from #514 (fix 1 of 5, the only part still a gap on master — see review there). Completes the read-side (零验活) half of #403 — the piece the instance registry (#419) did not cover. (Note: #514's body cited #405, but that issue is the unrelated ACP_PASSTHROUGH observability bug; no auto-close intended.)

## Problem

`resolveProxyOrigin()` (src/mcp.ts) resolves env → instance file → default with **no liveness check**. When a proxy crashes or is killed, its JSON instance record stays behind naming a port nothing listens on — spawned MCP shells then dial the dead origin forever, and every tool call fails until the host client is restarted (or a new proxy happens to bind that exact port).

## Fix

Rebuilt on the `src/instance.ts` primitives (#394/#403/#417) rather than the HTTP-probe chain from #514:

- A JSON instance record with a live pid (`isPidAlive`, kill-0 — synchronous, no network on the hot path) is trusted as today.
- A record whose pid is dead is stale: skip to the default origin (`http://127.0.0.1:8787`, where a replacement proxy binds) with a **one-time** stderr note suggesting `BILI_MCP_PROXY` if the proxy runs elsewhere. Self-healing: the next tool call re-resolves, so a restarted proxy on the default port is picked up with no shell restart.
- `pid: 0` (record makes no liveness claim, foreign writer) and legacy plain-URL pointers keep the old trust semantics.

Why pid-check instead of #514's `/__bili/health` probe: the JSON record already carries the writer's pid, the check is synchronous (probe would have to become async + cached), and it exactly matches how the launcher attach and plugin-install already gate on the same file.

## Tests

tests/instance-governance.test.ts: dead-pid JSON record → default origin; live pid → record origin; pid 0 → trusted.

## Pre-flight

typecheck ✅ · full suite 1031/1033 (2 known plugin-agent env fails, unchanged baseline) · build ✅

### DIFF
--- a/src/mcp.ts
+++ b/src/mcp.ts
@@ -9,7 +9,7 @@
 import fs from "node:fs";
 import path from "node:path";
 import { fileURLToPath } from "node:url";
-import { readProxyInstanceFile } from "./instance.js";
+import { isPidAlive, isProxyInstanceFile, readProxyInstanceFile, type ProxyInstanceFile } from "./instance.js";
 
 const VERSION = (() => {
     try {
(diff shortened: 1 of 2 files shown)


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 Done — PR: https://github.com/ranxianglei/billion-context/pull/522

**Triage verdict**

- **Duplicate screening**: no duplicate PRs/issues. This is the rescoped split of closed #514 (fix 1 of 5) per your review there; fix 5 lives in sibling #517. No overlap found.
- **Verified the gap on master**: `resolveProxyOrigin()` in `src/mcp.ts` resolves env → instance file → default with no liveness check, while both sibling readers gate on the same file's pid — `proxyOriginForInstall()` in `src/plugin-install.ts` throws when `isPidAlive(inst.pid)` fails, and the launcher attach path does the same. So a dead proxy leaves a JSON record naming a dead port and spawned MCP shells dial it forever until the host restarts. Gap confirmed real, not a symptom-level rephrase of something else.
- **Design**: implemented exactly as scoped in the OP (per your #514 review): live pid → trusted; dead pid → skip to `http://127.0.0.1:8787` + one-time stderr note mentioning `BILI_MCP_PROXY`; `pid: 0` and legacy plain-URL pointers keep old trust semantics (deliberate asymmetry vs plugin-install, which refuses them); env still wins. No HTTP probe — the record already carries the writer's pid, the check is synchronous, and it matches how launcher attach / plugin-install already gate.
- **Self-healing verified**: origin is re-resolved on every call (no cache), so a restarted proxy writing a fresh record is picked up on the next tool call without a shell restart.

**Pre-flight**

- typecheck ✅
- full suite 947/948 — sole failure is the known env-dependent `resolveClientCommand: codex/claude resolve to themselves` test (this machine has `/usr/bin/codex`); identical on clean master (baseline 946/947)
- build ✅
- manual E2E on built `dist/mcp.js` against a stale record (dead pid, unrouted port): one-time note fired, and the tool-call error now names the fallback origin instead of the dead port

**Notes (accepted limitations, report-only)**

- PID reuse: if the dead pid is later reused by an unrelated process, the stale record looks live again. Same exposure as the existing plugin-install/launcher gating.
- Shells spawned with an explicit `BILI_MCP_PROXY` env pointing at a dead port still dial it — env is operator intent by design; recovery = repoint/reinstall.

No version bump, no lockfile change. Merge is yours — I can't merge PRs.

ranxianglei:
Closing as duplicate of #522 — same fix (skip dead-pid proxy-origin records in MCP shell origin resolution), #522 is the rebased/current version (updated 2026-09-06). Tracking continues there.

PR #522, opened 2026-09-04
fix: skip dead-pid proxy-origin records in MCP shell origin resolution (#405)

### LINKED ISSUE #405: ACP_PASSTHROUGH 一键静默关死全部压缩:无日志/无横幅/面板不可见且不可清除,同族 env 旋钮全部生效于无声
## 症状

`ACP_PASSTHROUGH=1`(环境变量)或配置文件 `passthrough: true` 会**静默关闭 100% 压缩**:无启动日志、无横幅提示、web 面板零显示、`/__bili/config` GET 不含该键、PUT 甚至**不接受该键**(面板无法查看也无法解除),launcher 还会把它透传给子代理(`proxyStartArgs`)。用户以为 bili 在工作(代理活着、面板绿、请求正常转发),实际一个 token 都没省——排查方向还被"看起来健康"的面板引偏。

同族静默旋钮:`ACP_MODEL_CONTEXT_LIMIT`、`ACP_COMPRESS_TOOL=0`/`ACP_NO_INJECT_TOOL`、`BILI_PERSIST=0`——全部生效于无声。

## 根因
`loadOptions`(config.ts:359,dist 47768)读 env/file → `server.ts:881`(dist 56045)`if (!opts.passthrough …)` 直接跳过压缩分支;listen 横幅(server.ts:370,dist 55721)逐项打印 MITM/overrides/debug 配置**独缺 passthrough**;`handleConfigGet`(dist 50995-51006)字段集不含它。

## 建议修复
- 有效 passthrough=true 时:启动打一行 warn(`PASSTHROUGH — compression OFF`)+ 横幅追加状态 + 面板红角标;
- `/__bili/config` GET 增补 `{passthrough, passthroughSource}`、PUT 允许清除;
- 所有 env 级静默覆盖在 boot 时逐条打印一次(值+来源)。

## 验收标准
- [ ] `ACP_PASSTHROUGH=1 bili start`:日志有 warn、面板显示 OFF;去掉即恢复
- [ ] 面板可读写 passthrough 状态
- [ ] 任一静默 env 生效时 boot 有痕

### LINKED ISSUE #599: MCP shell origin resolution must skip dead-pid proxy-origin records
### Problem
Proxy-origin records live in the instance registry. A stale record whose owning pid is dead can still be adopted when the MCP shell resolves its origin, pointing tool injection at a dead proxy port.

### Impact
MCP plugin spawn-time injection fails or attaches to the wrong instance — residual of the #403/#405 proxy-origin hijack chain after #514/#527's fixes.

### Suggested fix
Liveness-check (pid alive) before adopting an origin record in MCP shell origin resolution. PR #522 carries the fix (+49/−2, merges clean).

ranxianglei:
Closes #599 — tracking issue for this PR.

Split from #514 (fix 1 of 5, the only part still a gap on master — see review there). Fixes #405 (MCP half).

## What

`resolveProxyOrigin()` (src/mcp.ts) resolved env → instance file → default origin with **no liveness check**. When a proxy crashes or is killed, its JSON instance record stays behind naming a port nothing listens on — spawned MCP shells then dial the dead origin forever, and every tool call fails until the host client is restarted (or a new proxy happens to bind that exact port).

Rebuilt on the `src/instance.ts` primitives (#394/#403/#417) rather than the HTTP-probe chain from #514:

- A JSON instance record with a **live** pid (`isPidAlive`, kill(0) — synchronous, no network on the hot path) is trusted as today.
- A record whose pid is **dead** is stale: skip to the default origin (`http://127.0.0.1:8787`, where a replacement proxy binds) with a **one-time** stderr note suggesting `BILI_MCP_PROXY` if the proxy runs elsewhere. Self-healing: the next tool call re-resolves, so a restarted proxy is picked up with no shell restart.
- `pid: 0` (record makes no liveness claim, foreign writer) and legacy plain-URL pointers keep the old trust semantics — deliberate asymmetry vs plugin-install, which refuses them.
- `BILI_MCP_PROXY` env still wins over everything (operator intent).

Why pid-check instead of #514's `/__bili/health` probe: the JSON record already carries the writer's pid, the check is synchronous (probe would have to become async + cached), and it exactly matches how the launcher attach and plugin-install already gate on the same file.

## Tests

tests/instance-governance.test.ts: dead-pid JSON record → default origin (re-resolved on every call); live pid → record origin; pid 0 → trusted; stderr note fires once and mentions BILI_MCP_PROXY.

## Pre-flight

- typecheck ✅
- full suite 947/948 — the single failure is the known env-dependent `resolveClientCommand: codex/claude resolve to themselves` test (machine has /usr/bin/codex), identical on clean master (baseline 946/947)
- build ✅ (dist/mcp.js contains the new behavior)
- manual E2E on built dist/mcp.js: stale record (dead pid, unrouted port) → one-time stderr note → tool-call error names the fallback origin http://127.0.0.1:8787, not the dead port

## Notes

- Known limitation (accepted): PID reuse — if the dead pid is later reused by an unrelated process, the stale record looks live again. Same exposure as the existing plugin-install/launcher gating.
- Shells spawned with an explicit `BILI_MCP_PROXY` env pointing at a dead port still dial it — env is operator intent by design; recovery = repoint/reinstall.
- No version bump, no lockfile change.

### DIFF
--- a/src/mcp.ts
+++ b/src/mcp.ts
@@ -9,7 +9,7 @@
 import fs from "node:fs";
 import path from "node:path";
 import { fileURLToPath } from "node:url";
-import { readProxyInstanceFile } from "./instance.js";
+import { isPidAlive, isProxyInstanceFile, readProxyInstanceFile } from "./instance.js";
 
 const VERSION = (() => {
     try {
@@ -33,12 +33,27 @@ type McpToolDef = {
 // BILI_CONVERSATION_ID (launcher-spawned hosts like codex) has no matching
 // request id — binding is headless (next NEW session).
 const DEFAULT_PROXY_ORIGIN = "http://127.0.0.1:8787";
+let warnedStaleRecord = false;
 
 export function resolveProxyOrigin(): string {
     const fromEnv = process.env.BILI_MCP_PROXY?.trim();
     if (fromEnv && fromEnv.length > 0) return fromEnv;
     const discovered = readProxyInstanceFile();
-    if (discovered && /^https?:\/\/\S+$/.test(discovered.origin)) return discovered.origin;
+    if (discovered && /^https?:\/\/\S+$/.test(discovered.origin)) {
+        // #405: a JSON record naming a dead writer pid points at a port nothing
+        // listens on — skip it for the default origin instead of dialing it
+        // forever. The `pid > 0` guard is deliberate: pid 0 (foreign writer)
+        // and legacy plain-URL pointers make no liveness claim and keep the
+        // old trust semantics (unlike plugin-install, which refuses them).
+        if (isProxyInstanceFile(discovered) && discovered.pid > 0 && !isPidAlive(discovered.pid)) {
+            if (!warnedStaleRecord) {
+                warnedStaleRecord = true;
+                process.stderr.write(`[bili-mcp] recorded bili proxy (${discovered.origin}, pid ${discovered.pid}) is not running — falling back to ${DEFAULT_PROXY_ORIGIN}; set BILI_MCP_PROXY if your proxy listens elsewhere\n`);
+            }
+            return DEFAULT_PROXY_ORIGIN;
+        }
+        return discovered.origin;
+    }
     return DEFAULT_PROXY_ORIGIN;
 }
 
--- a/tests/instance-governance.test.ts
+++ b/tests/instance-governance.test.ts
@@ -110,6 +110,38 @@ test("resolveProxyOrigin: env wins, JSON file parsed, default fallback", () => {
     }
 });
 
+test("resolveProxyOrigin: dead-pid record falls back to default origin, live pid and pid 0 trusted (#405)", () => {
+    const st = tmpStateDir();
+    const prevEnv = process.env.BILI_MCP_PROXY;
+    const origStderrWrite = process.stderr.write;
+    const stderrOut: string[] = [];
+    process.stderr.write = ((chunk: string | Uint8Array) => {
+        stderrOut.push(String(chunk));
+        return true;
+    }) as typeof process.stderr.write;
+    try {
+        delete process.env.BILI_MCP_PROXY;
+
+        atomicWriteInstanceFile(sampleInstance({ origin: "http://127.0.0.1:4242", pid: deadPid() }));
+        assert.equal(resolveProxyOrigin(), "http://127.0.0.1:8787", "dead-pid record skipped");
+        assert.equal(resolveProxyOrigin(), "http://127.0.0.1:8787", "re-resolves on every call");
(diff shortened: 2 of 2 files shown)


ranxianglei:
**Review vs master 3cd85b6:** ✅ merge-ready.

- `src/mcp.ts` `resolveProxyOrigin()` is unchanged on master since the PR — applies clean.
- Semantics verified: `isProxyInstanceFile(discovered) && discovered.pid > 0 && !isPidAlive(pid)` → one-time stderr note + fallback to `DEFAULT_PROXY_ORIGIN` (8787). `pid 0` (foreign writer / legacy plain URL) keeps old trust semantics — deliberate and documented.
- Test covers all three cases (dead → fallback + one-time note, live → trusted, pid 0 → no liveness claim).

Behavior question (non-blocking): when the dead record is skipped we dial 8787 by default; if the user runs the proxy on a custom port this still fails — but that was equally true pre-PR (dialing a dead origin), and the stderr note now tells them to set `BILI_MCP_PROXY`. Strictly better than before.

(tracking: #599)

ranxianglei:
> 🤖 ework agent · qwen3.8-27b

Confirmed your review independently against **current** master (`1330fc5`). One thing worth adding beyond it:

Your check covered `src/mcp.ts` being unchanged on master (true — `resolveProxyOrigin()` is untouched since the branch point). But the PR was pre-flighted off `09ef208`, and master has moved under it since: **#527** reworked `src/instance.ts` (the module this PR now imports from) *and* both #527 and this PR touch `tests/instance-governance.test.ts`. So "applies clean" wasn't guaranteed by the `mcp.ts` check alone. Re-verified against current master:

- Dry-run 3-way merge of the PR into `master`: **clean**, no textual conflict (the shared test file auto-merges — the two edits sit in different regions).
- The three imported primitives — `readProxyInstanceFile`, `isProxyInstanceFile`, `isPidAlive` — are intact on current master with unchanged signatures. #527 only rebuilt the cross-instance *registry* (per-instance marker files); it did not touch the proxy-origin record functions this change relies on.
- Built the actual merged tree (`master` + PR) and ran: `typecheck` ✅; the full `tests/instance-governance.test.ts` **11/11 pass** (exercises the new test *and* #527's registry rework together); the new `#405` test in isolation ✅.

So it's merge-ready against *current* master, not just the pre-#527 base. (I ran typecheck + the directly-affected governance file, not the whole ~948-test suite — CI re-runs that on merge.)

Behavior question — agreed, no action needed: the custom-port case was equally broken pre-PR (dialing a dead origin), and now at least emits an actionable `BILI_MCP_PROXY` hint. Strictly better than before.

Next step is yours to take — I can't merge (human-only per AGENTS.md). It's a content PR (no version bump, no lockfile), so on merge it lands on master and ships with the next release.

PR #528, opened 2026-09-04
fix: instance registry lost-update — per-instance marker files (#527)

### LINKED ISSUE #527: instance registry lost-update: shutdown unregister can wipe a concurrently-starting proxy's registration
## Repro (observed live)

1. Proxy A running (large sessions dir → graceful flush takes seconds).
2. Restart: SIGTERM to A, immediately start proxy B (this is the normal `bili start`/launcher restart cadence).
3. A's `finishShutdown` runs `unregisterInstance(A)` **after** its async `flushAllSessions()` completes — which can be *after* B's `registerInstanceAndWarn` already wrote.
4. A's read-modify-write is not cross-process safe: A read the pre-B list, B writes `[B]`, A writes back its stale computed list `[…]` → **B's entry is wiped**.

Observed at the 2026-09-04 13:45 restart: pid 2212910 (port 8787) alive ever since, but `instances.json` is `{"instances":[]}` and `proxy-origin` absent — found while verifying #526.

## Impact

- Registry-based discovery (#394/#417) silently degrades: no dual-writer warnings for the running instance, MCP shells fall back to the default origin.
- `clearProxyInstanceFile` is ownership-guarded (only unlinks when `instanceId` matches), but `unregisterInstance` rewrites the whole registry file and also prunes-by-write, so it can clobber.

## Suggested fix

Guard the whole read-modify-write with a cross-process lock — e.g. a lock directory (`fs.mkdirSync(lockDir)` is atomic on POSIX/Win32; retry with backoff, stale-lock break by pid check), acquired by both `registerInstanceAndWarn` and `unregisterInstance`. Alternative: per-instance marker files (registry = directory listing), making removal an unlink of exactly one file — no shared rewrites at all.

Found while verifying #526; report originally in #446 thread (comment 5537296486).

ranxianglei:
Fixes #527.

## What & why

The multi-instance liveness registry (`instances.json`, #394/#417) was an unsynchronized, non-atomic read-modify-write. On the normal `bili start`/launcher restart cadence — SIGTERM A while A is still flushing sessions, immediately start B — A's `finishShutdown` runs `unregisterInstance(A)` *after* its async `flushAllSessions()` completes, which can land *after* B's `registerInstanceAndWarn` already wrote. A's stale computed list then overwrites B's entry, wiping B's registration while B stays alive. Observed live at the 2026-09-04 13:45 restart (pid 2212910 / port 8787 alive, `instances.json` = `{"instances":[]}`).

## Fix

Replace the shared file with **one marker file per instance** under `<state>/instances/<id>.json`. The registry IS the directory listing, so:

- `unregisterInstance(id)` unlinks exactly **one** file (ownership-scoped by filename), so a stale unregister can no longer clobber a concurrently-starting instance. This removes the defect class rather than papering over it with a lock.
- `registerInstanceAndWarn` writes only its own marker atomically (tmp+fsync+rename), warns on each live non-self entry (unchanged #394 warning text), and opportunistically reaps dead owners' markers (liveness-gated; never touches live/self entries).
- Legacy `instances.json` is folded in **read-only** for one-restart continuity and is never written back.

`registerInstanceAndWarn` / `unregisterInstance` signatures are unchanged; `src/server.ts` is untouched. `proxy-origin` (single-instance discovery) was already ownership-guarded and is unaffected.

## Trade-off considered

The issue suggested a cross-process lock directory as an alternative. Chose per-instance marker files because it eliminates the shared rewrite entirely — no lock lifecycle, no stale-lock break-by-pid (unsafe on Linux due to pid reuse) — and matches the file's existing atomic-write idiom.

## Tests / pre-flight

- `tests/instance-governance.test.ts`: registry test now asserts against the marker dir (`a.json`, `b.json` → unregister `a` → `b.json`); added a legacy-fold read-only test (a live legacy entry triggers the warning and the legacy file is left unrewritten).
- `npm run typecheck` pass, `npm run build` pass, full `npm test` = 947 pass / 1 fail.
- The 1 failure (`resolveClientCommand: codex/claude resolve to themselves`) is **pre-existing and environment-dependent**: this sandbox has `/usr/bin/codex` and `/usr/bin/claude` installed (needed by the e2e suite), so the bare-name assertion fails. Verified it fails identically on the unmodified tree (my changes stashed). Unrelated to this change.

### DIFF
--- a/src/instance.ts
+++ b/src/instance.ts
@@ -1,4 +1,4 @@
-import { randomUUID } from "node:crypto";
+import { createHash, randomUUID } from "node:crypto";
 import fs from "node:fs";
 import path from "node:path";
 import { stateDir } from "./paths.js";
@@ -71,15 +71,15 @@ export function instanceFilePath(): string {
 }
 
 /** tmp+fsync+rename (same shape as web/api.ts atomicWriteConfig) — a torn
- *  write must never leave a half-origin behind (#406 family). */
-export function atomicWriteInstanceFile(info: ProxyInstanceFile, file?: string): void {
-    const filePath = file ?? instanceFilePath();
+ *  write must never leave a half-file behind (#406 family). Shared by the
+ *  proxy-origin record and every registry marker. */
+function atomicWriteJson(obj: unknown, filePath: string): void {
     fs.mkdirSync(path.dirname(filePath), { recursive: true });
     const tempPath = `${filePath}.${process.pid}.${randomUUID()}.tmp`;
     let descriptor: number | undefined;
     try {
         descriptor = fs.openSync(tempPath, "wx", 0o644);
-        fs.writeSync(descriptor, JSON.stringify(info) + "\n", null, "utf8");
+        fs.writeSync(descriptor, JSON.stringify(obj) + "\n", null, "utf8");
         fs.fsyncSync(descriptor);
         fs.closeSync(descriptor);
         descriptor = undefined;
@@ -95,6 +95,10 @@ export function atomicWriteInstanceFile(info: ProxyInstanceFile, file?: string):
         } catch {}
         throw error;
     }
+}
+
+export function atomicWriteInstanceFile(info: ProxyInstanceFile, file?: string): void {
+    atomicWriteJson(info, file ?? instanceFilePath());
 }
 
 /** Remove our record on shutdown — but only if the file still carries OUR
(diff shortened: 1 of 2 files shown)


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

**Verdict: reviewed, independently verified, merge-ready.** 👍

**Independent pre-flight** (checked out PR head `0e8a9e9`, ran locally):
- `npm run typecheck` → pass
- `npm run build` → pass
- `npm test` → **948 tests: 947 pass / 1 fail** (matches your report)
- The 1 failure is `resolveClientCommand: codex/claude resolve to themselves` (`tests/launcher.test.ts:716`): it expects the bare name but resolves to the absolute binary because this environment has the codex/claude binaries on PATH. I reverted the PR's two files back to clean master and re-ran that file → **same single failure**. Confirmed pre-existing + environment-dependent, unrelated to this change.

**Design — correct and well-motivated:**
- Traced the race to `src/server.ts`: `shutdown()` → `server.close(cb)` → `flushAllSessions().finally(finishShutdown)` (server.ts:480-483) → `unregisterInstance(instanceId)` (server.ts:467). Because `flushAllSessions()` is async, A's unregister can land *after* B's register. In the old shared-file code, A's read→compute→write are two separate sync fs calls with **no cross-process atomicity**, so A's stale snapshot (read before B wrote) followed by A's later write clobbers B → the observed `{"instances":[]}`.
- Per-instance markers remove the shared rewrite entirely: `unregisterInstance` unlinks exactly one filename-scoped file, so it structurally cannot touch another instance. That kills the defect class rather than papering over it with a lock. Agree markers > lock dir: no lock lifecycle, no stale-lock break-by-pid (unsafe on Linux from pid reuse), and it reuses the existing tmp+fsync+rename idiom.
- Reaping is liveness-gated and skips self/live entries — worst case a stale marker lingers, never a live entry deleted. Safe direction.
- Legacy `instances.json` is folded in read-only and never written back → gives one-restart continuity plus cross-version awareness (new sees old live instances) during rollout. Verified no other code reads the `{instances:[…]}` shape (only `src/instance.ts`), so the migration is self-contained.
- Scope holds: signatures unchanged, `src/server.ts` untouched, `proxy-origin` behaviorally unchanged (`atomicWriteInstanceFile` is now a thin wrapper over the extracted `atomicWriteJson`; `clearProxyInstanceFile` still instanceId-guarded).
- Tests: registry test now asserts the marker dir (`a.json`,`b.json` → unregister `a` → `b.json`); the new legacy-fold test covers warn-on-live-legacy + file-left-unrewritten. Good coverage.

**Minor, non-blocking notes** (no action needed to merge):
1. The legacy `instances.json` is never GC'd — if an old-code instance crashes without a clean shutdown, its entry lingers in that one file indefinitely. Bounded (single file), filtered by liveness so it triggers no spurious warning, and explicitly best-effort; fine as-is.
2. A stale marker whose pid later gets reused can yield a transient false-positive "another bili instance is running" warning. This is inherent to the shared `isPidAlive` model (the old code had the same property), not something this PR introduces — and it's precisely why you avoided lock-break-by-pid. Worth a footnote, not a blocker.

Not merging — that's human-only. Ready for you to merge whenever.

PR #479, opened 2026-09-03
fix: bound session storage growth — single-copy leaf blocks, boot GC, one-pass startup (#478)

ranxianglei:
Closing as superseded by #530. Scope change decided in #401: the retention/size GC half is dropped (sessions are permanent — a year-long session must never be lost); #530 keeps the single-copy leaf-block dedup, the one-pass boot, and the once-ever migration marker, and carries this PR's rebase fixes. The ework-agent work here (and in #478) was the correct diagnosis and the base for #530 — thank you.

PR #512, opened 2026-09-04
docs: clarify billion-context / billion-context-pi / opencode-acp relationship (#511)

ranxianglei:
Closes #511 — 更新 README 介绍，写明白 billion-context / billion-context-pi / opencode-acp 三者的关系、什么时候选哪个，以及 omp 怎么选。

## What changed

Added a new **"Which do I need?"** section to both `README.md` and `README.zh-CN.md` (kept in sync):

- A table + diagram showing how the three tools relate:
  - **billion-context (`bili`)** — the universal proxy + built-in client-side plugins (`dist/agent/{pi,omp,opencode}.js`). The one you install.
  - **billion-context-pi** — the original standalone pi extension (0.1.x reference impl). Its pi plugin has since moved *into* `billion-context`; the old package self-disables behind `bili`. Superseded — don't install it separately.
  - **opencode-acp** — standalone opencode-only native compression, no proxy needed; self-disables when run behind `bili`.
- "When to pick which" guidance:
  - Default → `billion-context` (any client via `/bili/` or `bili <client>`).
  - pi / omp → `billion-context`'s built-in plugin; do NOT install legacy `billion-context-pi`.
  - **omp specifically** → rides `billion-context`'s own `dist/agent/omp.js` over a persistent overlay home (real config untouched); omp goes through `billion-context`, not `billion-context-pi`.
  - opencode → standalone `opencode-acp` (no proxy) OR `bili opencode` (built-in plugin; host `opencode-acp` self-disables).
- Updated the outdated Status-section line that still called `billion-context-pi` "the reference implementation".

## Why these facts (verified against source)

- `src/plugin-install.ts:101` — "the legacy billion-context-pi package (0.1.x … before the plugin moved into billion-context itself)".
- `src/plugin-install.ts:116` (`isBiliPiEntry`) — legacy `billion-context-pi` entries deliberately excluded; they self-disable under `BILLION_CONTEXT_PROXY`.
- `src/launcher.ts:1886-1902` — omp is pi-based, uses `dist/agent/omp.js`, overlay home `<ompHome>-bili`, real config untouched.
- `src/launcher.ts:1903-1912` + `CONFIGURATION.md:558` — `opencode-acp` self-disables via `BILLION_CONTEXT_PROXY`.

## Pre-flight

Docs only (two markdown files); no source changes, so typecheck/test/build are unaffected. No version bump on this content branch (AGENTS.md §4).

PR #530, opened 2026-09-04
fix: single-copy leaf blocks + one-pass session boot (#401)

ranxianglei:
Supersedes #479 (and #478) with the scope agreed in the #401 thread: **sessions are permanent — the GC half of #479 is dropped entirely** (design goal: a session is used for a year and must never be lost); what remains is the storage-growth halving and the boot-cost fix.

## What's in

1. **Single-copy leaf blocks** — `blockContents` caches two rendered views (`one` one-level / `full` all-originals); on leaf blocks they are byte-identical, which was 50% of the field's bytes. Leaves now persist as `one: null` + `full`; legacy duplicate pairs normalize to the single-copy form on load, so every file shrinks by its leaf-duplicate half the next time it is written.
2. **One-pass boot** — `boot()` = ONE `loadAll` walk+parse, then the #286 identity migration over the same parsed map. The old `migrateLegacyIds()+loadAll()` pair walked and parsed the whole tree twice per start.
3. **Once-ever migration marker** — `.bili-migration-286.done` dotfile (invisible to the kernel's `.json` walk) written after the first migration pass, so it never re-scans/re-logs on later boots.
4. **No deletion, ever** — no retention window, no size budget, no env knobs. Boot on the real corpus below deleted 0 files.

## Real-corpus validation (copy of the live sessions dir: 453 files / 240.2MB)

| metric | before | after |
|---|---|---|
| boot | ~1.8s (2x walk+parse) | **814ms** (single pass) |
| leaf blocks one/full byte-identical | 518/518 (duplicated) | 518/518 collapsed on load |
| leaf-duplicate bytes reclaimable on rewrite | — | **38.1MB** (50% of blockContents, 16% of corpus) |
| files deleted at boot | n/a | **0** (permanent) |

## What is deliberately NOT here (see #401 comment for numbers)

The 63% storage hog is `messages` (lastMessages full-conversation snapshot, 152MB of the 240MB corpus, sole consumer `bili export`). Fixing it needs a design call (gzip field / snapshot-on-export / accept) and possibly kernel cooperation — tracked in the issue, not stuffed into this PR. Lazy boot (stat-only walk) is likewise future work.

## Preflight

- typecheck clean
- 1038 tests / 1036 pass / 2 known env fails (`plugin-agent.test.ts`, pre-existing on master)
- build ok
- 4 new tests in `tests/fix-401-storage.test.ts` (single-copy round-trip, legacy-pair normalize, one-walk boot that deletes nothing, once-ever marker)

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -634,6 +634,10 @@ Compression state (blocks, summaries, original message cache) lives **in the pro
 - If you point the client back at the real upstream (or stop the proxy), the client replays its **full local history** every turn. After a long compressed session this can exceed the model's context window (`context_window_exceeded`).
 - There is no way to "unpack" a compression block into the client's local history — the client never saw the compressed form.
 
+### Storage lifecycle at boot (#401)
+
+The proxy boots session storage with a **single** directory walk+parse (load and the one-time #286 identity migration run over the same parsed map). Sessions are kept permanently — the design goal is that a year-long conversation is never lost — so there is no retention or size-budget pruning. The #286 migration writes a `.bili-migration-286.done` marker after its first pass, so it never re-scans or re-logs on later boots.
+
 ### Migrating off the proxy
 
 Export the session and paste it into a fresh conversation as a handoff:
--- a/src/decompress-shared.ts
+++ b/src/decompress-shared.ts
@@ -91,7 +91,9 @@ export function resolveDecompress(
         // Honor the full flag: `one` = direct msgs + nested child summaries,
         // `full` = all original messages. Returning the cached full text
         // unconditionally would break the default one-level semantics.
-        const view = full ? cached.full : cached.one;
+        // `one === null` means the two views were byte-identical at cache
+        // time and deduped to one copy (#401).
+        const view = full ? cached.full : (cached.one ?? cached.full);
         body = view.text;
         count = view.count;
     } else {
--- a/src/persist.ts
+++ b/src/persist.ts
@@ -1,11 +1,12 @@
 import { createHash } from "node:crypto";
+import { existsSync, mkdirSync, writeFileSync } from "node:fs";
 import { rm } from "node:fs/promises";
 import * as path from "node:path";
 import { StateStore, flatFileNameFor, type PersistedEnvelope } from "acp-kernel/persist";
 import { sessionsDir } from "./paths.js";
 import { log as loggerLog } from "./logger.js";
 import { createInitialState, type CompressionState, type CoreMessage } from "acp-kernel";
-import type { Session, BlockContent } from "./session.js";
+import type { Session, BlockContent, BlockView } from "./session.js";
 
 /**
  * On-disk persistence for proxy sessions.
@@ -59,6 +60,9 @@ import type { Session, BlockContent } from "./session.js";
  */
 
 const PERSIST_VERSION = 3;
+/** Dotfile (invisible to the kernel's .json walk): written after the first
+ *  successful #286 migration pass so later boots skip the scan entirely. */
+const MIGRATION_MARKER = ".bili-migration-286.done";
 
 interface PersistedSession {
     version: number;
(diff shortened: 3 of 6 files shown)


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

Reviewed #530 (branch `2026-09-04_sessions-storage-slim`, core commit `52faa04`). **Verdict: LGTM — approve, with 2 non-blocking notes.** The core change is correct; I reproduced preflight myself rather than taking it on faith.

**Verified firsthand**
- `npm run typecheck` → clean.
- `npm run build` → success (`dist/index.js` 2.63 MB, acp-kernel inlined).
- Full suite: 1038 tests. My run: **1037 pass / 1 fail**. That one failure is *not* this PR's (note 2 below). Your "2 env fails (`plugin-agent`)" passed here; the count differs only by which CLIs happen to be on PATH per machine.
- The 4 new tests in `tests/fix-401-storage.test.ts` all pass and genuinely exercise each claim (single-copy round-trip, legacy-pair normalize, single-walk boot that deletes nothing, once-ever marker).

**Core correctness (I read the code, not just the diff)**
- *Single-copy leaf blocks* — `BlockContent.one` is now `BlockView | null` (`src/session.ts:13`). The writer sets `one: sameView ? null : {...}` where `sameView = one.text===full.text && one.count===full.count` (`src/stream.ts:95`); load normalizes legacy dup pairs with the same predicate (`src/persist.ts:417-422`). I traced every consumer of `.one`: `src/decompress-shared.ts:96` uses `(cached.one ?? cached.full)`, `src/export.ts:99` reads only `.full`, `orphan-gc.ts` only deletes keys. **No null-dereference path exists**, and decompress output stays byte-identical to before (null only ever replaces a copy that equaled `full`). Correct in both plugin mode and proxy mode — it's a pure storage optimization, no wire change.
- *One-pass boot* — `boot()` (`src/persist.ts:211`) does one `loadAll`, then `applyLegacyMigration(loaded)` mutates that same map (no second walk), then hydrates. Test 3 asserts exactly N fs reads for N files.
- *Idempotent migration* — the re-run hazard I probed for is handled: `if (label === id) { claimed.add(id); continue; }` (`src/persist.ts:268`) skips already-rekeyed files instead of deleting them, and the `claimed.has(label)` branch (`:286`) drops a legacy file that collides with a surviving current-format session. Marker write is try/catch'd for read-only dirs.
- *Once-ever marker* — `.bili-migration-286.done` dotfile, gated inside `applyLegacyMigration` (`:250`) so both `boot()` and the retained `migrateLegacyIds()` respect it; invisible to the kernel's `.json` walk.
- *No deletion* — confirmed no retention window / size budget / env knob added. Only the pre-existing #286 dedup remains (self-terminating, guarded). "0 files deleted" is corpus-specific (no legacy dupes present); the general guarantee is "no new GC," which holds.

**Notes (non-blocking)**
1. **This branch bundles two unrelated, already-reviewed side-PRs beyond the #401 scope**: `#526`/`#446` (launcher ephemeral port — `src/launcher.ts`, `src/cli.ts` help text, `tests/launcher.test.ts`) and `#512` (README tool-relationship docs — `README.md`, `README.zh-CN.md`). Neither is on master yet, so merging #530 will land them too. Not wrong, but it couples merge order. Recommend landing #526 and #512 independently first and rebasing #530 onto them — or confirming you intend the bundle.
2. **Pre-existing, env-dependent test failure (not introduced here)**: `tests/launcher.test.ts:715` "resolveClientCommand: codex/claude resolve to themselves" hardcodes `{command:'codex'}` and fails whenever `codex` is on PATH. I reproduced it against **master's code in this same environment** (identical `not ok`) — `/usr/bin/codex` is installed here. Suggest making it PATH-independent (assert `endsWith('codex')`, or pin a controlled PATH inside the test). Reporting only per triage rules since it predates this PR.

Compliance: no `package.json`/version or dependency changes in the diff; acp-kernel still pinned to exact `0.0.51`.

I did not independently reproduce the real-corpus MB figures (no live sessions dir in this sandbox), but the mechanism behind each number is verified above and the percentages are internally consistent. Happy to help split the bundle if you want note 1 addressed before merge.

PR #534, opened 2026-09-04
fix: status panel counts outbound system prompt + tool schema (#532)

ranxianglei:
## Problem

The ACP Context Analysis panel's `Sent to LLM (after compression, est.)` total and `Token Breakdown (sent view)` systematically undercounted the host system prompt and tool schemas: `handlePluginStatus` passed a hardcoded `systemPromptTokens: 0` into `buildStatusPanel`, while the kernel panel computes `sentTotal = classified + systemPromptTokens` and skips rows with value <= 0 — so the `SysPrompt` row never rendered and large system+tools payloads were invisible (a 62k/35k session showed Sent=181/219).

The quantity was already measured elsewhere in bili (`estimateInputTokens` in `src/server.ts`) but never handed to the panel.

## Fix

- `src/server.ts`: new exported `countSystemAndToolsTokens(systemText, tools)` — same counting as `estimateInputTokens`, which now reuses it (behavior unchanged, existing backstop tests still pass). Each prepare path stores the measured **outbound** system+tools tokens on `session.metadata.systemPromptTokens`:
  - **anthropic**: the final `injectSystem` result (includes the compress prompt when injected)
  - **openai**: the hoisted client system + compress system prompt (`sysParts`), skipped for title-gen side requests (their tiny own system must not clobber the main session value)
  - **responses**: the developer message content, only when the transform succeeded
- `src/plugin.ts`: passes that value into `buildStatusPanel` instead of `0` (with a numeric guard), and adds the same amount to `unprunedTokens` so the `Session-only (compressed originals)` derivation stays on one scale.

### Why this shape

- The adapters hoist the head system out of the message projection (`openaiToCore` → `{msgs, systemText}`, Responses → `systemParts`), so the kernel breakdown's `bd.system` ≈ 0 by design — feeding the stored value as `systemPromptTokens` (the kernel's explicit "host-measured system" slot) does NOT double-count anything the breakdown already sees. Self-consistency follows automatically: the five breakdown rows sum exactly to the Sent total.
- Measuring the **outbound** text (what actually reaches the LLM) rather than the raw client input keeps the number truthful when bili injects its compress system prompt.
- Known residual: Responses-API `preamble` items (`additional_tools` / `mcp_list_tools`) are not counted — consistent with the existing `estimateInputTokens` / #453 clamp convention, which also ignores them. Small relative to real tool schemas; can be added later if it matters.

## Tests (`tests/issue532-panel-system-tokens.test.ts`, new)

1. Counter unit test: empty-safety and additivity vs per-part counting.
2. Panel math (acceptance #2): with non-zero `systemPromptTokens`, `Sent == classified + sysPrompt`, the `SysPrompt` row renders, and **sum of all breakdown rows == Sent total**; zero case keeps the row hidden.
3. Anthropic plugin-mode integration (acceptance #1): ~80k-char system (~20k tokens) + ~40k-char tool description (~10k tokens) through a real proxy → `SysPrompt` row renders within [client system + tools, +compress-prompt bound] and `Sent >= client system + tools`. Pre-fix this session showed Sent≈219.
4. Same integration via the OpenAI chat path.

## Pre-flight

- `npm run typecheck` — clean
- `npm test` — 1035/1036; the single failure (`launcher.test.ts` "codex/claude resolve to themselves") reproduces identically on pristine master in any environment where `codex` is installed at an absolute PATH location (resolver returns `/usr/bin/codex`, test expects bare `codex`) — pre-existing/environmental, unrelated to this change
- `npm run build` — success

### DIFF
--- a/src/plugin.ts
+++ b/src/plugin.ts
@@ -409,15 +409,24 @@ export function handlePluginStatus(conversationId: string, res: import("node:htt
     }
     let panel: string | undefined;
     try {
+        // #532: the kernel breakdown classifies messages only; bili measured
+        // the outbound system+tools overhead at prepare time (same source as
+        // estimateInputTokens) and stored it on the session. Feeding it in is
+        // what makes Sent/SysPrompt reflect reality instead of undercounting
+        // by the full system+tools size every turn. unprunedTokens gets the
+        // same addition so the Session-only derivation (unpruned − sent) still
+        // isolates pruned originals on one scale.
+        const sysTokRaw = session.metadata.systemPromptTokens;
+        const systemPromptTokens = typeof sysTokRaw === "number" && Number.isFinite(sysTokRaw) && sysTokRaw > 0 ? sysTokRaw : 0;
         panel = buildStatusPanel({
             version: `billion-context@${PROXY_VERSION}`,
             tokenCount: session.hostContextTokens ?? session.stats.lastInputTokens,
-            systemPromptTokens: 0,
+            systemPromptTokens,
             state: session.state,
             nudge,
             modelContextLimit,
             unprunedTokens: mem && mem.original.length > 0
-                ? mem.original.reduce((sum, m) => sum + defaultCountTokens(m.text ?? ""), 0)
+                ? mem.original.reduce((sum, m) => sum + defaultCountTokens(m.text ?? ""), 0) + systemPromptTokens
                 : undefined,
         });
     } catch {
--- a/src/server.ts
+++ b/src/server.ts
@@ -1834,6 +1834,10 @@ function prepareAnthropic(
         log("warn", `[${sessionId}] kernel transform failed, forwarding unchanged: ${String(err)}`);
         processedMessages = [];
     }
+    // #532: measure the outbound system+tools overhead for the status panel's
+    // SysPrompt row — the kernel breakdown classifies messages only, and on
+    // this wire the system rides the top-level `system` field outside the fold.
+    session.metadata.systemPromptTokens = countSystemAndToolsTokens(extractSystem(systemOut), toolsOut);
     snapshotMessages(session, originalMessages);
     markDirty(session);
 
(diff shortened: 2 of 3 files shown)


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 **PR review — verified, LGTM.** I checked out `2026-09-04_fix-panel-system-prompt-tokens` (bb49736, base = current master, clean 1-commit / 3-file / +315−5 diff) and independently re-ran everything:

**Bug is real (triple-confirmed)**
- Master's `handlePluginStatus` passes hardcoded `systemPromptTokens: 0`; the kernel panel computes `sentTotal = classified + systemPromptTokens` and skips rows with `value <= 0` (acp-kernel `panel/index.js`), so the SysPrompt row could never render.
- Ran this PR's two integration tests against **pristine master**: both fail with "SysPrompt row must render" — the tests genuinely capture the pre-fix bug.

**No double-counting — verified per wire in the kernel source**
- `computeContextBreakdown` classifies *messages only* (never tool schemas), so adding `toolsOut` tokens is purely additive on all wires.
- anthropic: system rides the top-level field, never enters `msgs` → `bd.system = 0` always; counting `extractSystem(systemOut)` (final `injectSystem` result, incl. compress prompt) is exact.
- openai: `openaiToCore` hoists only the leading system/developer (`msgs.length === 0`) into `systemText`; mid-conversation systems stay in `msgs` and are counted by `bd.system`, while `sysParts` counts only hoisted head + compress prompt → disjoint. Title-gen skip (`max_tokens <= 200`) correctly prevents clobbering the main session value.
- responses: `responsesToCore` hoists **all** system/developer items + `instructions` into `systemParts` → `bd.system = 0`; counting `devContent` (either injection branch, `?? ""` tools-only fallback) + `toolsOut` is exact when `transformOk`. Preamble residual is documented and consistent with the `estimateInputTokens`/#453 convention.
- Self-consistency: since `classified` includes no system on these wires, the five rows sum exactly to Sent. One edge: an OpenAI-wire session with **mid-conversation** system messages would have `bd.system > 0` included in Sent but with no breakdown row for it (the kernel panel has no System row at all — pre-existing property, not introduced here). Not blocking; worth knowing.

**Both compression modes considered** (AGENTS.md §6): the change adds only `session.metadata` writes at prepare time + one read in `handlePluginStatus` — zero wire changes (rebuilt body byte-identical). Panel is served exclusively via `/__bili/plugin/status` (pi/opencode/omp/hermes agents); proxy mode has no panel path, so it's unaffected. The `estimateInputTokens` refactor is algebraically identical. `unprunedTokens += systemPromptTokens` keeps the Session-only derivation (`unpruned − sentTotal`) on one scale — verified against kernel line 70.

**Pre-flight (reproduced independently)**
- `npm run typecheck` — clean
- `npm test` — 1035/1036; the single failure (`launcher.test.ts` "codex/claude resolve to themselves") reproduces identically on pristine master in this environment (resolver returns `/usr/bin/codex`) — pre-existing/environmental, as claimed
- `npm run build` — success
- E2E codex suite not run: needs a live upstream not available in this sandbox; not applicable in practice since no request-pipeline bytes changed.

**Minor nits (non-blocking, report-only)**
1. `src/server.ts:1778-1781` comment says "A head system message already in the rebuilt view is classified by the kernel breakdown" — that's backwards for the hoisted head system (it's *not* classified, which is exactly why it must be added back); it's mid-conversation systems that get classified. The issue description states it correctly; consider rewording the inline comment.
2. Transform-failure asymmetry (harmless): anthropic writes fresh metadata from raw parsed values after a failed transform; openai/responses keep the previous turn's value. Both are reasonable estimates.

Conventions all check out: branch naming, version untouched, acp-kernel pin untouched, no `as any`/`@ts-ignore`. No new issues to file from this review.

Ready for human merge per repo rules (I can't merge PRs): https://github.com/ranxianglei/billion-context/pull/534

PR #537, opened 2026-09-04
feat(#535): zero-file injection phases 2+3 — absolute-form forward proxy, hermes/dsh env routing

ranxianglei:
## What

Continues #535 (注入零文件化) after PR #536 (phase 1, pi). Three commits:

1. **`feat: absolute-form http forward-proxy requests ride the full pipeline`** (src/server.ts)
   `resolveUpstream` now recognizes standard forward-proxy request lines (`POST http://host/path`) and routes them exactly like the `/bili/` absolute-URL form: full compression pipeline + `tunnel: true`, so #409 destination admission applies (self/link-local denied, private per client locality). Malformed URLs fall through to passthrough. This is what lets hermes's httpx stack route plaintext-http upstreams through bili via `HTTP_PROXY`.

2. **`feat: hermes/dsh zero-file injection`** (src/launcher.ts + tests)
   - **hermes**: `prepareHermesHome` / `HERMES_HOME` overlay deleted. https upstream hosts are MITM-whitelisted (`HTTPS_PROXY` + `HERMES_CA_BUNDLE` → `combined-ca.pem`, since httpx builds its own bundle from that env — combined keeps system roots for blind-tunnelled hosts); plaintext-http upstreams ride `HTTP_PROXY` (absolute-form). Inherited `NO_PROXY` is loopback-sanitized in the child env so local sglang still reaches bili. The session-identity plugin (#286) is installed **idempotently into the real** `~/.hermes/plugins/model-providers/bili-session-identity/` (bili-owned subdir, coexists with user plugins; old skip-on-user-plugins guard removed).
   - **dsh**: split by destination. Non-loopback https → MITM whitelist; non-loopback plaintext-http → `HTTP_PROXY`; built-in deepseek route captured by whitelisting `api.deepseek.com` instead of clobbering `$DEEPSEEK_BASE_URL`. **Loopback destinations keep the legacy `settings.yaml` overlay rewrite** — documented exception: dsh bypasses proxies for localhost/127.0.0.1/::1 unconditionally (`LOOPBACK_NO_PROXY`, "the bypass is not optional"), so no env route exists there.
   - New exported helpers: `isLoopbackHost`, `stripLoopbackNoProxy`. `DiscoveredRoutes` += `httpEnvRoutes` / `providerKeys`.
[...]

PR #544, opened 2026-09-05
feat(#535): dsh phase 4 — split by destination (last overlay shrinks to loopback)

ranxianglei:
## What

Phase 4 of #535 (the last remaining issue item): dsh no longer force-routes **all** upstreams through the `/bili/` settings rewrite. Discovery splits `settings.yaml` endpoints by destination, matching dsh's actual fetch behavior:

| upstream | mechanism |
|---|---|
| loopback (http **or** https) | `/bili/` rewrite via persistent `DSH_HOME` overlay (`~/.dsh-bili`) — the documented exception: dsh's fetch stack bypasses proxy envs for loopback targets unconditionally (`LOOPBACK_NO_PROXY`, "the bypass is not optional") |
| non-loopback https | CONNECT + cert MITM — host added to the MITM whitelist, child gets `HTTPS_PROXY` + `SSL_CERT_FILE` → `combined-ca.pem` (it *replaces* dsh's trust store, so system roots must be bundled — same pattern as codex) |
| non-loopback plain-http | absolute-form forward-proxy requests via `HTTP_PROXY` (server support landed in #538) |
| built-in deepseek-official route | unchanged: `$DEEPSEEK_BASE_URL` wrap |

Behavioral parity: proxy envs are set only when something actually routes through them — a launch with no non-loopback custom providers produces the same child env as before. Inherited proxy vars are fully stripped from the child (same contract as hermes); dsh's loopback exclusion comes from its built-in policy, so no `NO_PROXY` juggling is needed.

## Docs (EN+ZH)

- README: new “Injection priority — no files unless unavoidable” section next to *Two compression modes* (env > CLI/extension API > generated files; bili never owns user data; legacy overlays left in place, never merged back) + updated dsh one-liner.
- CONFIGURATION: redirect table, discovery table and the generated-files section updated for the dsh split — **plus** the pi/omp/hermes rows and bullets that phases 1–2 left stale (they still described the removed `PI_CODING_AGENT_DIR` / `HERMES_HOME` overlays). All phases are merged now, so the docs finally match the code.

## Pre-flight

- `npm run typecheck` clean
- `npm test`: 1028/1029 pass — sole failure is the sandbox-only `resolveClientCommand: codex/claude resolve to themselves` (`/usr/bin/codex` exists in this sandbox; fails on pristine master too)
- `npm run build` OK

New tests: `isLoopbackHost` units; dsh discovery split (loopback→rewrite, https→whitelist, http→env-routes, wrapped-value unwrap, dedupe, malformed skip); runLaunch dsh combined fixture (proxy envs + NO_PROXY strip + MITM-domain capture + overlay rewrites ONLY the loopback line + real home byte-identical); runLaunch dsh no-loopback fixture (no `DSH_HOME`, patch-only `-bili` dir).

## Notes

- This is the final phase of #535. After merge, the only generated-file mechanisms left in the launcher are opencode's temp `opencode.json` (removed on exit) and the dsh loopback overlay — both documented as last-resort exceptions. Full `refreshOverlayHome` abolition can be revisited once dsh gains a settings-path env or an upstream loopback opt-out.
- No version bump (release flow territory).

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -387,7 +387,7 @@ Full command surface (`bili --help` prints an abridged version). Precedence ever
 | `bili omp [opts --] [args]` | Proxy + **omp** (pi-based) |
 | `bili opencode [opts --] [args]` | Proxy + **opencode** |
 | `bili hermes [opts --] [args]` | Proxy + **hermes-agent** (`/bili/` rewrite) |
-| `bili dsh [opts --] [args]` | Proxy + **deepseek-harness** (`/bili/` rewrite; args like `--profile web "task"` pass through) |
+| `bili dsh [opts --] [args]` | Proxy + **deepseek-harness** (non-loopback upstreams via proxy envs, loopback via `/bili/` rewrite — #535; args like `--profile web "task"` pass through) |
 | `bili test pi` | Non-polluting end-to-end smoke test of the pi path |
 | `bili export [session] [--full] [--output FILE]` | List persisted sessions / export one as a Markdown handoff — see [Sessions & Migration](#sessions--migration) |
 | `bili update` | Check for & install a newer version now (bypasses the 3-minute throttle) |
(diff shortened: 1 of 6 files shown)


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 Reviewed PR #544 against master (1502287) — **LGTM, verified, ready for merge** (merge itself stays with you).

## What I independently re-verified

**Pre-flight (re-run on head a819ed8):** `npm run typecheck` clean; `npm test` 1028/1029; `npm run build` OK. The sole failure (`resolveClientCommand: codex/claude resolve to themselves`) is environmental — I checked out pristine master in a separate worktree and it fails there too (launcher.test.ts 98/99 on master), so it's not a regression.

**Code (src/launcher.ts):**
- Destination split in `discoverRoutes` is correct: loopback (http *or* https) → `/bili/` rewrite keys; non-loopback https → MITM whitelist (deduped by lowercased hostname, `/bili/-wrapped values unwrapped first); non-loopback plain-http → `httpEnvRoutes`; malformed entries skipped; dedupe by real upstream preserved from before.
- `isLoopbackHost` (localhost / LOCALHOST / ::1 / [::1] / 127.0.0.0/8) is a **superset** of dsh's exact-match bypass list (`LOOPBACK_NO_POLICY = ['localhost','127.0.0.1','::1','[::1]']`, policy.ts:33 per #535). That direction is safe: any endpoint dsh would bypass is also routed by bili via the rewrite path, so there is no silent-bypass direction. The only theoretical divergence (a host dsh treats as non-loopback that bili also sees as non-loopback) lands on proxy envs, which dsh honors — consistent.
- Behavioral parity claim confirmed against master code: master's dsh branch did not strip inherited proxy vars either, and phase 4 only applies `stripInheritedProxy` when `usesProxyEnv` is true — so a launch with no non-loopback custom providers produces the identical child env as before.
- `prepareDshHome` implementation is byte-for-byte unchanged (docstring only) — the overlay now just receives fewer lines to rewrite, which the combined fixture proves (`overlayTxt` contains the rewritten loopback line AND the untouched `https://api.anthropic.com` line; real home byte-identical).
- Warning ladder correctly updated (loopback rewrite failure now says "loopback custom providers will NOT go through the proxy; other routes still do").

**Server-side dependency:** absolute-form forward-proxy handling is present in master from #538 (src/server.ts:294-311 — absolute-form request line treated like the `/bili/` embedded form, full pipeline + #409 destination admission). The `HTTP_PROXY` route for plain-http non-loopback upstreams has its server half landed.

**MITM wiring:** `routes.httpsDomains` flows into the proxy spawn via `domains = dedupeInOrder([...routes.httpsDomains, ...params.mitmDomains])` (src/launcher.ts:1710) → `BILI_MITM_DOMAINS` (:1518). The new runLaunch fixture captures the proxy-spawn env and asserts `api.anthropic.com` is in it — end-to-end wiring covered, not just discovery.

**Tests (+129):** match the description exactly — split matrix incl. wrapped-value unwrap/dedupe/malformed-skip, `isLoopbackHost` units (incl. `1270.0.0.1` negative), combined fixture (HTTPS_PROXY + combined-ca.pem, NO_PROXY stripped, HTTP_PROXY absent when unused, overlay rewrites ONLY the loopback line, profiles symlink, patch file + `--patch` splice), no-loopback fixture (inherited DSH_HOME passes through unchanged, `-bili` dir holds only `.bili-acp.patch.yml`).

**Docs (EN+ZH):** accurate and consistent with the code. The stale pi/omp rows (`PI_CODING_AGENT_DIR`) and hermes row (`HERMES_HOME`, "all upstreams /bili/") left over from phases 1–2 are fixed, the generated-files section is retitled to last-resort framing, and the new README "Injection priority" section sits next to *Two compression modes* as described. ZH mirrors EN without drift.

## Non-blocking notes

1. **No real-machine dsh e2e was possible here** (no dsh binary in this environment) — the change hinges on dsh honoring `HTTPS_PROXY`/`HTTP_PROXY`/`SSL_CERT_FILE`, which is source-verified (#535 OP, deepseek-harness `packages/util/http-proxy`) but not runtime-exercised in CI. Suggested post-install smoke: one loopback upstream (overlay path), one remote https provider (MITM path), optionally one remote plain-http (absolute-form path).
2. `DEEPSEEK_BASE_URL` wrap intentionally kept (per your floor-12 scoping); the whitelist alternative from the #537 review (add `api.deepseek.com` instead of overriding user-set env) remains available as a follow-up if you ever want zero env override there.
3. After merge, all #535 checklist items are landed — the umbrella issue can be closed whenever you're ready (the two documented exceptions — opencode temp config, dsh loopback overlay — are tracked in this PR's notes rather than open work items).

No version bump present, release-flow rules respected. Nothing here touches the request pipeline (server.ts/loop/adapters), so the codex E2E suite isn't triggered by this change.

Issue #549, opened 2026-09-05
`extractEventType` rejects valid `data:`-only SSE streams → synthetic "upstream returned no response"


Juna9969:
**Project**: billion-context
**Version**: v0.1.85 (npm, bin `bili`)
**Environment**: Windows 10/11, Node via npm global install, used as OpenAI-compatible reverse proxy in front of a custom SSE gateway

## Summary

In `dist/index.js`, `extractEventType()` only recognizes SSE events that carry an explicit `event:` line. Streams that emit valid SSE with **only `data:` lines** (which is legal: per the WHATWG server-sent events spec, `data:` is the only required field and `event:` defaults to `message`) get **every event skipped**. `parseStream` never sees a terminal event, so when the upstream stream ends normally the proxy synthesizes:

```
event: response.failed
error: {"code": "server_error", "message": "upstream returned no response"}
```

…and the client (codex) shows `stream disconnected before completion: upstream returned no response`, even though the upstream completed successfully and delivered a full `response.completed` event.

## Root cause

`extractEventType` (≈ `dist/index.js:52400`):

```js
function extractEventType(rawEvent) {
  for (const l of rawEvent.split("\n")) {
    if (l.startsWith("event:")) return l.slice(6).trim();
  }
  return null;
}
```

`parseStream` (≈ `dist/index.js:52564`) then does:

```js
const type = extractEventType(eventStr);
if (!type || !dataLine) continue;   // ← all data-only events skipped
```

`iterSseEvents` (≈ `dist/index.js:52372`) splits frames on `\n\n`, which is fine. The problem is only the `event:` dependency.

Many real-world OpenAI-compatible servers (including official clients' own minimal gateways, and custom proxies) emit events as pure `data: {"type":"response.created",...}` lines with no `event:` line — this is what the codex CLI itself tolerates (it reads the `type` field from the JSON payload). The proxy should do the same as a fallback.

## Minimal reproduction

1. Run any SSE server that responds to `POST /v1/responses` with `stream: true` and emits only:
   ```
   data: {"type":"response.created"}
   data: {"type":"response.completed"}
   ```
   (no `event:` lines) — e.g. a 20-line Python `BaseHTTPRequestHandler`.
2. Start the proxy: `bili start --port 8786` (or `bili codex`).
3. Point a client at `http://127.0.0.1:8786/bili/http://127.0.0.1:<port>/v1` and make a streaming request (send an `x-session-id` header — required by this proxy).
[...]

PR #550, opened 2026-09-05
fix: accept data-only SSE streams in responses adapter (#549)

ranxianglei:
## What changed

`src/loop/adapter-responses.ts` — `parseStream` now falls back to the `type` field inside the `data:` JSON payload when a frame carries no explicit `event:` line.

Per the WHATWG SSE spec `event:` is optional (defaults to `message`), and many OpenAI-compatible gateways emit pure `data:` frames. Previously every such frame was skipped (`if (!type || !dataLine) continue;`), so no terminal event was ever observed and a cleanly-finished upstream stream produced the synthetic `response.failed` / "upstream returned no response" error reported in #549.

- An explicit `event:` line still wins over the payload `type`.
- Strictly additive: non-JSON data or payloads without a string `type` are still skipped exactly as before.
- The payload is parsed once; the fallback reads the already-parsed object instead of re-parsing inside `extractEventType` (same semantics as the patch proposed in the issue, without the double parse).

## Tests

Three regression tests added in `tests/loop-adapters.test.ts` (F8):
1. data-only stream (created → item.added → text.delta → completed) yields text + usage + `done/completed`, not a synthetic truncation failure;
2. explicit `event:` takes precedence over the payload `type`;
3. end-to-end drain of a data-only stream emits `response.completed` and never "upstream returned no response".

Verified tests 1 and 3 fail against the unfixed code.

Note: `parseAnthropicSse` in `src/loop/adapter-anthropic.ts` has the same strictness (requires an `event:` line). Left unchanged — Anthropic streams always carry event lines and no compatible gateway omitting them has been reported; flagged for awareness.

## Pre-flight

- `npm run typecheck`: pass
- `npm test`: 949/950 pass; the single failure (`resolveClientCommand: codex/claude resolve to themselves`) fails identically on the clean tree in this environment (client binaries not installed) — unrelated
- `npm run build`: pass

### DIFF
--- a/src/loop/adapter-responses.ts
+++ b/src/loop/adapter-responses.ts
@@ -316,15 +316,21 @@ export function createResponsesAdapter(textProtocol?: boolean, projection?: Resp
                 }
             };
             for await (const eventStr of iterSseEvents(upstream)) {
-                const type = extractEventType(eventStr);
+                const explicitType = extractEventType(eventStr);
                 const dataLine = extractDataLine(eventStr);
-                if (!type || !dataLine) continue;
+                if (!dataLine) continue;
                 let obj: Record<string, unknown>;
                 try {
                     obj = JSON.parse(dataLine);
                 } catch {
                     continue;
                 }
+                // #549: `event:` is optional per the WHATWG SSE spec (defaults to
+                // "message"); many OpenAI-compatible gateways emit pure data: frames.
+                // Responses payloads carry a `type` equal to the event name — fall
+                // back to it when no event: line is present (explicit line still wins).
+                const type = explicitType ?? (typeof obj.type === "string" ? obj.type : null);
+                if (!type) continue;
                 const rawBuf = Buffer.from(eventStr + "\n\n", "utf8");
                 if (round === 1 && typeof obj.output_index === "number") {
                     outputIndex = Math.max(outputIndex, obj.output_index + 1);
--- a/tests/loop-adapters.test.ts
+++ b/tests/loop-adapters.test.ts
@@ -41,7 +41,7 @@ function makeCtx(id: string, messages: CoreMessage[] = []): {
 async function drain(
     stream: ReadableStream<Uint8Array>,
     ctx: ReturnType<typeof makeCtx>,
-    adapter: ReturnType<typeof createOpenaiAdapter> | ReturnType<typeof createAnthropicAdapter>,
+    adapter: ReturnType<typeof createOpenaiAdapter> | ReturnType<typeof createAnthropicAdapter> | ReturnType<typeof createResponsesAdapter>,
     requestBody: Record<string, unknown>,
 ): Promise<string> {
     const chunks: Buffer[] = [];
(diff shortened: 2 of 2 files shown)


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 Review complete — **LGTM, ready for human merge** (merging stays human-only per repo rules).

**Bug verification (independent, before touching code)**
Traced the full failure chain in `src/loop/adapter-responses.ts`: `extractEventType()` returns null for any frame without an `event:` line → old `if (!type || !dataLine) continue;` dropped *every* frame of a data-only stream → no terminal event ever observed → `done/failed/truncated` yielded at line ~483 → `emitCompletion()` synthesizes `response.failed` with `"upstream returned no response"` (line ~506–524) whenever `finishReason === "failed"` and no `response.completed` object was captured. This matches the #549 report exactly (cleanly-finished upstream stream → synthetic failure). Root cause is frame classification in the loop's Responses adapter — not the gateway, not the client — and the patch fixes precisely that layer, not a symptom.

**Semantics audit**
- Strictly additive, as claimed: the only frames whose handling changes are those with no `event:` line + valid JSON + string `type`. Non-JSON data still hits the parse catch and is skipped; JSON without a string `type` is still skipped; an explicit `event:` line still wins; a bare `event:` (empty name) is still falsy → skipped, same as before.
- Consistency: `src/stream-terminal.ts` (`terminalFromSseLine`) already implements exactly this event-line-first-then-payload-`type` fallback for the native-compaction observer — the Responses adapter was the outlier, and this brings it in line with the codebase's own established interpretation of the spec.
- Both compression modes considered: `parseStream` is shared by plugin mode and proxy mode (the adapter is selected by protocol only, `src/loop/index.ts`), and the diff touches no tool/nudge injection or suppression logic — the fix applies uniformly to both modes with no mode-specific risk.
- Leaving `parseAnthropicSse` strict is fine — no compatible Anthropic gateway omitting `event:` lines has been reported; worth a follow-up issue only if one surfaces.

**Pre-flight (re-run independently on branch `2026-09-05_responses-data-only-sse`, commit 1d8527f)**
- `npm run typecheck`: pass
- `npm test`: 949/950 — the single failure is `resolveClientCommand: codex/claude resolve to themselves` (`tests/launcher.test.ts`); environmental (a `codex` binary exists on PATH in this sandbox, so it resolves to an absolute path). The diff touches neither `launcher.ts` nor its test — unrelated.
- `npm run build`: pass
- Negative check: reverted only `src/loop/adapter-responses.ts` to master while keeping the new tests → the two data-only F8 tests fail (23 pass / 2 fail in that file) while the `event:`-precedence test passes. Confirms the new tests actually pin the regression and the fix is what flips them.

Scope is clean: single commit, 2 files (+56/−3), no version change. No duplicate open issues/PRs found covering this fix.

ranxianglei:
简述这个主要是啥问题

ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 一句话:**上游 SSE 流没带 `event:` 行时,bili 把整个流当成"没有响应"了。**

展开一点:

- SSE 规范里 `event:` 行是可选的(缺省就是 `message`),不少 OpenAI 兼容网关只发纯 `data:` 帧。
- 但 responses 适配器的 `parseStream` 要求每帧必须有显式 `event:` 行,否则整帧跳过(`if (!type || !dataLine) continue;`)。
- 结果:data-only 流的每一帧都被丢掉,包括结尾的 `response.completed` → bili 认为"上游从未给出终态"→ 明明上游正常结束,却合成一个假的 `response.failed`(报错 "upstream returned no response")返回给客户端。

修复:帧上没有 `event:` 行时,回退读 `data:` JSON 里的 `type` 字段当事件类型(显式 `event:` 仍然优先)。改动严格增量——其他情况行为完全不变。

PR #556, opened 2026-09-05
fix: align undici transport timeouts with idle watchdog; add BILI_UPSTREAM_TIMEOUT_MS (#551)

### LINKED ISSUE #551: 上游长时间 prefill(首 token 前无输出)被 undici 默认 bodyTimeout=300s 掐断,UPSTREAM_TIMEOUT_MS=10min 配置形同虚设,大上下文请求必超时
## 现象

代理转发请求到本地大模型(llama.cpp,150,528 窗口)时,若上下文很大(全量重放 ~13 万+ token,27B Q4 Vulkan prefill 需要 5-10 分钟才出首 token),bili 在固定 ~305s 处掐断流,客户端 5-6 次自动重试全部同样失败,最终报「Request timed out」。

bili 日志(bili.log,UTC)精确复现:

```
11:58:42 [acp-loop] round 1: upstream stream truncated (no completion event; round 1, 0 text chars + 0 reasoning chars received)
12:03:49 ...(同上,间隔精确 ~305s)
12:08:56 ...
12:14:03 ...
12:19:14 ...
12:24:27 ...
```

## 根因

`src/fetch-util.ts` 的 `fetchWithTimeout`(L48-108)自带空闲看门狗:

```ts
// L12
export const UPSTREAM_TIMEOUT_MS = 10 * 60 * 1000;  // 10 分钟,chunk 到达会 rearm(L110-134)
// L83
const finalOpts: Omit<RequestInit, "dispatcher"> & { dispatcher?: object } = { ...opts, signal: controller.signal };
```

但底层是 undici 的全局 `fetch`,其 **默认 `bodyTimeout` = 300000ms(300s)**(读 body 数据之间的空闲超时)。`finalOpts` 没有传 `bodyTimeout`/`headersTimeout`,所以 undici 的 300s 看门狗必然先于 bili 自研的 10 分钟定时器触发:prefill 阶段 0 个 body chunk → 300s 无数据即被 undici abort。

观测到的 ~305s 与 undici 默认 300s + 调度延迟吻合;而 `UPSTREAM_TIMEOUT_MS = 10min` 从未有机会触发——形同虚设。

## 影响

- 任何「首 token 延迟 >300s」的上游(本地大模型大上下文重放、云端高负载排队)都会让 bili 必断流,且重试徒劳(每轮同样 305s);
- 与 fork 会话的 preflight 退化叠加时几乎必现:压缩重放导致 fork 会话 `lastInputTokens=0`,preflight 退化为估算(`tokenCount = max(lastInputTokens, payloadEstimate)`),大包裸转不压缩 → prefill 更长 → 更容易超过 300s;
- 10 分钟可配置超时对用户产生误导(配了不生效)。

## 建议修复

在 `fetchWithTimeout` 的 `finalOpts` 中显式传入与自有看门狗一致的超时(或做成可配置项/环境变量):

```ts
const finalOpts = {
  ...opts,
  signal: controller.signal,
  bodyTimeout: UPSTREAM_TIMEOUT_MS,      // 与 L12 的 10 分钟 idle 看门狗对齐
  headersTimeout: UPSTREAM_TIMEOUT_MS,   // 首字节前等待同样对齐
};
```

本地验证:手工 patch 后(15 分钟)大上下文请求不再被 300s 掐断。

## 相关观察(已拆分为独立 issue)

以下观察已分别提为独立 issue 跟踪:

1. 侧请求绕过 preflight(→ [#554](https://github.com/ranxianglei/billion-context/issues/554)):`max_tokens<=200` 的请求判定为 side request 直接 passthrough,大 payload 也会裸转上游,上游 400 `exceed_context_size_error`(198,277>150,528)后又无法触发 learned 自适应;
2. fork 会话 `lastInputTokens=0` → preflight 用估算,大包可能不压缩(→ [#553](https://github.com/ranxianglei/billion-context/issues/553));
3. 窗口口径三处不一致:settings.yaml(GUI 显示 140,000)/ bili effective(140,000 − max_tokens 输出预留 = 110,000,`reserveOutputHeadroom`)/ llama 后端真实窗口 150,528,排查问题时容易误判。

ranxianglei:
Fixes #551.

## Problem

Upstreams with long prefills (no output before the first token — e.g. ~130K-token replays on local models taking 5–10 min) were always cut at a fixed ~305s: Node's global undici agent applies default `headersTimeout`/`bodyTimeout` of **300s** to every direct request, and bili never overrode them. That transport cap fired before bili's own `UPSTREAM_TIMEOUT_MS` (10 min) idle watchdog could, making the watchdog ineffective for exactly the slow-prefill case it exists for. Client retries were equally futile — every round died at the same 305s wall.

Note: undici's fetch layer ignores per-request `bodyTimeout`/`headersTimeout` init options (verified against undici 7.29 source and empirically), so the effective lever is **dispatcher-level** configuration.

## Changes

- `src/fetch-util.ts`: `fetchWithTimeout` now injects a direct `undici.Agent` (cached per timeout value) whose `headersTimeout`/`bodyTimeout` match the idle watchdog when the caller passes no dispatcher. New `BILI_UPSTREAM_TIMEOUT_MS` env var (ms, read live like the other `BILI_*` knobs) makes the budget configurable; default unchanged (10 min).
- `src/upstream-proxy.ts`: `proxyDispatcher(proxyUrl, timeoutMs?)` builds the `ProxyAgent` with matching `headersTimeout`/`bodyTimeout`; explicit `factory`/`clientFactory` pools ensure every hop — including the CONNECT tunnel client and the http-origin-via-http-proxy wrapper — inherits the budget instead of falling back to 300s.
- Call sites: `/__bili/upstream/test` and the registry fetch pass their short explicit budgets (15s) to both layers; the main request path and compress-loop paths resolve the budget once, live, identically in both layers.
- `CONFIGURATION.md`: documents `BILI_UPSTREAM_TIMEOUT_MS`.

Design choice: **align** the transport layer with the existing watchdog (same policy value, two independent enforcement layers) rather than disabling undici's caps (`0`) — defense in depth; a regression in either layer is caught by the other.

Mode reasoning (AGENTS.md §6): transport-layer only. Plugin mode and proxy mode both route through `fetchWithTimeout`/`fetchWithRetry`, so behavior is identical in both.

## Tests

New `tests/fix-551-upstream-timeout.test.ts`:
- env parsing (default / passthrough / garbage values);
- silence longer than the budget is cut at the budget (**pre-fix this case completed at 3s** — both old layers allowed it);
- silence within the budget completes (no premature kill of a healthy slow prefill);
- `ProxyAgent` cache keying per (url, timeout).

## Pre-flight

- `npm run typecheck` — clean
- `npm test` — 1050/1051; the single failure (`resolveClientCommand: codex/claude`) reproduces identically on clean master in any environment where a `codex` binary happens to be on PATH (the test assumes none) — pre-existing, unrelated
- `npm run build` — success
- E2E (`tests/e2e/e2e-codex.test.ts`) not run here: it needs a Responses-compatible upstream not available in this sandbox. Please trigger the manual-dispatch CI e2e workflow before merging.

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -353,6 +353,7 @@ Environment variables take precedence over the config file. They are useful for 
 | `ACP_PROVIDERS` | Path to an external `providers.json` (legacy / shared file). |
 | `BILI_REPLAY_RETRY_BASE_MS` | Base backoff delay (ms) for acp-loop replay retries after a transient upstream rejection (default `1500`; set `0` to disable the delay). See #189. |
 | `BILI_REPLAY_RETRY_MAX` | Total attempts for acp-loop replay retries (default `3`; set `1` to disable retries entirely — legacy fail-fast behavior). See #189. |
+| `BILI_UPSTREAM_TIMEOUT_MS` | Idle budget (ms) for upstream requests: time-to-first-byte and time between body chunks (default `600000` = 10 min). A healthy stream that keeps producing chunks is never cut mid-flight; a silent one is. The same value drives the underlying HTTP client's transport timeouts, so this single knob bounds long local-model prefills end-to-end (#551). |
 | `ACP_SESSION_HEADER` | Conversation-id header name (default `x-acp-session`). |
 | `ACP_REASONING_KEEP` | Responses API only: set `none` to drop all reasoning items. Default routes reasoning through the compression pipeline so it is hidden automatically once its turn is summarized (prevents the unbounded accumulation that broke Codex's prompt-cache prefix). |
 | `ACP_LOG_FILE` | Log file path (default XDG state path; `off` disables the file, keeps stderr). Auto-rotates at 10 MB. |
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -353,7 +353,7 @@ Environment variables take precedence over the config file. They are useful for 
 | `ACP_PROVIDERS` | Path to an external `providers.json` (legacy / shared file). |
 | `BILI_REPLAY_RETRY_BASE_MS` | Base backoff delay (ms) for acp-loop replay retries after a transient upstream rejection (default `1500`; set `0` to disable the delay). See #189. |
 | `BILI_REPLAY_RETRY_MAX` | Total attempts for acp-loop replay retries (default `3`; set `1` to disable retries entirely — legacy fail-fast behavior). See #189. |
-| `BILI_UPSTREAM_TIMEOUT_MS` | Idle budget (ms) for upstream requests: time-to-first-byte and time between body chunks (default `600000` = 10 min). A healthy stream that keeps producing chunks is never cut mid-flight; a silent one is. The same value drives the underlying HTTP client's transport timeouts, so this single knob bounds long local-model prefills end-to-end (#551). |
+| `BILI_UPSTREAM_TIMEOUT_MS` | Idle budget (ms) for upstream requests: time-to-first-byte and time between body chunks (default `720000` = 12 min). A healthy stream that keeps producing chunks is never cut mid-flight; a silent one is. The same value drives the underlying HTTP client's transport timeouts, so this single knob bounds long local-model prefills end-to-end (#551). |
 | `ACP_SESSION_HEADER` | Conversation-id header name (default `x-acp-session`). |
 | `ACP_REASONING_KEEP` | Responses API only: set `none` to drop all reasoning items. Default routes reasoning through the compression pipeline so it is hidden automatically once its turn is summarized (prevents the unbounded accumulation that broke Codex's prompt-cache prefix). |
 | `ACP_LOG_FILE` | Log file path (default XDG state path; `off` disables the file, keeps stderr). Auto-rotates at 10 MB. |
--- a/CONFIGURATION.zh-CN.md
+++ b/CONFIGURATION.zh-CN.md
@@ -356,6 +356,7 @@
 | `ACP_LOG_FILE` | 日志文件路径（默认 XDG state 路径；`off` 关闭文件只保留 stderr）。10 MB 自动轮转。 |
 | `ACP_DUMP_SSE` | 调试用：转储原始 SSE 帧的目录。 |
 | `BILI_UPSTREAM_PROXY` | 代理自身出站连接的上游代理 —— 优先级最高，高于 per-URL/per-provider 配置。见 README「上游代理」一节。 |
+| `BILI_UPSTREAM_TIMEOUT_MS` | 上游请求的空闲预算（毫秒）：首字节时间（TTFB）与响应体块之间的间隔（默认 `720000` = 12 分钟）。持续产出数据块的健康流永远不会被中途切断；静默的流才会。同一个值同时驱动底层 HTTP 客户端的传输层超时，因此这一个旋钮即可端到端约束本地大模型的超长 prefill（#551）。 |
 | `BILI_PERSIST` | 设 `0` 关闭会话持久化（仅内存，重启即丢）。 |
 | `BILI_PERSIST_DEBOUNCE_MS` | 持久化写盘的防抖窗口（毫秒，默认 `500`）。 |
 | `BILI_MAX_SESSIONS` | 内存中最多保留的会话数（默认 `256`；LRU 淘汰 —— 磁盘是事实源）。 |
--- a/src/fetch-util.ts
+++ b/src/fetch-util.ts
@@ -1,3 +1,5 @@
+import { Agent } from "undici";
+
 /** HTTP robustness helpers for the proxy.
 
   - readBody is capped: an unbounded request body is a memory-exhaustion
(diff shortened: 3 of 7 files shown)


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 Review in progress. Plan: re-verify the root-cause claim against the undici source actually pinned here, audit the full diff (both compression modes, all call sites of `fetchWithTimeout`/`proxyDispatcher`), re-run typecheck/test/build, and independently reproduce the claimed pre-existing `resolveClientCommand` failure on clean master to confirm it's unrelated. Will follow up with findings; E2E needs a manual-dispatch CI run before merge per AGENTS.md.

ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 **Review verdict: APPROVED — every technical claim independently verified.** Ready for the manual-dispatch E2E CI run, then human merge (merge stays human-only per AGENTS.md).

## Root cause & design — confirmed against the pinned source (undici 7.29.0, exactly what `npm ci` resolves here)

- `undici/lib/dispatcher/client.js:261-262`: `this[kBodyTimeout] = bodyTimeout != null ? bodyTimeout : 300e3` (same for headers) — the hidden 300s transport cap is real and applies to every Client/Pool, including Node's global agent. Matches the observed ~305s wall in #551.
- `undici/lib/web/fetch/index.js`: **zero** occurrences of `bodyTimeout`/`headersTimeout` — the fetch layer cannot see per-request timeout init options. Empirically confirmed too: `fetch(url, { bodyTimeout: 800, headersTimeout: 800 })` against a 3s-silent upstream completed at ~3005ms. So the dispatcher-level lever is indeed the only one, and the original reporter's suggested patch (fields into `finalOpts`) would not have worked. The correction you made in #551 floor 6 stands.
- `factory` and `clientFactory` are both real, load-bearing `ProxyAgent` options in 7.29.0 (`proxy-agent.js:101-104, 133, 157, 165`). Critically, the CONNECT-tunnel client is built at line 165 as `clientFactory(url, { connect })` — a bare option bag that top-level constructor options never reach. Without your `clientFactory`, the CONNECT hop would keep the 300s cap. Your comment in `upstream-proxy.ts` describing this is accurate.

## Empirical matrix (all on local loopback servers, this Node version)

| Case | Result |
|---|---|
| Plain `fetch`, 3s post-headers silence | completes (~3s, below 300s cap) |
| Old-style bare `ProxyAgent({ uri })` through a CONNECT proxy, 3s silence | **completes** — reproduces the pre-fix behavior your test describes ("pre-fix this case completed at 3s") |
| PR-style `ProxyAgent` (top-level timeouts + `factory` + `clientFactory`), 800ms budget, 3s silence | cut at ~1s, `UND_ERR_BODY_TIMEOUT` |
| Same PR-style agent vs a proxy that accepts TCP but never answers CONNECT | cut at ~1s, `UND_ERR_HEADERS_TIMEOUT` — proves `clientFactory` covers the CONNECT hop specifically |
| Direct `new Agent({ headersTimeout: 800, bodyTimeout: 800 })` injected into global `fetch` | cut at ~1s (surfaces as `TypeError: terminated` on the body stream — same mid-stream-truncation class as today's 305s kills, so no new failure mode for the stream handlers) |

Cross-copy injection (package undici 7.29 `Agent`/`ProxyAgent` passed to Node's built-in global fetch) works — same mechanism bili already ships for the proxied path.

## Call-site audit (both compression modes)

All 7 call sites are consistent — either explicit 15s in **both** layers or live-default in **both**: `/__bili/upstream/test` (`src/server.ts:828-830`) and registry fetch (`src/registry.ts:103,109`) pass `15_000` to both; main request path (`src/server.ts:2741/2758`), fake-completion retry (`src/server.ts:3267-3273`), preflight (`src/preflight.ts:189-195`), compress-loop responses (`src/compress-loop-responses.ts:216-220`), and loop core (`src/loop/core.ts:227-233`) all omit the arg in both layers, resolving `upstreamTimeoutMs()` live. No site pairs a short watchdog with a long transport cap (or vice versa). Transport-layer-only change → identical behavior in plugin and proxy mode, per AGENTS.md §6.

One wording nit (non-blocking): the description says the main path "resolves the budget once" — actually each layer reads the env independently (two statements apart). Functionally identical unless the env var mutates mid-request.

## Regression check: explicit direct `Agent` vs Node's hidden global agent

Verified Node's global fetch here does **not** honor `http(s)_proxy` env vars (set the vars, pointed at a logging proxy: zero hits, response came direct) — consistent with the existing `registry.ts` comment. So swapping the hidden global agent for an explicit one changes nothing for env-proxy-dependent setups; bili manages proxies explicitly regardless.

## Pre-flight (re-run independently, not taken on faith)

- `npm run typecheck` — clean
- `npm run build` — clean
- `npm test` — **1050/1051**. The single failure is `resolveClientCommand: codex/claude resolve to themselves` (`tests/launcher.test.ts`), and I reproduced it **identically on clean master**: it fires wherever a `codex` binary happens to be on PATH (assertion expects the bare name, gets the absolute path). Pre-existing, environment-dependent, unrelated to this diff.
- E2E (`tests/e2e/e2e-codex.test.ts`) — not runnable in this sandbox (needs a Responses-compatible upstream). **Please trigger the manual-dispatch CI e2e workflow before merging**, as noted in the PR body.

## Housekeeping notes (non-blocking)

- Master moved ahead while this was in flight (PR #555 logger rotation: `src/logger.ts`, `tests/logger-rotate.test.ts`). **Zero file overlap** with this PR's six files — it will merge cleanly against current master with no rebase.
- If `BILI_UPSTREAM_TIMEOUT_MS` were changed at runtime, each distinct value creates a new cached `Agent` and older ones stay open until process exit. In practice the knob is set once per process, so the cache is bounded (≤ a couple of agents). FYI only.
- New tests are well-targeted (env parsing incl. `"0"`/`"-5"`/`"1.5"` garbage, cut-at-budget, within-budget completion, `(url, timeout)` cache keying) and all four pass in the suite. `_resetFetchUtilForTest()` closes cached agents properly.

Bottom line: root cause verified in source, fix mechanics verified end-to-end including the CONNECT hop, both compression modes covered, pre-flight green modulo the known pre-existing env-dependent failure. Ship it after the CI e2e green — and the merge itself is yours to click.

PR #563, opened 2026-09-05
fix: forward-proxy requests no longer misread as self, restoring per-upstream window config (#562)

### LINKED ISSUE #562: bug: 标准 HTTP 代理请求被误判为自身请求，导致上游模型窗口配置失效
## 背景

关联 #535。希望保留零文件化接入的目标：客户端不生成配置副本、不改写模型地址，通过 HTTP_PROXY 等代理环境变量接入 bili。

我们在本地将 DSH 改为正向代理接入后，发现请求仍能正常转发，但上游模型窗口配置没有生效。DSH 的接入方式是本地改动；下述路由判断也存在于上游 master 的 020d45577a40f871299e3ed98a01c65341b3b51a。

本文使用示例主机名和模型名，不包含实际部署地址、账户信息或会话内容。

## 请求形式与复现

使用 DSH 当前的 HTTP 传输代码连接本地测试接收器，捕获到的请求形式如下（地址已替换）：

```http
POST http://model-server.example.invalid:8080/v1/responses HTTP/1.1
Host: model-server.example.invalid:8080
```

网络连接目标是代理 127.0.0.1:8787；HTTP Host 标识的是上游服务，而不是代理自身。

配置条件：全局压缩窗口 200000，modelContextLimitStrategy 为 min；providers 中为上述上游的 example-model 配置 context: 120000。请求的 model 同样为 example-model，max_output_tokens 为 32768。

本地以该请求形式调用实际路由和配置解析代码，并与 /bili/ 包装地址做对照，结果为：

| 请求形式 | 解析出的窗口 | 扣除输出预留后的容量 |
| --- | ---: | ---: |
| 正向代理：绝对 URL，Host 为上游 | 200000（错误） | 167232 |
| /bili/http://... 对照 | 120000（正确） | 87232 |

真实运行日志也出现了 167232。这里验证的是请求捕获及路由/配置解析；没有把测试接收器当作真实模型来验证生成质量。

## 原因

[resolveUpstream() 的判断](https://github.com/ranxianglei/billion-context/blob/020d45577a40f871299e3ed98a01c65341b3b51a/src/server.ts#L301-L313)：

```ts
const u = new URL(reqUrl);
const ownHost = (req?.headers.host ?? "").toLowerCase();
if (u.host.toLowerCase() !== ownHost) {
    return { upstream: `${u.protocol}//${u.host}`, rewrittenUrl: reqUrl, tunnel: true };
}
```

正常正向代理请求的 URL authority 与 Host 相等，因此被误判为自身请求，返回 undefined。这段判断由 94136fff 引入；旧的 /bili/ 接入不经过该分支，所以切换接入方式后才暴露问题。

后续两个处理不一致：

- [配置解析](https://github.com/ranxianglei/billion-context/blob/020d45577a40f871299e3ed98a01c65341b3b51a/src/server.ts#L1034)使用 route?.rewrittenUrl，路由为空就无法匹配上游的模型窗口配置。
- [实际转发](https://github.com/ranxianglei/billion-context/blob/020d45577a40f871299e3ed98a01c65341b3b51a/src/server.ts#L2437-L2440)还有完整 URL 的备用处理，因此仍能发往上游。

结果是聊天继续工作，但窗口配置静默失效，压缩压力可能被低估，提醒或兜底触发推迟。本 issue 不主张这是所有未压缩现象的唯一原因。

## 期望行为

希望修正判断依据，同时保留防止代理自循环和管理面访问控制的原始目标，不是简单删除自身目标检查：

- 用实际监听端点或可信配置判断目标是否为代理自身，不将客户端提供的 Host 当作代理身份。
- 转发和上游配置匹配使用一致的目标解析结果。
- 真正的自身转发请求应被明确阻止，不能在路由检查后又经备用转发逻辑放行。
- 保留现有管理接口保护及 /bili/、CONNECT 接入行为；不要求客户端重新改写模型地址或恢复配置副本。

## 建议回归覆盖

1. 完整上游 URL + 上游 Host 的真实正向代理请求能正常转发。
2. 上述请求正确命中模型窗口 120000，而非全局 200000。
3. 同一上游与模型分别经正向代理和 /bili/ 接入，配置解析一致。
4. 真正指向代理自身的转发仍被阻止，管理接口保护不退化。

现有 absolute-form 路由测试使用的是代理地址作为 Host，建议补上真实传输中 Host 为上游地址的用例。

ranxianglei:
Fixes #562.

## What changed

`resolveUpstream()` (`src/server.ts`) decided whether an absolute-form request targeted the proxy *itself* by comparing the URL authority against the **client-provided `Host` header**. In a real HTTP forward proxy the client sets `Host` to the **upstream** (== the URL authority), so every legitimate forward-proxy request was misclassified as a self-request and dropped to `undefined`. Consequences:

- Config resolution (`embeddedUrl = route?.rewrittenUrl`) received `undefined` → the per-upstream model-window config was silently missed (fell back to the global window).
- The fallback in `buildForwardTarget` still forwarded via the full URL → chat kept working while the window config did not take effect.

## Fix

Every well-formed absolute-form request is now marked a tunnel and routed through `checkTunnelDestination`, whose self-layer (destination port == our bound port AND a local-machine IP) 403s a genuine self-forward **before any forwarding** — no silent fall-through to the fallback path. Self-detection now uses the actual listening endpoint instead of the client-provided `Host` header. This preserves the original goals (prevent self-loop + keep the management plane unreachable through the tunnel) while fixing the misclassification. `/bili/`, CONNECT/MITM, and the admin gate are unchanged.

## Regression coverage (mapped to the issue's list)

1. A real forward-proxy request (full upstream URL + upstream `Host`) forwards correctly — the existing `forward-absolute-url` test now exercises the previously-broken branch (before, it only passed via the fallback forward).
2. New unit test: that request resolves the per-upstream window (120000) via `resolveConfiguredContextLimit` on the resolved `rewrittenUrl`, not the global default.
3. New unit test: the forward-proxy and `/bili/` forms of the same upstream resolve to the *same* window.
4. New integration test: a self-targeting absolute URL is denied with **403** (`tunnel_destination_denied`) instead of being forwarded; admin/tunnel protection does not regress (self-targets still carry the tunnel marker + admission gate).

## Pre-flight

- `npm run typecheck` — pass
- `npm test` — 1049/1050. The single failure (`tests/launcher.test.ts` → `resolveClientCommand: codex/claude resolve to themselves`) is **pre-existing and environment-specific**: this sandbox has a real `codex` binary at `/usr/bin/codex`, so `which codex` resolves to the absolute path rather than the bare name the test expects. Verified it fails identically with these changes stashed. Unrelated to this change.
- `npm run build` — pass
- E2E (`tests/e2e/e2e-codex.test.ts`) not run here: it requires a live local sglang upstream at `127.0.0.1:8199`, which is outside this sandbox's permitted network boundary. Note the e2e drives codex via origin-form / `/bili/` requests, which do not exercise the forward-proxy (absolute-form) branch touched here; that branch is covered directly by the new unit + integration tests above.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -295,17 +295,23 @@ export function resolveUpstream(_opts: ProxyOptions, reqUrl: string, req?: http.
     // sent this request in absolute form (`GET http://host/path HTTP/1.1`) —
     // exactly what httpx emits for plain-http base URLs through a proxy. Route
     // it like the /bili/ embedded form: the absolute URL IS the upstream, same
-    // tunnel semantics. A target equal to this request's own Host header means
-    // the client is asking the proxy to fetch ITSELF — forwarding that would
-    // loop the request straight back into handle(), so fall through (own-API
-    // routing) instead.
+    // tunnel semantics.
+    //
+    // #562: do NOT decide "is this the proxy itself?" from the client-provided
+    // Host header. In a real forward proxy the client sets Host to the UPSTREAM
+    // (== the URL authority), so comparing u.host against req.headers.host
+    // misclassified every legitimate forward-proxy request as a self-request
+    // and dropped it to `undefined` — losing the per-upstream context-window
+    // config (route?.rewrittenUrl undefined) while the fallback forward still
+    // reached the upstream (chat kept working, the window silently didn't).
+    // Self-targeting is decided by the ACTUAL listening endpoint instead: mark
+    // this a tunnel and let checkTunnelDestination's self-layer (destination
+    // port == our bound port AND a local-machine IP) 403 a genuine self-forward
+    // before any forwarding — no silent fall-through to the fallback path.
     if (reqUrl.startsWith("http://") || reqUrl.startsWith("https://")) {
         try {
             const u = new URL(reqUrl);
-            const ownHost = (req?.headers.host ?? "").toLowerCase();
-            if (u.host.toLowerCase() !== ownHost) {
-                return { upstream: `${u.protocol}//${u.host}`, rewrittenUrl: reqUrl, tunnel: true };
-            }
+            return { upstream: `${u.protocol}//${u.host}`, rewrittenUrl: reqUrl, tunnel: true };
         } catch {
             // malformed absolute URL
         }
--- a/src/tunnel-guard.ts
+++ b/src/tunnel-guard.ts
@@ -44,15 +44,30 @@ export function parseIpLiteral(s: string): string | null {
     return null;
 }
 
+/** IPv4-mapped v6 in ANY form → the plain dotted quad: both the dotted
+ *  `::ffff:1.2.3.4` and the HEX form `::ffff:102:304` the WHATWG URL parser
+ *  canonicalizes hostnames to (`new URL("http://[::ffff:127.0.0.1]/")` yields
+ *  `[::ffff:7f00:1]`), which otherwise sails past every dotted-only matcher.
+ *  Range classification and self-membership then see one shape. */
+export function normalizeIpLiteral(ip: string): string {
+    const t = ip.trim().toLowerCase();
+    if (!t.startsWith("::ffff:")) return t;
+    const dotted = t.match(/^::ffff:(\d{1,3}(?:\.\d{1,3}){3})$/);
+    if (dotted) return dotted[1];
+    const hex = t.match(/^::ffff:0*([0-9a-f]{1,4}):0*([0-9a-f]{1,4})$/);
+    if (hex) {
+        const hi = parseInt(hex[1], 16);
+        const lo = parseInt(hex[2], 16);
+        return `${(hi >> 8) & 255}.${hi & 255}.${(lo >> 8) & 255}.${lo & 255}`;
+    }
+    return t;
+}
+
 export function classifyIp(ip: string): IpClass {
-    const lit = parseIpLiteral(ip);
+    const lit = normalizeIpLiteral(parseIpLiteral(ip) ?? "");
     if (!lit) return "public";
-    // v6 (incl. ::ffff:a.b.c.d mapped — normalize BEFORE the v4 branch, which
-    // would otherwise split on the dots of a mapped literal and NaN out).
     if (lit.includes(":")) {
         if (lit === "::1" || lit === "::") return "loopback";
-        const mapped = lit.match(/^::ffff:(\d+\.\d+\.\d+\.\d+)$/);
-        if (mapped) return classifyIp(mapped[1]);
         if (lit.startsWith("fe8") || lit.startsWith("fe9") || lit.startsWith("fea") || lit.startsWith("feb")) return "linkLocal";
         if (lit.startsWith("fc") || lit.startsWith("fd")) return "private"; // fc00::/7 ULA
         return "public";
@@ -122,7 +137,7 @@ export async function checkTunnelDestination(origin: string, ctx: TunnelCheckCon
     let ips: string[];
     const literal = parseIpLiteral(host);
     if (literal) {
-        ips = [literal];
+        ips = [normalizeIpLiteral(literal)];
     } else {
         try {
             ips = await (ctx.resolveHost ?? dnsResolveHost)(host);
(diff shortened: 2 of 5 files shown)


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 Reviewed #563 (fix for the forward-proxy self-misclassification). **Verdict: LGTM — correct, root-cause, and well-covered.** No blocking issues.

## Triage (bug verification)

**Bug is real — confirmed at code level.** In base `resolveUpstream()` (`src/server.ts`), the forward-proxy branch compared `u.host` against the client-provided `Host` header:

```js
const ownHost = (req?.headers.host ?? "").toLowerCase();
if (u.host.toLowerCase() !== ownHost) { return { …, tunnel: true }; }
// else → fall through to `return undefined`
```

In a real HTTP forward proxy the client sets `Host` to the **upstream** (== the URL authority), so `u.host === ownHost` is true for every legitimate request → the guard never fired → route dropped to `undefined`. Downstream:
- `embeddedUrl = route?.rewrittenUrl` → `undefined` (`src/server.ts:1040`)
- `resolveConfiguredContextLimit(routes, undefined, model)` bails immediately at `src/config.ts:205` (`if (!model || !upstreamUrl) return undefined`) → per-upstream window silently fell back to global/registry.
- Meanwhile `buildForwardTarget`'s fallback (`isAbsoluteUrl ? reqUrl : …`, `src/server.ts:2445`) still forwarded via the full URL → chat kept working while config didn't apply. Exactly as described.

**Layer:** the diagnosis hits the root cause, not a symptom. Using a *client-supplied* header as a self-detection signal is fundamentally unsound for a forward proxy (the header IS the upstream). The fix removes that signal and decides self-ness from the actual listening endpoint via `checkTunnelDestination` — the right layer.

## Solution assessment

Correct and strictly safer than the status quo:
- Self-targets are now marked `tunnel: true` and 403'd by `checkTunnelDestination` Layer 1 (dest port == bound port **and** local-machine IP, `src/tunnel-guard.ts:136-141`) **before any forwarding** — no silent fall-through.
- As a side effect this also stamps `x-bili-tunnel` on *all* absolute-form forwards (`src/server.ts:2462`). Under the old bug those carried no marker (route was `undefined`), so the management-plane-through-tunnel guarantee was actually weaker than intended; the fix restores it for chained/NAT-hairpin cases too.
- `/bili/`, CONNECT/MITM, and the admin gate are untouched. Single call site of `resolveUpstream` (`handle`, `src/server.ts:922`), so no other consumer depends on the old "self→undefined" behavior.

## Pre-flight (ran locally on the PR head)

- `npm run typecheck` — **pass**
- `npm test` — **1049/1050**. The one failure, `tests/launcher.test.ts → resolveClientCommand: codex/claude resolve to themselves`, is **pre-existing & environment-specific**: I stashed the PR's three changed files and re-ran `tests/launcher.test.ts` on the clean base — it fails identically there (this sandbox has a `codex` binary resolvable on PATH, so `which` returns an absolute path instead of the bare name the test expects). Unrelated to this change. Confirmed your note.
- `npm run build` — **pass** (`dist/index.js` produced, acp-kernel inlined).
- New tests pass individually: both `tests/forward-absolute-url.test.ts` cases (forward correctness; self-target → **403 + `tunnel_destination_denied`**, which is only emitted at `src/server.ts:936` so it definitively proves the tunnel self-guard fired) and the new `upstream-proxy-routing` assertions (per-upstream window 120000 resolved off `rewrittenUrl`; forward-proxy vs `/bili/` parity).
- E2E not run here — needs a live local sglang upstream outside this sandbox's network boundary. Agreed with your reasoning: e2e drives origin-form / `/bili/` traffic, not the absolute-form branch touched here, so direct unit+integration coverage is the appropriate scope.

## Minor notes (non-blocking)

- The fix restores **more than just the window**: every per-upstream lookup keys off `route?.rewrittenUrl` (proxy decision at `src/server.ts:2483`, `resolveCompressProtocol`, compress-settings). Forward-proxy requests therefore regain their full per-upstream config, not only the context window. Worth a line in the PR body / changelog since it broadens the impact beyond "window config".
- Self-guard depends on `req.socket.localPort` being the bound port (`src/server.ts:929`) — correct for a plain `http.Server`; fine as-is.

Recommendation: merge-ready. Per repo rules I can't merge PRs (human-only) — please merge when you're ready: https://github.com/ranxianglei/billion-context/pull/563

PR #565, opened 2026-09-05
fix: keep Responses assistant history valid after compression (#564)

### LINKED ISSUE #564: Compression breaks Responses assistant history ordering, causing repeated invalid_assistant_history 400s
## 现象

通过 bili 使用 Responses-compatible upstream 时，压缩成功后，内部 re-request 或后续历史重放可能返回 `400 invalid_assistant_history`。实际遇到过两种错误：

```text
input Item 9 reasoning cannot follow another reasoning Item; supported assistant Item order is reasoning, message content, then function calls

input Item 56 reasoning cannot follow assistant message content
```

当重试仍使用同一份损坏的 folded history 时会重复失败；并非所有压缩都会触发，也不意味着会话永久不可恢复。另一次改变折叠结果的压缩后，后续请求曾恢复成功。

## 版本与范围

- 对照主分支 `020d455`（v0.1.86），使用 `acp-kernel@0.0.53`。
- 真实错误来自本地集成分支，并非声称在未经修改的发布包上完成了全部复现。
- 问题涉及 kernel fold 后的 assistant 分组，以及 `src/server.ts` 的 `stripKernelSummaries()` / `refreshFolded()` 调用顺序。
- 与 #539 不同：这里已有的 Responses reasoning items 仍然存在，损坏的是它们与 assistant 文本、工具调用之间的顺序，不是新生成的 `reasoning_content` 字段被遗漏。

## 路径一：折叠工具交互后，原本独立的 assistant 轮次合并

示意输入，所有标识和内容均为最小测试数据：

```text
user
reasoning A
assistant text A
function_call bash_a
function_call_output bash_a
reasoning B
assistant text B
function_call bash_b
function_call_output bash_b
```

当 reasoning / assistant text 被保护，而范围中的 bash 调用可以压缩时，工具调用被移除、孤立工具结果被清理，原本隔开的 assistant 轮次可能直接相邻：

```text
reasoning A -> assistant text A -> reasoning B -> assistant text B
```

另一个变体是 reasoning 所属轮次的正文及工具调用全部消失，只留下 reasoning；折叠后可以出现相邻的 reasoning items。原始合法历史因此变成上游拒绝的输入。

## 路径二：摘要去重再次删除轮次分隔

即使在 `core.processTurn()` 返回时已经维持了轮次分隔，后续去重仍可能破坏它。

实际定位到的折叠尾部结构：

```text
reasoning A
assistant text A
acp_summary_<block>
reasoning B
assistant text B
```

`refreshFolded()` 随后追加当前 compress loop 的记录。当 active block 的 `compressCallId` 对应调用已存在时，`stripKernelSummaries()` 删除重复的 `acp_summary_<block>`。这是避免重复摘要所需要的处理，但这个摘要恰好也是两轮 assistant 之间唯一剩余的分隔。

去重后发往上游的序列变成：

```text
reasoning A
assistant text A
reasoning B       <- reasoning follows assistant content
assistant text B
...
compress call / result
```

因此只在 kernel fold 后修复还不够；需要检查摘要去重之后的最终输入。位于尾部的 compress call/result 不能隔开更早的两组 assistant items。

## 最小服务端复现步骤

1. 使用上述两轮 Responses 输入，reasoning items 带固定的测试 `encrypted_content`。
2. 启动真实 bili server 和本地 mock upstream；设置 `preserveRecentMessages = 20`，测试中将 `minCompressRange` / `minSummaryLength` 设为 0。
3. mock upstream 的第一轮流式返回 `compress` 调用，范围从第一条 bash call 到第二条 bash call，摘要为 `Inspected files A and B.`。
4. 捕获 bili 实际发出的内部第二轮 HTTP 请求，而不是只检查 kernel 返回值。
5. 检查最终 Responses input 中 reasoning、assistant content、function calls 的顺序。

本地回归在补上摘要去重后的修复前，确定性失败于：

```text
AssertionError [ERR_ASSERTION]: reasoning rs_b follows assistant content
```

## 本地修复方向及验证

本地使用原始历史记录 assistant 分组，而不是仅凭折叠结果猜测：

- 只有当某 reasoning 原有的正文 / 工具调用全部被折叠移除时，才移除这个失去关联内容的 reasoning；原本 reasoning-only 的消息不因此删除。
- 被压缩掉的非 assistant 消息原本隔开的两组 assistant 内容，折叠后相邻时，插入说明中间交互已被压缩的 user-role 消息，保留轮次区别。
- 在 `stripKernelSummaries()` 完成后再次恢复这些分隔，包括普通 prepare 路径和内部 re-request。保留 reasoning items 的原始内容及 `encrypted_content`，不合并或改写签名内容，也不修改原始历史与压缩块。

这是本地可工作的处理方式，最终实现可放在合适的 kernel / host 层；关键是摘要去重之后仍然保持原始 assistant 轮次关系。

本地新增回归覆盖：

- kernel fold 及持久化重放中的 detached reasoning / assistant 分组。
- 真实 server 执行 compress 后的内部 re-request。
- 不携带 compress call 的普通后续请求，保留摘要载体。
- 携带 compress call 的 plugin 请求，去重摘要后仍保持有效顺序。
- 工具调用与结果成对移除，compress call/result 保留，原始输入和保留的签名 reasoning 不变。

本地执行：

```sh
node --import tsx --test tests/compression-reasoning-pairs.test.ts tests/compression-summary-boundaries.test.ts
```

结果：12 passed，0 failed。这两个测试文件目前属于本地修复，尚未提交到本仓库。上述服务端回归使用 mock provider；不将其等同于修复后的真实上游端到端验收。

## 期望行为

压缩和摘要去重不能把原本合法、独立的 assistant 轮次变成同一轮非法历史。应针对最终发往上游的请求增加回归，而不只验证去重之前的 kernel fold 输出。

ranxianglei:
## What

Fixes #564: after a successful compression, bili could send an **invalid Responses assistant history** upstream, producing repeated `400 invalid_assistant_history`:

```
input Item 9 reasoning cannot follow another reasoning Item; supported assistant Item order is reasoning, message content, then function calls
input Item 56 reasoning cannot follow assistant message content
```

## Root cause

Responses item order equals CoreMessage order (`coreToResponses` maps each core message to an item in sequence). Two original assistant turns were being merged into one run by two steps:

1. folding removes the tool calls/results that separated two turns, and
2. `stripKernelSummaries()` then deletes the `acp_summary_<block>` that was the last remaining separator (the block's compress call is present in history, so the summary is redundant and gets stripped).

Result: `[reasoningA][messageA][reasoningB][messageB]` — a single run violating the Responses invariant (within a run the order must be `reasoning* -> message* -> function_call*`, with at most one reasoning item). This is Responses-specific: OpenAI chat merges consecutive assistant messages into one object, so it never surfaces there.

Distinct from #539 (which dropped the `reasoning_content` field on re-request): here the existing reasoning *items* survive, but their order relative to assistant text/tool-calls is corrupted.

## Fix

New `repairResponsesAssistantOrdering(folded, original)` in `src/server.ts`, applied **after** summary dedup at all three Responses wire-build sites:

- `prepareResponses` (initial request / plugin-mode replay)
- `prepareResponsesCompact` (forged compacted history handed back to the client)
- `refreshFolded` (post-compress internal re-request)

It rebuilds turn boundaries from the **original client history**:

- drops a reasoning whose turn's non-reasoning body was wholly folded away (its content now lives in the block summary), while keeping originally-reasoning-only turns' reasoning;
- otherwise inserts a neutral user-role marker between the merged runs.

Signed reasoning items are preserved verbatim (`encrypted_content` untouched); the original history and compression blocks are never mutated.

## Tests

`tests/issue564-responses-ordering.test.ts`:

- end-to-end: drives the real `runCompressLoop` + responses adapter through a live compress and asserts on the **actual re-request body** (captured via a fetch probe) — not just the pre-dedup fold output;
- unit: orphaned-reasoning drop, reasoning-only-turn retention (runs stay separated), and surviving-`acp_summary`-carrier preservation (no spurious marker injected).

## Pre-flight

- `npm run typecheck` — clean
- `npm test` — 1051 pass / 1 fail. The single failure (`launcher.test.ts`: `resolveClientCommand: codex/claude resolve to themselves`) is **pre-existing and environmental**: this checkout has `/usr/bin/codex` installed, so resolution returns the full path where the test expects the bare name. Reproduced identically on pristine master; unrelated to this change.
- `npm run build` — clean

Note: the real-codex-through-bili E2E suite (`tests/e2e/e2e-codex.test.ts`) requires a local Responses-compatible upstream plus the codex binary and could not be executed in this sandbox (loopback upstreams are blocked by network policy); the new regression pins the final wire ordering directly instead.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -1572,7 +1572,7 @@ const ACP_TAG_MARK = "\x3cacp ";
 // nonexistent (preflight), so acp_summary survives as the carrier and
 // systemToUser later re-voices the survivors as USER messages (leaving them at
 // their anchors) for strict backends (#377).
-function stripKernelSummaries(messages: BiliMessage[], state: CompressionState): BiliMessage[] {
+export function stripKernelSummaries(messages: BiliMessage[], state: CompressionState): BiliMessage[] {
     const carried = new Set<string>();
     for (const b of state.blocks) {
         if (!b.active || !b.compressCallId) continue;
@@ -1581,6 +1581,62 @@ function stripKernelSummaries(messages: BiliMessage[], state: CompressionState):
         }
     }
     return messages.filter((m) => !(m.id ?? "").startsWith("acp_summary_") || !carried.has(m.id));
+}
+
+// #564: folding + stripKernelSummaries can merge two assistant turns into one
+// run, which Responses rejects (run order reasoning* -> message* ->
+// function_call*, <=1 reasoning). Rebuild boundaries from the ORIGINAL history
+// AFTER dedup: drop a reasoning whose turn body was wholly folded away (keep
+// originally-reasoning-only turns), else separate the runs with a user marker.
+const RESPONSES_TURN_SEPARATOR = "[The exchange between these two assistant turns was compressed.]";
+
+export function repairResponsesAssistantOrdering(folded: CoreMessage[], original: CoreMessage[]): CoreMessage[] {
+    const runOf = new Map<string, number>();
+    const runHasBody = new Map<number, boolean>();
+    let run = 0;
+    let inRun = false;
+    for (const m of original) {
+        if (m.role === "assistant") {
+            if (!inRun) { run++; inRun = true; }
+            runOf.set(m.id, run);
+            if (m.contentType !== "reasoning") runHasBody.set(run, true);
+        } else {
+            inRun = false;
+        }
+    }
+    const survivorCount = new Map<number, number>();
+    for (const m of folded) {
+        const r = m.role === "assistant" ? runOf.get(m.id) : undefined;
+        if (r !== undefined) survivorCount.set(r, (survivorCount.get(r) ?? 0) + 1);
+    }
+
+    const out: CoreMessage[] = [];
+    let phase = -1;
+    let seenReasoning = false;
+    let sepSeq = 0;
+    const pushSeparator = (): void => {
+        sepSeq++;
+        out.push({ id: `acp_turn_sep_${sepSeq}`, role: "user", contentType: "text", text: RESPONSES_TURN_SEPARATOR });
+        phase = -1;
+        seenReasoning = false;
+    };
+    for (const m of folded) {
+        if (m.role !== "assistant") {
+            out.push(m);
+            phase = -1;
+            seenReasoning = false;
+            continue;
+        }
+        const kind = m.contentType === "reasoning" ? "reasoning" : m.contentType === "tool-call" ? "tool-call" : "message";
+        const r = runOf.get(m.id);
+        if (kind === "reasoning" && r !== undefined && runHasBody.get(r) && survivorCount.get(r) === 1) continue;
+        if ((kind === "reasoning" && (phase > 0 || seenReasoning)) || (kind === "message" && phase === 2)) pushSeparator();
+        out.push(m);
+        if (kind === "reasoning") { phase = Math.max(phase, 0); seenReasoning = true; }
+        else if (kind === "message") phase = Math.max(phase, 1);
+        else phase = Math.max(phase, 2);
+    }
+    return out;
 }
 
 function diagTagSummary(messages: CoreMessage[], sessionId: string, strategy: string): string {
@@ -2024,7 +2080,7 @@ function prepareResponses(
         log("info", diagTagSummary(turn.messages, sessionId, "text-only"));
         const willInjectNudge = opts.compress.injectNudge && !!turn.nudge && shouldInject && !isCompactionTrigger && (turn.nudge.shouldInject || emergencyNudge(turn.nudge));
         log("info", diagNudge(turn, sessionId, tokenCount, config.modelContextLimit, parsed.model, willInjectNudge));
-        processedMessages = stripKernelSummaries(turn.messages, turn.state);
+        processedMessages = repairResponsesAssistantOrdering(stripKernelSummaries(turn.messages, turn.state), originalMessages);
         reapOrphanBlocks(session, msgs, deactivateBlock);
         rebuiltInput = patchResponsesInput(projection, processedMessages);
         // Fallback path: when the echo did NOT come back this turn (client
@@ -2271,7 +2327,7 @@ function prepareResponsesCompact(
             session.state = prevState;
             return base;
         }
-        const processed = stripKernelSummaries(turn.messages, turn.state);
+        const processed = repairResponsesAssistantOrdering(stripKernelSummaries(turn.messages, turn.state), projection.msgs);
         const output = patchResponsesInput(projection, processed);
         if (typeof output === "string") {
             session.state = prevState;
@@ -3074,7 +3130,7 @@ async function forward(
                 });
                 prepared.session.state = turn.state;
                 const records = current.filter((m) => typeof m.id === "string" && m.id.startsWith("acp_loop_"));
-                return stripKernelSummaries([...turn.messages, ...records] as BiliMessage[], turn.state) as CoreMessage[];
+                return repairResponsesAssistantOrdering(stripKernelSummaries([...turn.messages, ...records] as BiliMessage[], turn.state), prepared.originalMessages);
             };
             const loop = runCompressLoop(
                 streamToRead,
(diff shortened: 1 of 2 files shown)


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 **Review of PR #565** (fixes #564)

Reviewed the diff (`src/server.ts` + `tests/issue564-responses-ordering.test.ts`) and independently reproduced the pre-flight on the PR commit (`3a5df63`). **Verdict: LGTM to merge** — correct root-cause fix, well-tested, clean build. One non-blocking hardening note at the end.

**Duplicate screen:** no other open/closed issue or PR targets this (searched by symptom + fix). Confirmed distinct from #539 as you framed it — #539 dropped the `reasoning_content` *field* on replay; here the reasoning *items* survive but their order relative to assistant text/tool-calls is corrupted.

**Root cause — verified.** Responses enforces per-run order `reasoning* -> message* -> function_call*` with ≤1 reasoning item. `coreToResponses` maps CoreMessage→item in sequence, so item order == core order. Folding removes the tool calls/results that separated two assistant turns; `stripKernelSummaries()` then deletes the `acp_summary_<block>` that was the last remaining separator (its compress call is present in history → redundant). Net: one run `[reasoningA][messageA][reasoningB][messageB]` → repeated `400 invalid_assistant_history`. OpenAI chat never surfaces it because it merges consecutive assistant messages into one object. Diagnosis is accurate.

**Fix — sound.** `repairResponsesAssistantOrdering(folded, original)` rebuilds turn boundaries from the *original* (pre-fold) client history, applied **after** dedup:
- drops a reasoning whose turn's non-reasoning body was wholly folded away (content now lives in the block summary);
- keeps originally-reasoning-only turns' reasoning;
- otherwise inserts a neutral user-role marker between the merged runs.

Signed reasoning is preserved verbatim (`encrypted_content` untouched); inputs are never mutated. Applied after `stripKernelSummaries` at all three wire-build sites:
- `prepareResponses` — src/server.ts:2083
- `prepareResponsesCompact` — src/server.ts:2330
- `refreshFolded` in `forward` — src/server.ts:3133

I hand-traced the drop/separator logic against the orphaned-reasoning, reasoning-only-turn, and surviving-carrier cases — all consistent. Importantly, a separator is only ever inserted directly *after* assistant content (the phase/seenReasoning flags reset on every non-assistant item), so it cannot create two consecutive same-role items.

**Both compression modes:** the repair is carrier-independent (pure CoreMessage transform) and covers both plugin-mode replay and proxy-mode re-request. ✓

**Pre-flight (reproduced independently on `3a5df63`):**
- `npm run typecheck` — clean
- `npm run build` — clean
- `npm test` — **1051 pass / 1 fail**. The single failure is `launcher.test.ts › resolveClientCommand: codex/claude resolve to themselves`; I reproduced it identically on pristine master (`2cbbeab`). It's pre-existing/environmental (a CLI binary resolving to its full path where the test expects the bare name) and has zero code-path overlap with this change.

**Tests:** strong. The e2e case drives the real `runCompressLoop` + responses adapter through a live compress and asserts on the **actual re-request body** (captured via a fetch probe), not just the pre-dedup fold — plus unit coverage for orphan-drop, reasoning-only retention, and surviving-`acp_summary` no-spurious-marker. This pins the final wire ordering directly.

**Non-blocking observation (optional hardening):** `refreshFolded` is the shared loop path (`src/loop/core.ts` calls it for every protocol after a successful compress), so `repairResponsesAssistantOrdering` also runs for OpenAI/Anthropic re-requests. I verified it stays structurally safe there (separator only lands after assistant content → `assistant/user/assistant`, valid alternation, never two consecutive users), but it injects a spurious user turn and breaks the prefix cache for protocols that never had this bug. Two options:
1. Gate the refresh-path call to `protocol === "responses"` for minimal blast radius, or
2. Keep it universal if you want stricter turn-fidelity across all protocols (also defensible — presenting two distinct turns as one is arguably wrong everywhere).

My lean is **option 1**: it matches the actual bug scope and avoids introducing untested behavior against real codex (the E2E suite couldn't run here). Not a blocker either way.

**E2E note:** the real-codex-through-bili suite needs a local Responses-compatible upstream + the codex binary and could not be executed in this sandbox (loopback upstreams blocked); the fetch-probe regression pins the final ordering instead, which is adequate. Recommend running the manual-dispatch `ci-e2e.yml` post-merge for full confidence.

Per repo policy I won't merge (human-only). Happy to implement the responses-gating follow-up if you'd like it.

Issue #570, opened 2026-09-06
Learned context window (self-heal) treats any truncated upstream stream as a window overflow: false positives permanently shrink the session window, persisting across restarts


shengbuding:
## Environment

- billion-context v0.1.86 (npm), proxy mode; upstream = local llama.cpp server (llama-server) with a real 150,528-token window; client = dsh CLI; sessions also run concurrently against the same llama server (two slots, shared KV pool).

## Symptom

The self-heal mechanism learned a *wrong* conservative window for the local model: `140000 鈫?121815`, and it sticks (re-applied on every request, persisted in the session store; visible across restarts):

```
03:48:38 [<session-a>] weak overflow signal 1/3 (usage 96%, round 1: upstream stream truncated without a completion event)
03:53:45 [<session-a>] weak overflow signal 2/3 (usage 96%, round 1: upstream stream truncated without a completion event)
03:58:52 [<session-a>] weak overflow confirmed (3脳 high-usage failures 鈥? 鈥?learned conservative window 121815 for qwen3.8-27b-abliterated-q4_k_m-9fbdc3cd (was unset); arming emergency shrink
03:58:54 [<session-a>] self-healed context window: 140000 鈫?121815 (learned from an upstream overflow)
04:04:05 [<session-a>] self-healed context window: 140000 鈫?121815 (鈥?
04:16:12 / 04:16:24 / 04:26:23 / 04:39:10 [<session-a>] self-healed context window: 140000 鈫?121815 (鈥?   鈫?every request
```

Root cause of the *truncations* here was NOT the model window: with two ~100K-token sessions pre-filling concurrently, llama.cpp's shared KV pool ran out (`failed to find free space in the KV cache`, `W srv try_clear_id: purging slot 0 with 112764 tokens`, `failed to find N available cells`). The stream was truncated by the *server's KV pressure*, but the proxy counted it as "high-usage failure 鈫?overflow" and shrank the learned window 14% below the true window (121,815 vs real 150,528). Once learned, the window is min()'d into every future request config for that model, throttling the session below its true capacity; and since it is persisted with the session, restart does not clear it (a manual edit of the session store was needed to unstick it).

Not a one-off: earlier sessions had learned windows 115,951 / 91,575 / 99,032 for the same model under the same KV-contention conditions, all below the real 150,528.

## Observations (facts only, no proposed fix)

[...]

PR #572, opened 2026-09-06
fix: separate confirmed overflow evidence from weak truncation guesses in window self-heal (#570)

ranxianglei:
## Problem (#570)

The weak-overflow heuristic (3 high-usage stream truncations in 15 min) and the strong path (a real non-2xx overflow rejection) both wrote into the same per-model `learnedContextLimits`. Any mid-stream death — llama.cpp shared-KV exhaustion (`failed to find free space in the KV cache`), OOM, network reset — looked like an overflow, so one misjudged session permanently shrank the model's window for every later session, across restarts, with no correction path: the only upward self-heal is gated on fallback windows, never on the authoritative configured value.

## Fix (three parts)

1. **Provenance split.** Strong evidence (an actual non-2xx overflow rejection, incl. the payload-size upper bound) now lands in `metadata.confirmedContextLimits` / `confirmedContextLimit`; the weak heuristic keeps writing only to `learnedContextLimits` / `learnedContextLimit`. Resolution order is confirmed > speculative, so a weak guess can no longer clobber a confirmed window.

2. **Retraction.** A successful turn whose upstream-reported input exceeds a learned window beyond a margin (`max(256 tokens, 3%)`) proves it stale — KV-pressure false positive, resized server, … — and removes it *before* this request's config is resolved. True overflows cannot be contradicted (a turn cannot succeed above the real window); the armed emergency value (== the learned value exactly) sits below the margin and never retracts itself.

3. **Ground truth for llama.cpp.** `inspectContextOverflow` now recognizes `exceed_context_size_error` and parses the real limit from the `(N / M > W)` body shape, so a genuine overflow learns W instead of letting the weak path guess.

Both compression modes considered: plugin mode (agent-owned compression; proxy suppresses injection) and proxy mode — the changed paths are upstream-facing (error inspection, per-request limit resolution, session metadata) and identical on both; no wire-format change.

## Pre-flight

- `npm run typecheck` — clean
- `npm test` — 1056/1057 pass; the single failure (`launcher.test.ts: resolveClientCommand codex/claude resolve to themselves`) is pre-existing on clean master in this environment (sandbox stub at `/usr/bin/codex` makes the resolver return an absolute path) and unrelated to this change
- `npm run build` — success
- New unit coverage: provenance precedence (confirmed governs; weak can't clobber), retraction (stale removed, margin respected, armed no-op, confirmed retracts too, other models untouched), llama.cpp body parsing → window 150528

Note: the real-client E2E suite (`tests/e2e/`, needs `ACP_TEST_E2E=1` + live Responses upstream) was not runnable in this sandbox; the local e2e-style self-heal tests (`tests/e2e-overflow-selfheal.test.ts`, `issue-393`, `issue496`) all pass.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -71,6 +71,7 @@ import { consumePluginRegisterFor, flushConversations, handlePluginCompact, hand
 import { setupMitm, readMitmUpstream } from "./mitm.js";
 import type { BiliMessage } from "acp-kernel/wire";
 import { backfillHostUsage, isLoopbackAddress, inspectContextOverflow, reserveOutputHeadroom, shouldReserveOutputHeadroom, systemToUser, usageTotals, type WireProtocol } from "./util.js";
+import { resolveConfirmedLimit, resolveLearnedLimit, resolveSpeculativeLimit, retractStaleLearnedLimits } from "./weak-overflow.js";
 import { BILI_TUNNEL_HEADER, checkTunnelDestination, tunnelAllowlistFromEnv } from "./tunnel-guard.js";
 
 import { decodeRequestBody } from "./content-encoding.js";
@@ -1390,10 +1391,10 @@ async function handle(
             // hammering the upstream). Gate on the raw body estimate and fail
             // fast locally instead of forwarding (#301 precedent).
             const reqModel = (parsed as { model?: string }).model;
-            const learnedMap = session.metadata.learnedContextLimits as Record<string, number> | undefined;
-            const learnedLimit =
-                (reqModel && learnedMap ? learnedMap[reqModel] : undefined) ??
-                (session.metadata.learnedContextLimit as number | undefined);
+            // #572-merge: read through the resolver (confirmed → speculative, per-model
+            // then scalar) — the learner now writes the confirmed channel, so the
+            // legacy direct map read would miss windows learned from real 400s.
+            const learnedLimit = resolveLearnedLimit(session, reqModel);
             const guard = sideRequestGuard(parsed, protocol, reqConfig.modelContextLimit, learnedLimit);
             if (guard.blocked) {
                 log("warn", `[${session.id}] side request (~${guard.estimate} tokens) ≥ effective window ${guard.limit} (model=${reqModel ?? "?"}) — NOT forwarded: guaranteed upstream 400 (side requests bypass preflight by design, #388)`);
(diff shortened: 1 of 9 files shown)


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 **Review complete — preflight independently verified, one defect found in review and fixed on the branch.**

**Preflight (all three re-run by me):**
- `npm run typecheck` — clean ✓
- `npm test` — 1056/1057; the single failure (`launcher.test.ts: resolveClientCommand codex/claude resolve to themselves`) reproduces identically on clean master in this environment, so it's environmental, not from this change ✓
- `npm run build` — success ✓

**Design review:** provenance split, the retraction margin (`max(256 tokens, 3%)`), and the llama.cpp ground-truth parsing all check out against the code. One deliberate divergence worth noting (not a bug): `handle()` resolves limits in order cPM→cS→sPM→sS while `resolveLearnedLimit` uses cPM→sPM→cS→sS — the former drives the provenance log label, the latter drives `noteWeakOverflow`'s usage-% denominator where the tighter bound is preferred. Harmless; left as-is.

**Defect found in review (fixed):** retraction reads `session.stats.lastInputTokens`, which is a high-water mark that *failed* turns also write — `noteWeakOverflow` arms it at the failing input size, and `forward()`'s overflow floor takes `max(prev, floor)`. Consequence: three mid-stream deaths at 185k input (KV-pressure pattern) → mark at 185k; then a genuine parseable overflow learns W=150528 into `confirmedContextLimits`; on the next request, retraction sees 185000 − 150528 ≥ margin and **deletes the just-learned ground truth**, logging "a later turn succeeded at 185000" when nothing succeeded. That contradicts the PR's own stated invariant ("True overflows cannot be contradicted") — and a second variant (kills above an already-governing confirmed window) deletes a *correct* value. Impact is narrow and transient (one wasted overflow cycle + misleading log pair), so non-blocking — but it undermines the guarantee this PR exists to provide, so I fixed it on the branch:

1. `forward()`: when the overflow body parses a window, arm at **exactly** it — no turn can succeed above the real window, so any higher armed value provably came from a failed turn; unparsed cases keep the max-floor.
2. `noteWeakOverflow()`: while a confirmed window governs, cap the armed value at it — a mid-stream death above the real window is a non-window kill mode (KV pressure / OOM / reset), not evidence about the window.
3. Regression coverage: unit tests for both caps; a new e2e rig (three 370k KV-pressure deaths + a parsed 128k overflow, asserting the confirmed window survives the next request's retraction check — fails without the fix); tightened the existing armed-value assertion to exact equality.

Re-ran everything after the fix: typecheck clean, `npm test` 1058/1059 (same sole pre-existing failure), build success. Commit pushed to the PR branch. Merge remains yours: https://github.com/ranxianglei/billion-context/pull/572

ranxianglei:
已在 35d2809 将本 PR 语义合并到当前 master 并重写为线性单提交(修 #575 同样的 diff-爆炸问题)。全量 1146/1146 通过、typecheck 0 错、build 成功。语义合并要点:#570 weak-overflow:util.ts 取 master #554 的超集正则;server 的保守估计保留 raw-body 兜底但走 confirmed 通道命名;side-request guard 改读 resolveLearnedLimit(confirmed→speculative),老 #554 测试断言对齐 confirmed 通道。CI 绿后请人工合并。

ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 Verified the rewritten branch independently:

- **Linearity**: `35d2809` is a single linear commit on current master (parent `3f57ce4`) ✓
- **Guard survival**: the review-fix (exact arm at parsed window in `forward()`, `armInput` cap in `noteWeakOverflow()`, plus both regression tests incl. the e2e rig) is present in the semantic merge ✓
- **Preflight** (fresh `npm ci` — note the base now carries acp-kernel 0.0.54, stale deps initially surfaced two unrelated #386 CJK failures that cleared after install): typecheck clean ✓, full suite **1152/1153** — sole failure is the environmental `launcher.test.ts: resolveClientCommand codex/claude` sandbox-stub issue that reproduces identically on clean master here ✓, build success ✓

Everything checks out. On the merge itself: I can't merge PRs — system rules forbid it. Please merge yourself: https://github.com/ranxianglei/billion-context/pull/572

PR #467, opened 2026-09-02
fix(#453): hard backstop for plugin-mode context overflow

ranxianglei:
## What & why

#453: in plugin mode (`bili pi`) bili trusts the ACP agent to run `compress`. When a weak model ignores the nudge, the payload grows past the window and the upstream **400s on input+output** (vLLM rejects an oversized total instead of clamping) — then the client falls back to native compaction and loses all working state (compounds #14).

Proxy mode had `preflightCompressIfNeeded` as a backstop, but its trigger estimate (`estimateCoreMessages`) counts message text only — it **ignores system + tool definitions** and can lag by a turn — so it skipped exactly when the real input crossed the reserved limit. Repro numbers: real input 131745 + completion 131072 = 262817 > 262144 window; the trigger saw ~117k and stayed silent.

## Two host-side fixes (both modes)

**1. Clamp the outgoing output budget (the hard backstop).** In `prepareOpenai` / `prepareResponses`, cap `max_tokens` / `max_completion_tokens` / `max_output_tokens` to `window − estimated_input − margin`. The proxy owns both sides of the sum, so input+output can no longer exceed the window on request-rebuilding upstreams. `estimateInputTokens()` takes the max of the real previous-turn input and a fresh count that *includes* system + tools, closing the undershoot that let preflight skip. Anthropic is exempt (its input limit is independent of `max_tokens`). When the cap would drop below a floor (input alone nearly fills the window) it leaves the budget untouched — that's preflight/self-heal territory, not output starvation.

**2. Escalate the nudge (mitigation).** Force-inject every turn once usage ≥ `EMERGENCY_NUDGE_ESCALATION_PCT` (0.70) even when the kernel's cadence/growth gate stays silent. `diagNudge` now labels these `INJECT-ESC` instead of logging `idle`.

> **Note on the escalation threshold:** tracing the kernel's `decideNudge` showed the **pressure branch (usage ≥ `nudge.maxContextLimitPct`, default 0.75) has no cadence gate** — the kernel already force-injects every turn there. So the *only* cadence-silent zone is below 0.75, which is why the line sits at **0.70** (just under it) rather than the "≥85%" floated in the issue — 85% lands inside the zone where the kernel already injects, making it a no-op. It's a single named constant, trivially tunable. Please confirm you're happy with 0.70 (or whether you'd like it exposed across the three compress config levels).

## Tests

New `tests/issue453-backstop.test.ts`:
[...]

PR #576, opened 2026-09-06
fix: count system+tools wire overhead in the preflight trigger (#470)

### LINKED ISSUE #470: track: plugin-mode proxy-side emergency truncation (last-resort backstop, deferred from #467)
来源: #453 (https://github.com/ranxianglei/billion-context/issues/453) 分析时发现；owner 在 #453 分阶段方案确认中明确要求「另开 issue 单独跟踪」；PR #467 (https://github.com/ranxianglei/billion-context/pull/467) 有意未包含此项。

## Background

#453: in plugin mode (`bili pi`) bili trusts the ACP agent to run `compress`. When a weak model ignores the nudge, the payload grows past the window and the upstream 400s on input+output (vLLM rejects an oversized total instead of clamping) — then the client falls back to native compaction and loses all working state.

PR #467 ships two host-side backstops that cover the common path:

1. **Clamp the outgoing output budget** — `max_tokens`/`max_completion_tokens`/`max_output_tokens` capped to `window − estimated_input − margin`, so input+output can no longer exceed the window on request-rebuilding upstreams (both modes).
2. **Escalate the nudge** — force-inject every turn once usage ≥ `EMERGENCY_NUDGE_ESCALATION_PCT` (0.70), filling the sub-0.75 cadence-silent zone (kernel's pressure branch ≥0.75 already force-injects; verified against `decideNudge`).

This issue tracks the third line of defense **deliberately NOT included** in #467: **proxy-side emergency truncation in plugin mode**.

## Why it is deferred (carrier-consistency risk)

The existing preflight/self-heal truncation path builds tool-call-less blocks whose `acp_summary` is re-voiced as a **user message** by `systemToUser` — that is **proxy-mode semantics**, where the proxy owns compression and the ephemeral tool call never enters the client's history.

In plugin mode the contract is inverted: the ACP-native agent OWNS compression — the `compress` call + result live in the agent's own re-sent history, and the wire carrier is the **tool call** itself (proxy suppresses tool/nudge injection of its own; the agent's view never renders the kernel's `acp_summary`). If the proxy unilaterally compresses/truncates ranges the agent has not compressed:

- the agent's next re-send may contain the **unfolded original range** → the proxy's block state and the agent's view diverge (double-compress, orphan blocks, or a summary the agent never produced);
- summary carriers become inconsistent within one conversation (agent-produced tool-call summaries vs proxy-produced user-voiced `acp_summary`), which strict backends (e.g. SGLang's single-system-at-index-0 constraint, cf. #377) do not tolerate well.

So this needs a design pass before any implementation, e.g.:

- can the proxy restrict emergency truncation to ranges the agent provably cannot claim (already-blocked / protected-only zones)?
- or notify/reserve the ranges in the agent's ref space so its next re-send stays consistent?
- what happens to the agent's pending compress calls that targeted now-truncated refs?

## Expected shape (for discussion, not a commitment)

- Last resort only: engage when input alone nearly fills the window (the territory where the output-clamp floor no-ops) **and** preflight/self-heal did not already fire.
- Must preserve consistency between the agent's re-sent history and the proxy's block/state store.
- Strictly lower priority than #467's backstops: the clamp guarantees no more input+output 400s, the escalation nudge gives the agent repeated chances to self-compress; truncation is the final fallback for models that ignore both.

## References

- #453 — root problem, repro numbers, staged plan confirmed by owner
- #467 — PR implementing backstops 1+2 (this item split out per owner direction)

ranxianglei:
Closes #470 (residual gap from the analysis in the issue thread).

## Root cause

The preflight trigger and fit gates estimated the payload with `estimateCoreMessages` (messages only), but the billed input also carries the system prompt and tool definitions — the client's own (~10-20k on agent clients) plus the proxy-injected compress system prompt (~1.9k tokens) and ACP tools (~0.5k). Consequences:

- the trigger fired **late**: text alone "fit" the window while the real input already overflowed it (guaranteed upstream 400 on strict relays);
- the post-fold fit gate could pass on payloads whose **billed** input still exceeded the window.

#467 fixed this class of under-count only for the output clamp (`estimateInputTokens`); this PR applies the same term to preflight — the plugin-mode last resort.

## Change

- `src/server.ts`: new `estimateWireOverhead(protocol, body)` — counts system/instructions text (anthropic string/array forms, openai hoisted system/developer messages, responses instructions) + tools JSON from the rebuilt wire body, via `defaultCountTokens`.
- `preflightCompressIfNeeded`: trigger + both post-fold fit gates add the term.
- `src/preflight.ts`: `PreflightDeps.wireOverhead` — the fold loop's `currentTokens`/`payloadEstimate` floors include it, so the loop keeps folding until the true billed input fits. Without this the loop stopped at "text fits" and then fail-fasted against the caller's overhead-aware gate (502 on foldable payloads).
- `README.md`: the two-mode table's plugin-mode preflight cell said "No" — preflight runs unconditionally in both modes as the last-resort backstop; corrected.

## Test retuning (behavior change is intended)

- `tests/preflight-fail-fast.test.ts`: new e2e — text fits the 10k window alone, system+tools push the billed input over → preflight folds and forwards; forwarded body no longer carries the raw big message. The #330 e2e re-tuned (24k chars instead of 32k): its old payload was forwarded while the real billed input exceeded the window.
- `tests/issue453-backstop.test.ts`: the clamp e2e re-sized (window 400k / max_tokens 150k) to sit in the band where the clamp acts alone — the old shape now correctly triggers preflight (fold instead of output starvation), which is the better outcome.
- `tests/wire-overhead.test.ts`: 6 unit tests for the helper (protocol matrix, unparseable body → 0).

## Pre-flight

- `npm run typecheck` ✅
- `npm test` 1070/1072 — the only failures are the 2 known env-dependent ones in `plugin-agent.test.ts` (fail locally on this machine, pass in CI) ✅
- `npm run build` ✅

## Out of scope (per the issue analysis)

- The optional stateless request-scoped drop (deep-tail plugin-mode backstop) — not needed while the trigger is honest; can be revisited if real-world reports show inputs that fit preflight's accounting but still 400.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ The proxy runs in one of two modes, and **the mode decides who executes
 | Client | ACP-native agent with the bili extension (pi/omp) | Any OpenAI/Anthropic client, no extension |
 | Who executes `compress` | **The agent** (pi runs it locally) | **The proxy** (server-side compress loop) |
 | `compress` tool call in the re-sent history? | Yes — part of the agent's own conversation | No — ephemeral proxy-loop traffic |
-| Preflight blocks (no tool call)? | No — overflow forces the model to call `compress` | Yes — `src/preflight.ts` compresses behind the client's back |
+| Preflight blocks (no tool call)? | Last-resort backstop — the agent normally compresses on its own `compress` calls, but `src/preflight.ts` still fires (in both modes) when the input alone exceeds the window (#470) | Yes — `src/preflight.ts` compresses behind the client's back |
 | **Summary carrier on the wire** | **the `compress` tool call** | **an `acp_summary` user message** |
 | System messages on the wire | always exactly 1 (client + prompt) | always exactly 1 (client + prompt) — summaries ride on user messages |
 | SGLang "single system" 400 (#377) | cannot happen | cannot happen (summaries are user messages, not system) |
--- a/src/preflight.ts
+++ b/src/preflight.ts
@@ -44,6 +44,8 @@ export interface PreflightDeps {
     log: (level: string, msg: string) => void;
     /** Constant floor on the forwarded-payload size for this request (image bytes, #488). Folding only ever removes images, so adding this to every text estimate keeps the fit decision sound for multimodal payloads. */
     imageFloor?: number;
+    /** Constant wire overhead for this request (system prompt + tool definitions, #470). Folding never removes it, so every fit decision must add it — otherwise the loop stops with "text fits" while the billed input still overflows. */
+    wireOverhead?: number;
 }
 
 export type PreflightFailureKind = "upstream" | "exhausted" | "aborted";
@@ -247,7 +249,7 @@ function noEmergencyTruncate(config: Config): Config {
 
 export async function preflightCompress(deps: PreflightDeps, messages: CoreMessage[]): Promise<PreflightResult> {
     const limit = deps.config.modelContextLimit;
-    const result: PreflightResult = { compressedRanges: 0, savedTokens: 0, payloadEstimate: estimateCoreMessages(messages) + (deps.imageFloor ?? 0) };
+    const result: PreflightResult = { compressedRanges: 0, savedTokens: 0, payloadEstimate: estimateCoreMessages(messages) + (deps.imageFloor ?? 0) + (deps.wireOverhead ?? 0) };
     if (limit <= 0) return result;
     const budget = Math.max(MIN_CHUNK_TOKENS, Math.floor(limit * CHUNK_FRACTION));
     // applyCompression rejects ranges below config.compress.minCompressRange
@@ -284,10 +286,10 @@ export async function preflightCompress(deps: PreflightDeps, messages: CoreMessa
         // Floor on the session's measured input baseline: the upstream's
         // input_tokens also covers the system prompt + tool definitions, which
         // are not in turn.messages, so the direct estimate can undershoot.
-        currentTokens = Math.max(deps.session.stats.lastInputTokens, estimateCoreMessages(turn.messages) + (deps.imageFloor ?? 0));
+        currentTokens = Math.max(deps.session.stats.lastInputTokens, estimateCoreMessages(turn.messages) + (deps.imageFloor ?? 0) + (deps.wireOverhead ?? 0));
         // The caller's forward/fail-fast gate uses the payload's own estimate
         // (the floor can be stale — see PreflightResult.payloadEstimate).
-        result.payloadEstimate = estimateCoreMessages(turn.messages) + (deps.imageFloor ?? 0);
+        result.payloadEstimate = estimateCoreMessages(turn.messages) + (deps.imageFloor ?? 0) + (deps.wireOverhead ?? 0);
         if (startTokens < 0) startTokens = currentTokens;
         if (currentTokens < limit) break;
         const ranges = viableRanges(turn.nudge?.compressibleRanges ?? []);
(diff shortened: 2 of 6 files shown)


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 Done — opened **PR #577**: https://github.com/ranxianglei/billion-context/pull/577 (branch `wt-576-ework-daemon` → `master`).

**What changed**
- `src/server.ts`: new `estimateWireOverhead(protocol, body)` counts system/instructions text (anthropic string/array forms, openai hoisted system/developer messages, responses instructions) + tools JSON from the rebuilt wire body via `defaultCountTokens` (unparseable → 0; text-only so it never double-counts image tokens). Wired into the preflight trigger and both post-fold fit gates.
- `src/preflight.ts`: `PreflightDeps.wireOverhead` floors the fold loop's `currentTokens`/`payloadEstimate`, so it keeps folding until the true billed input fits (previously it stopped at "text fits" and then fail-fasted against the caller's now overhead-aware gate → 502 on foldable payloads).
- `README.md`: corrected the two-mode table's plugin-mode preflight cell ("No" → runs unconditionally in both modes as the last-resort backstop).

**Verification**
- New `tests/wire-overhead.test.ts` (6 unit tests) + a new e2e proving envelope-driven overflow folds & forwards — confirmed it **fails on master** (skips preflight entirely) and passes here.
- Retuned the affected fixtures to sit under honest accounting: codex-compact harness window 10k→12k (setup turn's billed ≈10.1k stays under while `bigConversation()` stays over), finality turn-1 filler trimmed, issue453 clamp re-sized to window 400k / max_tokens 150k so the clamp acts alone.
- `npm run typecheck` ✅ · `npm run build` ✅ · `npm test` **1070/1071** — the single failure is a pre-existing, environment-dependent CLI-launcher test (`resolveClientCommand: codex/claude resolve to themselves`), present on baseline master and unrelated to this change.
- Gated E2E regression (`tests/e2e/e2e-codex.test.ts`, `ACP_TEST_E2E=1`) couldn't run in this sandbox (no live upstream at the default local address; codex-cli present, dist built, zero-token preflight check passed). CI's manual-dispatch run covers it.

One clarification for your call: the issue's "#330 e2e re-tuned (24k chars instead of 32k)" has no matching test by that name in the tree (grep found none — only an unrelated `windowEnv(32_000)` in the gated e2e). That behavior is covered by the new `#470` e2e, so I didn't retune a separate test. If you had a specific existing test in mind, point me at it and I'll adjust.

Ready for review — not merging per repo rules.

ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 **Closed as duplicate of #577** (per @ranxianglei's cross-ref on #577: one of the two PRs implementing the #470 fix should be closed so the issue resolves through a single PR).

Why #577 was kept over this one:

1. **Clean base.** #577's head (`3b746a5`) is a single commit directly on current master (`a184ff9`) — merging it lands exactly the #470 fix and nothing else. This PR's head (`8e0a6ad`) carries the orphaned #337 merge commit (`6820e7d`) in its history; since current master does *not* contain #337's code, merging this PR would silently re-land the #330 soft-zone relax + #496 image-cost fix alongside the #470 fix, undisclosed (the body only says "Closes #470"). Master was deliberately rolled back after #337's merge today — if #337 should land, do it as its own disclosed PR (the code is intact at `6820e7d` / branch `2026-08-29_preflight-soft-zone-relax`).
2. **More complete iteration.** This PR's description claims retunes of `tests/codex-compact-e2e.test.ts` (window 10k→12k) and `tests/compaction-trigger-finality.test.ts` (trimmed turn-1 filler) — but neither file is in this diff. #577's diff *does* contain those retunes.
3. **Both heads verified green locally** anyway (typecheck clean; full suite fails only the known env-dependent `resolveClientCommand` launcher test, reproduced identically on master baseline), so no work is lost by closing this one.

Tagged `duplicate`. Reopen if you decide the #337 bundling was actually intended.

PR #577, opened 2026-09-06
fix: count system+tools wire overhead in the preflight trigger (#470)

### LINKED ISSUE #470: track: plugin-mode proxy-side emergency truncation (last-resort backstop, deferred from #467)

ranxianglei:
Closes #470 (residual gap from the analysis in the issue thread).

## Root cause

The preflight trigger and fit gates estimated the payload with `estimateCoreMessages` (messages only), but the billed input also carries the system prompt and tool definitions — the client's own (~10-20k on agent clients) plus the proxy-injected compress system prompt (~1.9k tokens) and ACP tools (~0.5k). Consequences:

- the trigger fired **late**: text alone "fit" the window while the real input already overflowed it (guaranteed upstream 400 on strict relays);
- the post-fold fit gate could pass on payloads whose **billed** input still exceeded the window.

#467 fixed this class of under-count only for the output clamp (`estimateInputTokens`); this PR applies the same term to preflight — the plugin-mode last resort.

## Change

- `src/server.ts`: new `estimateWireOverhead(protocol, body)` — counts system/instructions text (anthropic string/array forms, openai hoisted system/developer messages, responses instructions) + tools JSON from the rebuilt wire body, via `defaultCountTokens`. Unparseable → 0; text-only so it never double-counts image tokens.
- `preflightCompressIfNeeded`: trigger + both post-fold fit gates add the term.
- `src/preflight.ts`: `PreflightDeps.wireOverhead` — the fold loop's `currentTokens`/`payloadEstimate` floors include it, so the loop keeps folding until the true billed input fits. Without this the loop stopped at "text fits" and then fail-fasted against the caller's overhead-aware gate (502 on foldable payloads).
- `README.md`: the two-mode table's plugin-mode preflight cell said "No" — preflight runs unconditionally in both modes as the last-resort backstop; corrected.

## Test retuning (behavior change is intended)

- `tests/wire-overhead.test.ts` (new): 6 unit tests for the helper (protocol matrix, unparseable body → 0, Buffer parity).
- `tests/preflight-fail-fast.test.ts`: new e2e — `bigConversation()` is 13.1k tokens, under a 15k window on text alone (master skips preflight) but the injected compress prompt + ACP tools push the billed input to ~15.5k → preflight folds and forwards; the forwarded body no longer carries the early message and does carry the folded summary. Verified this e2e **fails on master** (skips preflight entirely) and passes here.
- `tests/codex-compact-e2e.test.ts`: widened the harness window 10k→12k so the setup turn's honest billed input (text + envelope ≈ 10.1k) stays under the window while `bigConversation()` remains over it — preserves master's exact message sizes and every forge/gate scenario.
- `tests/compaction-trigger-finality.test.ts`: trimmed turn-1 filler so its honest billed input sits just under the 10k window.
- `tests/issue453-backstop.test.ts`: the clamp e2e re-sized (window 400k / max_tokens 150k) to sit in the band where the clamp acts alone — the old shape now correctly triggers preflight (fold instead of output starvation), which is the better outcome.

Note on the issue's "#330 e2e re-tuned (24k chars instead of 32k)": there is no distinct existing test by that name in the tree (grep found none); the behavior it describes is covered by the new `#470` e2e above, so no separate retune was required.

## Pre-flight

- `npm run typecheck` ✅
- `npm test` 1070/1071 — the single failure is a pre-existing, environment-dependent CLI-launcher resolution test (`resolveClientCommand: codex/claude resolve to themselves`), unrelated to this change and present on baseline master too.
- `npm run build` ✅
- Gated E2E regression (`tests/e2e/e2e-codex.test.ts`, `ACP_TEST_E2E=1`) could not run in this sandbox (no live upstream at the default local address; codex-cli present, dist built, zero-token preflight check passed). CI's manual-dispatch run covers it.

## Out of scope (per the issue analysis)

- The optional stateless request-scoped drop (deep-tail plugin-mode backstop) — not needed while the trigger is honest; can be revisited if real-world reports show inputs that fit preflight's accounting but still 400.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ The proxy runs in one of two modes, and **the mode decides who executes
 | Client | ACP-native agent with the bili extension (pi/omp) | Any OpenAI/Anthropic client, no extension |
 | Who executes `compress` | **The agent** (pi runs it locally) | **The proxy** (server-side compress loop) |
 | `compress` tool call in the re-sent history? | Yes — part of the agent's own conversation | No — ephemeral proxy-loop traffic |
-| Preflight blocks (no tool call)? | No — overflow forces the model to call `compress` | Yes — `src/preflight.ts` compresses behind the client's back |
+| Preflight blocks (no tool call)? | Yes — active here too as a last-resort backstop (e.g. #453); normally the agent's own `compress` fires first via the nudge | Yes — `src/preflight.ts` compresses behind the client's back |
 | **Summary carrier on the wire** | **the `compress` tool call** | **an `acp_summary` user message** |
 | System messages on the wire | always exactly 1 (client + prompt) | always exactly 1 (client + prompt) — summaries ride on user messages |
 | SGLang "single system" 400 (#377) | cannot happen | cannot happen (summaries are user messages, not system) |
--- a/src/preflight.ts
+++ b/src/preflight.ts
@@ -44,6 +44,8 @@ export interface PreflightDeps {
     log: (level: string, msg: string) => void;
     /** Constant floor on the forwarded-payload size for this request (image bytes, #488). Folding only ever removes images, so adding this to every text estimate keeps the fit decision sound for multimodal payloads. */
     imageFloor?: number;
+    /** Constant billed-but-not-in-messages envelope cost (system/instructions + tools, #470). Folding never removes it, so it floors every size decision — without it the loop stops at "text fits" while the billed input overflows and the caller's overhead-aware gate fail-fasts a foldable payload. */
+    wireOverhead?: number;
 }
 
 export type PreflightFailureKind = "upstream" | "exhausted" | "aborted";
@@ -226,7 +228,7 @@ const ABORTED_FAILURE: PreflightFailure = { kind: "aborted", detail: "the client
 
 export async function preflightCompress(deps: PreflightDeps, messages: CoreMessage[]): Promise<PreflightResult> {
     const limit = deps.config.modelContextLimit;
-    const result: PreflightResult = { compressedRanges: 0, savedTokens: 0, payloadEstimate: estimateCoreMessages(messages) + (deps.imageFloor ?? 0) };
+    const result: PreflightResult = { compressedRanges: 0, savedTokens: 0, payloadEstimate: estimateCoreMessages(messages) + (deps.imageFloor ?? 0) + (deps.wireOverhead ?? 0) };
     if (limit <= 0) return result;
     const budget = Math.max(MIN_CHUNK_TOKENS, Math.floor(limit * CHUNK_FRACTION));
     // applyCompression rejects ranges below config.compress.minCompressRange
@@ -257,10 +259,10 @@ export async function preflightCompress(deps: PreflightDeps, messages: CoreMessa
         // Floor on the session's measured input baseline: the upstream's
         // input_tokens also covers the system prompt + tool definitions, which
         // are not in turn.messages, so the direct estimate can undershoot.
-        currentTokens = Math.max(deps.session.stats.lastInputTokens, estimateCoreMessages(turn.messages) + (deps.imageFloor ?? 0));
+        currentTokens = Math.max(deps.session.stats.lastInputTokens, estimateCoreMessages(turn.messages) + (deps.imageFloor ?? 0) + (deps.wireOverhead ?? 0));
         // The caller's forward/fail-fast gate uses the payload's own estimate
         // (the floor can be stale — see PreflightResult.payloadEstimate).
-        result.payloadEstimate = estimateCoreMessages(turn.messages) + (deps.imageFloor ?? 0);
+        result.payloadEstimate = estimateCoreMessages(turn.messages) + (deps.imageFloor ?? 0) + (deps.wireOverhead ?? 0);
         if (startTokens < 0) startTokens = currentTokens;
         if (currentTokens < limit) break;
         const ranges = viableRanges(turn.nudge?.compressibleRanges ?? []);
--- a/src/server.ts
+++ b/src/server.ts
@@ -1761,6 +1761,55 @@ export function clampOutputBudget(requested: number, inputEstimate: number, nati
     return cap;
 }
 
+function isObj(v: unknown): v is Record<string, unknown> {
+    return typeof v === "object" && v !== null;
+}
+
+/** #470: system/instructions + tools billed on top of the messages (which
+ *  estimateCoreMessages omits). Read from the rebuilt body so it's mode-correct;
+ *  text-only, so it never double-counts imageTokensInRawBody. Unparseable -> 0. */
+export function estimateWireOverhead(protocol: "anthropic" | "openai" | "responses", body: string | Buffer): number {
+    const s = typeof body === "string" ? body : body.toString("utf8");
+    let parsed: unknown;
+    try {
+        parsed = JSON.parse(s);
+    } catch {
+        return 0;
+    }
+    if (!isObj(parsed)) return 0;
+    const textOf = (content: unknown): string => {
+        if (typeof content === "string") return content;
+        if (!Array.isArray(content)) return "";
+        let out = "";
+        for (const part of content) if (isObj(part) && typeof part.text === "string") out += part.text;
+        return out;
+    };
+    let systemText = "";
+    if (protocol === "anthropic") {
+        const sys = parsed.system;
+        if (typeof sys === "string") systemText = sys;
+        else if (Array.isArray(sys)) for (const blk of sys) if (isObj(blk) && typeof blk.text === "string") systemText += blk.text;
+    } else if (protocol === "openai") {
+        const msgs = parsed.messages;
+        if (Array.isArray(msgs)) {
+            for (const m of msgs) {
+                if (!isObj(m) || (m.role !== "system" && m.role !== "developer")) continue;
+                systemText += textOf(m.content);
+            }
+        }
+    } else {
+        if (typeof parsed.instructions === "string") systemText += parsed.instructions;
+        const input = parsed.input;
+        if (Array.isArray(input)) {
+            for (const item of input) {
+                if (!isObj(item) || item.role !== "developer") continue;
+                systemText += textOf(item.content);
+            }
+        }
+    }
+    return defaultCountTokens(systemText) + defaultCountTokens(JSON.stringify(parsed.tools ?? []));
+}
+
 // Only override genuine cadence silences: skip the kernel's deliberate
 // "nothing compressible to offer" suppression (empty ranges).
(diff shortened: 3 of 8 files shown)


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 **Review complete — code verified, tests reproduced. One item needs your confirmation before merge; one description nit.**

**Verdict: the fix is correct and complete.** I checked out head `8e0a6ad` (base `a184ff9` = current master tip, so no rebase needed) and verified everything below locally.

## What I validated

- **Bug is real, correctly layered.** Master's `preflightCompressIfNeeded` sizes on `estimateCoreMessages` + images only; the billed input also carries system + tools, so in the band `text < limit < text+overhead` the trigger never fires → guaranteed upstream 400 on strict relays, and the post-fold gate could pass over-window billed input. Same class #467 fixed for the output clamp — this applies the same term to preflight. The layering is right (accounting fix, not a symptom patch).
- **All four consumption points covered:** trigger, both post-fold gates, and the fold-loop floors via `PreflightDeps.wireOverhead`. That last one matters — without it the loop stops at "text fits" and then fail-fasts against the caller's overhead-aware gate (502 on foldable payloads). Good catch.
- **No double-counting.** OpenAI: the kernel hoists the leading system/dev prefix out of `CoreMessage[]` (`openaiToCore`) and `injectOpenaiSystem` puts it back into the rebuilt body's messages → counted exactly once from the body scan. Anthropic/Responses: top-level `system`/`instructions`, not in messages. And `max(lastInputTokens, payloadEstimate)` is an alternative, not a sum — the measured baseline already includes overhead. One benign edge: a *mid-conversation* system/dev message on OpenAI would be counted twice (it stays in `processedMessages` AND matches the body scan) — conservative direction only (slightly earlier trigger), acceptable.
- **Both compression modes reasoned** (per AGENTS.md §6): preflight runs unconditionally (no `pluginMode` gate), and `estimateWireOverhead` reads `prepared.body` *post*-injection, so proxy mode counts the ACP tools + compress prompt while plugin mode counts the client's own manifest (bili tools suppressed by design). README table correction is right.
- **Post-fold gate recompute is sound.** Rebuilding via `runPrepare()` and gating on the actual forwarded payload eliminates the false-fit from `noEmergencyTruncate` trimming `result.payloadEstimate` more than normal prepare does. Reusing `overheadEstimate`/`imageTokens` across the rebuild is valid — system/tools/images are fold-invariant.
- **Pre-flight reproduced:** `npm run typecheck` ✅ · `npm run build` ✅ (self-contained dist) · `npm test` 1071/1072 — the single failure is the env-dependent `resolveClientCommand` launcher test; I reproduced the **identical failure on a master baseline worktree**, confirming it's pre-existing as claimed (your count was 1070/1071, mine 1071/1072 — count drift, same single failure). Targeted run of `wire-overhead`, `preflight-fail-fast`, `issue453-backstop`, `codex-compact-e2e`, `compaction-trigger-finality`: **all 32 pass**, including the new #470 e2e and the #330 relax e2e.
- **The bundled #330 relax code is coherent:** `noEmergencyTruncate` (×100 window) keeps usage under `truncate.threshold` so the loop reaches the relax step; relax-retry is gated on the payload's *own* estimate (not the stale-floor `currentTokens`); exhausted detail is actionable ("Raise the model context window or restart the session").

## ⚠️ Needs your confirmation: this PR silently re-lands orphaned PR #337

GitHub shows **#337 as merged** (`merged_at` 2026-09-06T06:31:59Z, merge commit `6820e7d`) — but that merge commit is **unreachable from current master** (`a184ff9`). Master contains none of #337's code (verified: no `relaxedConfig` / `noEmergencyTruncate` markers anywhere). This branch's history contains exactly that merge commit plus your fix, so **merging #577 lands #330 soft-zone relax + #496 image-cost fix atomically with the #470 fix**, even though GitHub displays #337 as already merged and the body only says "Closes #470".

The coupling is real (#470's e2e depends on the relax code; the early-exit restructuring interlocks with it), so bundling is probably intentional — but please confirm, and add a line to the PR body stating it re-lands #337 whose merge didn't stick to master, so future readers of master's history aren't confused.

## Nit: stale "Test retuning" section

The description claims retunes of `tests/codex-compact-e2e.test.ts` (window 10k→12k) and `tests/compaction-trigger-finality.test.ts` (trimmed turn-1 filler) — **neither file is in the diff** (both unchanged from master), and both pass as-is. Looks like a leftover from an intermediate iteration (the PR head is `wt-576-ework-daemon`, not the branch name cited in the body). No action needed beyond correcting the description.

## Pre-merge checklist item

`server.ts`'s request pipeline changed, so per AGENTS.md §3 the gated codex e2e should run at least once against a live upstream before merge (CI manual dispatch with `E2E_UPSTREAM_URL`/`E2E_UPSTREAM_KEY`, or local sglang) — you noted the sandbox can't reach one. Everything else is green on my side; once you confirm the #337 bundling, this is ready for a human merge.

ranxianglei:
Cross-ref: #576 implements the same #470 fix (same preflight wire-overhead root cause, opened ~1h earlier from the issue-thread analysis). Both are currently open against master — one of the two should be closed so #470 resolves through a single PR. #576 is mergeable against current master (53ce3ba) with pre-flight typecheck/test/build green (1067/1069 pass, 2 known plugin-agent env fails).

ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 **Dedup done: kept #577, closed #576 as duplicate** (tagged `duplicate`, explained in its thread). Details below.

## What I found

The two PRs are **two independent implementations of the same fix**, not one branch pushed twice:

| | #576 (closed) | #577 (kept, this PR) |
|---|---|---|
| Head | `8e0a6ad` @ `2026-09-06_preflight-wire-overhead` | `3b746a5` @ `wt-576-ework-daemon` |
| Base | orphaned #337 merge `6820e7d` + fix | **single commit directly on current master `a184ff9`** |
| Test retunes (`codex-compact-e2e` 10k→12k window, `compaction-trigger-finality` filler trim) | claimed in description, **absent from diff** | **present in diff** |
| Responses overhead accounting | `instructions` only | `instructions` **plus** developer items inside `input` |

Note: my earlier review in this thread was of `8e0a6ad` — which is actually **#576's** head (the issue body cited that branch name). The verification results carry over to this PR's head too, since I re-ran everything on `3b746a5`:

- `npm run typecheck` ✅ clean
- `npm test`: 1071 tests, **1070 pass, 1 fail** — the sole failure is the env-dependent `resolveClientCommand` launcher test, reproduced identically on a master baseline worktree (pre-existing, unrelated)
- New e2e `#470: text fits the window alone but system+tools push the billed input over → preflight folds and forwards` passes; post-fold rebuild gate and openai hoisted-system handling match what I reviewed and endorsed

## Why #577 wins

1. **Clean base = clean landing.** Merging #577 lands exactly the #470 fix against real current master. Merging #576 would have silently re-landed the rolled-back #337 work (#330 soft-zone relax + #496 image-cost fix), which its body never discloses. Master was deliberately reset after #337's merge today; if #337 should land, it deserves its own disclosed PR (code intact at `6820e7d` / branch `2026-08-29_preflight-soft-zone-relax`).
2. **#577 is the more complete iteration** — it contains the test retunes #576's description promised but didn't ship, plus slightly more thorough responses-protocol accounting.

## Two notes

- Your comment cites "current master (53ce3ba)" — remote master is still `a184ff9` (fresh fetch); `53ce3ba` doesn't exist as any remote ref here. If that's your local master with #337 merged in, then #576's green run was against a base that isn't remote master — moot now, since #577 is verified directly against `a184ff9`.
- Pre-merge checklist item still stands from my first review: `server.ts`'s request pipeline changed, so run the gated codex e2e once against a live upstream (CI manual dispatch or local sglang) before merging.

Everything else green on my side. Merge stays human-only per repo rules — ready when you are.

ranxianglei:
Superseded by #576 (same #470 fix, merged). This branch also conflicts with current master. Closing as duplicate — thanks for the parallel implementation.

Issue #387, opened 2026-08-31
acp_status 缝合报告:Compressible ranges/Nudge 读冻结的 ctx.nudge,压缩成功后旧区间与新块同框并存


wosnxn123:
## 症状(2026-08-31 取证,HEAD 仍在)

`compress m00001-m00025` **成功**返回(54125 tokens saved、块 b1 激活)后,紧接着的 `acp_status` 在**同一份输出**里:COMPRESSED BLOCKS 区正确显示 b1 活跃,而 "Compressible ranges" 区**仍列出刚被压掉的 m00001–m00025(25 msgs 54.1K)为可压缩**,Nudge 行仍印旧值 `ready: T1 54073`——而 bili.log(11487 行)显示 nudge 缓存早在 16:37:56 已翻成 `max compressible 4793`。模型极易据旧 ranges 发起重复/错位压缩(本仓库取证里就有一次)。

## 根因

`handleAcpStatus`(bili `src/loop/core.ts:136-156`,dist 51221-51236):
- base 报告读 `ctx.session.state`(压缩时被 `applyRanges` 原地更新,dist 50421)→ **live**;
- ranges/protected 读 `ctx.nudge` —— **frozen**:`ctx.nudge` 仅在 processTurn 的 recommend/nudge 节点于请求 prepare 时生成(buildNodes,dist 45578-45592),循环内**没有任何代码重新赋值 ctx.nudge**(loop/core.ts 里 `nudge` 只有类型字段与读取,loop 内 re-request 走 `adapter.buildRequest` 不重过 processTurn);plugin 工具路径同样读 `mem?.nudge`(plugin.ts:417)。
- 另:`formatRanges`(dist 43650-43725)把 compressible+protected 合成一列,陈旧的 protected 行还会伪装成可压缩项。

即一份报告内缝了两个新鲜度层级,状态互斥。

## 建议修复(择一,倾向前者)
1. `handleAcpStatus` 内以 `ctx.session.state + ctx.messages` 现算 ranges/protected(复用 `viableRanges` 过滤,参照 dist 56272 的一次性过滤语义);
2. (不推荐)在 Nudge/ranges 区块头标注 `as of last request` ——数据仍是错的,只是不撒谎。

## 验收标准
- [ ] compress 成功后的同轮 `acp_status`:被压区间不再出现在 Compressible ranges;ready/nudge 数值与 state 派生一致
[...]

PR #397, opened 2026-08-31
fix: acp_status derives ranges/nudge from live state, not the prepare-time snapshot (#389)

ranxianglei:
## Problem (#389)

`acp_status` stitched two freshness levels into one report: the base report read live `ctx.session.state`, while the appended "Compressible ranges" + "Nudge:" lines read `ctx.nudge` — a snapshot frozen at request-prepare time and never re-assigned inside the loop. After a successful `compress`, the same-turn `acp_status` still listed the just-compressed range as compressible with the stale nudge value, inviting duplicate/mis-targeted compress calls.

## Fix (issue option 1)

- **New `src/acp-status.ts`** — shared `handleAcpStatus(args, ctx)`. After the base report (and the existing `scope=` short-circuit, which already carried no stale data), it re-runs the kernel's **pure** `processTurn` over `ctx.session.state + ctx.messages` (verified: every pipeline node returns new objects; `syncBlocks` deep-copies state) with `tokenCount = session.stats.lastInputTokens` — the same estimate the next prepare uses — and derives `compressibleRanges` (through the same one-shot `viableRanges` filter the prepare path applies) + `protectedRanges` + the Nudge line from the fresh decision. `turn.state` is intentionally NOT adopted, so nudge cadence bookkeeping is untouched; a recompute failure degrades to the base-only report rather than serving stale data.
- **Converged all three dispatcher copies** (issue acceptance #3): `src/stream.ts` (`executeAnthropicProxyTool`) and `src/compress-loop-responses.ts` (`executeProxyTool`) previously called bare `buildStatusReport` and ignored every option — they now delegate to the shared handler, so all protocols emit the same output shape and honor `scope`/`view`/`tool`/`sort`/`limit`.
- **Removed `LoopCtx.nudge`** and its setters (`src/server.ts`, `src/plugin.ts`) — the frozen snapshot can no longer leak into the report, and the plugin tool path gets the live behavior for free.

## Tests (`tests/acp-status.test.ts`, 3 new)

1. **Acceptance #2**: state change (real `applyRanges` compress) → immediate `handleAcpStatus` with NO new prepare → compressed refs absent from "Compressible ranges", b1 active, Nudge line state-derived.
2. `scope=` short-circuit: base-only, no ranges/Nudge section.
3. **Acceptance #1** end-to-end through `runCompressLoop` + responses adapter: `compress` + `acp_status` in the same round — the acp_status marker no longer lists the just-compressed range.

## Pre-flight

- `npm run typecheck` — clean
[...]

PR #578, opened 2026-09-06
docs: correct false renumber claim in preflight comment (#387)

### LINKED ISSUE #595: docs: preflight.ts comment falsely claims compress renumbers surviving refs
### Problem
The round-loop comment in `src/preflight.ts` (~:365) still says:

> Re-run the pipeline each round: every successful compress renumbers the surviving refs, so the previous round's range refs are stale.

That was true before the #387 work, but refs are **stable per-session snapshots** now — what changes across rounds is which ranges remain compressible under them, not the numbering.

### Impact
The comment invites wrong "fixes" (re-adding renumbering logic, caching by ref) and misdocuments the core invariant. Docs-only.

### Suggested fix
Reword the comment to the #387-era semantics. PR #578 carries the exact 4-line reword (merges clean against master).

ranxianglei:
Closes #595 — tracking issue for this PR.

## What

Follow-up to ranxianglei/billion-context#387. The kernel fix (acp-kernel PR #175, merged 2026-08-31, shipped in acp-kernel — now pinned at 0.0.54) removed the false `a prior compress renumbered the remaining refs` claim from all user-visible text. One remnant of the same false claim survived in this repo as a code comment:

`src/preflight.ts` (preflight loop):
> ~~Re-run the pipeline each round: every successful compress **renumbers** the surviving refs, so the previous round's range refs are stale.~~

## Why it matters

Not user-visible, but it propagates the exact misconception that caused #387 to any future reader of the preflight loop. The corrected rationale: a successful compress **hides its range behind a new block**, changing which ranges remain compressible under the same (stable, per-session-snapshot) refs — that is why the pipeline must re-run each round.

## Scope

- Comment-only change (verified by diff): 1 file, 4 insertions, 2 deletions, zero code changes.
- After this lands, `grep -r renumber src/ tests/` returns nothing in this repo.

## Pre-flight

- No code changed; no test/typecheck delta possible (diff is entirely within one comment block).

### DIFF
--- a/src/preflight.ts
+++ b/src/preflight.ts
@@ -271,8 +271,10 @@ export async function preflightCompress(deps: PreflightDeps, messages: CoreMessa
             failure = ABORTED_FAILURE;
             break;
         }
-        // Re-run the pipeline each round: every successful compress renumbers
-        // the surviving refs, so the previous round's range refs are stale.
+        // Re-run the pipeline each round: a successful compress hides its
+        // range behind a new block, changing the visible view; refs stay
+        // stable per-session snapshots (#387), but which ranges are
+        // compressible under them does not.
         const turn = deps.core.processTurn({
             messages,
             state: deps.session.state,


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 **Already implemented & verified — PR #578 correctly does this; I did not push a duplicate.**

A prior daemon run had already opened PR #578 for exactly this fix. Since its branch name is taken and a second PR would be a pure duplicate, I verified the existing PR against the stated scope instead of re-doing it:

**Scope (authoritative, via the GitHub API):**
- ✅ Comment-only: `1 file changed, 4 insertions(+), 2 deletions(-)` — only `src/preflight.ts`, zero code changes. Matches the spec precisely.
- ✅ The removed line was the repo's only `renumber` occurrence, so post-merge `grep -rn renumber src/ tests/` returns nothing.

**New comment text:**
> Re-run the pipeline each round: a successful compress hides its range behind a new block, changing the visible view; refs stay stable per-session snapshots (#387), but which ranges are compressible under them does not.

This is faithful to the corrected rationale in the issue: the range is hidden behind a new block; refs stay stable (per-session snapshots); what changes is *which ranges are compressible* — hence the pipeline re-runs each round.

**Pre-flight:** comment-only ⇒ no typecheck/test delta is possible (as the issue notes). For completeness I ran `typecheck` + `build` + the full `test` suite on an equivalent commit: green, except one unrelated launcher test (`resolveClientCommand: codex/claude resolve to themselves`) that fails locally only where a `codex` CLI happens to be installed — it fails identically *without* this change and passes on CI runners.

**Status:** mergeable (base is simply ahead of head — normal). Ready for review/merge. Per project rules I don't merge PRs myself — please merge when you're happy: https://github.com/ranxianglei/billion-context/pull/578

ranxianglei:
**Review vs master 3cd85b6 (kernel 0.0.56):** ✅ merge-ready.

- The stale comment still sits at `src/preflight.ts:365` on master — unchanged target.
- New wording is semantically correct post-#397: ids are stable per-session (`acp-status.ts` recomputes ranges via a fresh `processTurn`); what changes after a fold is which ranges remain compressible under those refs.
- No code change, no test impact, applies clean.

No follow-ups needed. (tracking: #595)

ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 Re-verified against the live tree — agree, merge-ready.

Checked just now (master has since ticked forward to `1330fc5`, just past your `3cd85b6`; acp-kernel still pinned at 0.0.56):
- PR #578 reports `mergeable: true` — applies clean against the current tip.
- The stale comment is still the *only* `renumber` in the repo, unchanged, at `src/preflight.ts:365`.

Nothing changes your read, and the extra commits on master didn't disturb the target. Nothing further from my side — it's yours to merge whenever. (tracking #595 noted.)

Issue #552, opened 2026-09-05
billion-context: 400 Bad Request when codex sends `role: "developer"` messages — request a request-body compat transform hook


Juna9969:
## Environment

- **billion-context (bili)**: v0.1.85
- **Client**: codex CLI 0.153.2 (Codex TUI), using the `codex` / `/v1/responses` protocol
- **Upstream**: an OpenAI-compatible relay (`api.b.ai`, whose backend is Tencent Cloud Hunyuan) via `bili`'s `/bili/<upstream>` wrapping
- **Observer**: bili forwards the request body verbatim (only base_url is rewritten via `httpRewrites`); a local OpenAI-compatible gateway converts the Responses API body to Chat Completions before reaching the relay

## Problem

Since codex 0.153.x, the Codex TUI sends its built-in agent instructions as a **normal message item with `role: "developer"`** inside the Responses API `input` array, e.g.:

```json
{
  "model": "gpt-5.6-sol",
  "input": [
    {
      "type": "message",
      "role": "developer",
      "content": [
        {
          "type": "input_text",
          "text": "# Portable Codex agent instructions\n\nThese instructions..."
        }
      ]
    },
    { "type": "message", "role": "user", "content": [{ "type": "input_text", "text": "Hello" }] }
  ],
  "instructions": null
}
```

Every downstream OpenAI-compatible server that converts to the Chat Completions shape (or natively accepts `messages`) has **no `developer` role** — the standard mapping is `developer` → `system` (OpenAI's own chat protocol exposes only `system`, and OpenAI's docs state codex instructions map to the system role). When the message with `role: "developer"` is forwarded unchanged, upstream validation fails:

```
400 {"error":{"message":"The request is invalid: invoke model error: ... forward bad request, [HTTP 400] {\"error\":{\"message\":\"Invalid role: developer\", ...}}}}
```

(observed variant through a multi-hop backend: `http client codec StatusCode: Bad Request`, body rejected during JSON re-serialization at the second hop — Tencent Cloud Hunyuan)

This is not a hypothetical: it breaks **every reque it with codex ≥ 0.153 as client + a chat-completions/converting upstream**, which is the most common self-hosted/relay setup this tool targets.

## Root cause

[...]

PR #579, opened 2026-09-06
fix: compat.roles wire role rewrite for role-rejecting upstreams (#552)

ranxianglei:
Follow-up per the maintainer's UX feedback on this PR (thank you!) — the amendment is now implemented on top of `9603e42` as `d2b15b3` + `9b6efc2`.

## Auto-fix: no config needed for the common case

- **Hint direction corrected**: the log hint is no longer a global `compat` block — it prints a **per-provider snippet with the real upstream origin** (bili knows exactly which provider rejected), ready to copy-paste into the config file only if the user wants the fix to be permanent.
- **Learn is session-scoped, config stays user-owned**: on a role-rejection 400 bili rewrites `{role: "system"}`, retries once, and on success records the mapping in `session.metadata.learnedCompatRoles` (same precedent as `learnedContextLimits`). Later requests in the same session apply the learned map up-front and skip the 400 round-trip entirely. **The config file is never touched.**
- If the retry also fails, the original 400 passes through verbatim (exactly one retry, no loop).

New e2e coverage: E proves the learn (3 upstream hits: rejected → retried → pre-rewritten), F proves the verbatim 400 + single retry.

## Bonus bug found & fixed while testing (`9b6efc2`)

The new F test exposed a **fetch-util timer leak** that pins the event loop for 10 minutes: a response body's pending `pull` can resolve *after* the consumer already called `clearTimer` (abandoned body — exactly F's failed retry, but also any client abort racing the last chunk). The late chunk re-armed a fresh `UPSTREAM_TIMEOUT_MS` timer that nothing would ever clear. `rearm()` is now a no-op once `cleanup()` has run, and F asserts `_liveUpstreamTimersForTest() === 0` as a regression guard. This class was previously documented at the `#411` call site — it could plausibly explain stray long-lived idle procs in the wild.

Pre-flight: typecheck ✓ · targeted 13/13 ✓ · full suite 1080 pass / 2 known plugin-agent env fails (baseline) ✓ · build ✓.

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -107,6 +107,13 @@ Top-level keys that control how the proxy listens and behaves globally.
 - **Status:** ACTIVE
 - **Description:** Forward every request to the upstream **without** compression, tool injection, or nudging. Equivalent to `ACP_PASSTHROUGH=1`. Handy for A/B comparison against the uncompressed baseline.
 
+### `compat`
+
+- **Type:** `{ roles?: Record<string, string> }`
+- **Default:** `{}` (disabled)
+- **Status:** ACTIVE
+- **Description:** Global wire-compat role map. `roles` maps message roles to the role name your upstream accepts, e.g. `{"compat":{"roles":{"developer":"system"}}}` rewrites `developer` → `system` on the final forwarded body for upstreams that reject the `developer` role (#552, newer codex clients). Applies to `openai` chat-completions and `responses` requests; exact-match roles only, everything else in the body is untouched; re-sent compress-retry bodies carry the same rewrite. Per-provider `compat.roles` entries (see [Providers](#providers)) win per key. Default `{}` forwards bodies byte-for-byte unchanged.
+
 ### `proxy`
 
 - **Type:** `string`
@@ -170,6 +177,13 @@ A shallow key (`https://open.bigmodel.cn`) matches every path on that host. A de
 - **Default:** *(inherits global `compress`)*
 - **Status:** ACTIVE
 - **Description:** Per-provider compression overrides. This is **level 2 of 3** in the merge hierarchy — see [Compression Tuning](#compression-tuning).
+
+### `compat`
+
+- **Type:** `{ roles?: Record<string, string> }`
+- **Default:** `{}` (disabled)
+- **Status:** ACTIVE
+- **Description:** Per-provider wire-compat overrides. `roles` maps message roles to the role name this upstream accepts, e.g. `{"developer": "system"}` for upstreams that reject the `developer` role newer codex clients send (#552). Applied to the final forwarded `openai`/`responses` body — client-sent roles and bili's own injected prompt alike — and to every body the compress-retry loops re-send. Wins per key over the global `compat` block (see [Server Settings](#server-settings)). Default `{}` forwards byte-for-byte unchanged.
 
 ---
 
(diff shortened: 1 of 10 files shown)


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 **Review of PR #579 (`fix: compat.roles wire role rewrite`, head `9603e42`, 1 commit ahead of origin/master `53ce3ba`)**

**Verdict: implementation verified end-to-end, no functional blockers.** Three non-blocking findings below — two are description/comment accuracy, one is scope disclosure.

## Independently verified

Ran on the checked-out head (clean tree): `npm run typecheck` ✓, `npm run build` ✓ (self-contained dist), targeted `tests/compat-roles.test.ts` **10/10** ✓, full suite **1078/1079** — the single failure is `tests/launcher.test.ts` "resolveClientCommand: codex/claude resolve to themselves", which fails **identically on origin/master** (verified via worktree; environmental — no codex/claude binaries in this sandbox). `plugin-agent.test.ts` passes 35/35 here. Zero PR-related failures.

Design conformance, point by point:
- **Single choke point**: all 5 `forward()` call sites route through the rewrite, including MITM side-prepared (`src/server.ts:1360`) and passthrough (`:1556`). Protocol chain `prepared?.protocol ?? route?.explicitProtocol ?? inferWireProtocol(req.url)` is sound (query stripped, `/responses/compact` covered, `anthropic` excluded by the openai/responses check).
- **Content-length consistency claim holds**: incoming `content-length` is *not* copied into forwarded headers (`src/server.ts:241` skips it), so undici computes it from `wireBody` — re-stringified bytes stay consistent.
- **Default transparency**: empty map short-circuits before parsing; `applyCompatRoles` returns the original string when `rewritten === 0` (no re-stringify) — unit-tested and confirmed by e2e C (live proxy, no compat config → `developer` reaches the wire untouched, i.e. responses path confirmed as #552's live surface).
- **Retry loops**: `wireTransform` threaded through both `runCompressLoop` (`src/loop/core.ts`) and `compressLoopResponsesJson` (`src/compress-loop-responses.ts`) fetch sites — necessary, since the responses-path compress prompt itself carries `developer`.
- **Hot reload**: `opts.compat` refreshed at both the web Apply path (~`:795`) and `__bili/config/reload` (~`:3420`).
- **Zero kernel changes**: no acp-kernel bump; rewrite happens post-prepare on the serialized body only, so session state / block IDs / ACP tracking are untouched. Per-provider per-key precedence via `findRoute` longest-prefix matches the compress-settings convention.
- **Docs**: README section + CONFIGURATION.md / .zh-CN.md (global + per-provider) present and accurate.

## Findings (non-blocking)

**1. PR body misstates the chat-path mechanism.** It says the chat path "already normalizes developer→system during the rebuild (kernel coreToOpenai)". That is not what happens: acp-kernel 0.0.54's `coreToOpenai` **restores** developer — `node_modules/acp-kernel/dist/wire/index.js:357`: `m.originalRole === "developer" ? "developer" : "system"` (the same code quoted in issue floor 4). What actually keeps `developer` off the chat wire is bili's own `systemToUser` (`src/util.ts:201-212`), which re-voices **both** `system` and `developer` → `user` on the rebuilt body. The conclusion stands (chat carries no `developer`; the boundary still covers whatever does reach the wire), but please correct the mechanism in the PR description.

**2. Test B's comment overclaims what it proves** (`tests/compat-roles.test.ts:190-193`). By the time compat runs, the client's `developer` item has already become `user` (via `systemToUser`); what the `system→user` mapping actually rewrites is bili's **injected head system prompt**, re-added by `injectOpenaiSystem` *after* `systemToUser` (`src/server.ts:~1872`). The assertion is still valid — without compat, `seen[0][0]` would be `"system"` — so the openai boundary genuinely fires; the comment should just attribute the rewrite to the injected system message rather than "the rebuild normalizing developer".

**3. Unmentioned bundled change: acp_status unification in both compress loops.** This commit also deletes the stale local handlers — `src/loop/core.ts`'s `handleAcpStatus` (which read `ctx.nudge`, a field **no caller ever set**: the single call site `src/server.ts:3137` passes no `nudge`, so unified-loop acp_status was base+archive-only with a dead field) and `src/compress-loop-responses.ts`'s bare `buildStatusReport` (base-only, no archive) — replacing both with the shared live-state `handleAcpStatus` from `src/acp-status.js` (#389 semantics). Direction is right (loops now match the canonical handler; dead code removed; typecheck/tests green), but it's an agent-visible tool-output change unrelated to `compat.roles` and absent from the PR body. Please add a sentence describing it (or split it out).

**4. Nit (optional):** `wireTransform` mutates the round body in place; with a pathological *chained* map (e.g. `{developer:"system", system:"user"}`), repeated application across retries cascades roles beyond one hop (pass 1: developer→system, pass 2 on the same object: system→user). Realistic single-key maps are idempotent, so this is fine given opt-in — snapshotting the original or applying once per round would make it strictly safe if you ever want that guarantee.

LGTM pending the description fixes (#1 required, #3 required, #2/#4 optional). Merging stays with you — I don't merge PRs.

PR #581, opened 2026-09-06
fix: skip auto-update when running from a git source checkout (#580)

### LINKED ISSUE #580: auto-update overwrites tracked files when run from a git source checkout (findInstallDir resolves to repo root)
来源: 分析 [ranxianglei/billion-context#552](https://github.com/ranxianglei/billion-context/issues/552)(developer role 400)期间,为复现而运行本地构建的 `dist/index.js` 时发现。

## Symptom

When bili is started from a **git source checkout** (`node dist/index.js start`), auto-update treats the checkout as its install directory and silently merge-copies the published npm tarball over it. Tracked files are rewritten without any user action:

- `package.json` — `version` field and pinned `devDependencies` (e.g. the exact `acp-kernel` pin) replaced with the published values
- `README.md`, `README.zh-CN.md` and any other file present in the tarball — overwritten with the published version's content

Observed 3× deterministically in one session (local 0.1.84 vs registry 0.1.85, then 0.1.87): every startup of the local dist while a newer registry version existed rewrote the working tree. `git status` after each run showed `M package.json` (+ READMEs), requiring manual `git checkout --` to restore.

## Root cause

- `findInstallDir()` (`src/update.ts:101-114`) walks up from the running `dist/index.js` until it finds a `package.json` whose `name` matches — for a source checkout that resolves to the **repo root**.
- `installViaTarball()` (`src/update.ts:438`) then backs up that directory and does `cp(stagingDir → installDir, { recursive: true, force: true })` (`src/update.ts:571`) — a merge-copy that overwrites every file present in the tarball (`.git/`, `src/`, `tests/` survive because they aren't in the tarball).

The pre-flight only checks writability (`src/update.ts:451`); nothing distinguishes a real install from a developer checkout.

## Impact

1. Silently corrupts a developer's working tree — including the `version` field, which this repo's release governance reserves for `*_release-v*` branches. An unattended auto-update creates exactly the diff that rule forbids, at risk of being committed accidentally.
2. A developer testing an unreleased local build has their local code silently replaced on disk by the published tarball (recoverable from git, but surprising mid-work).
3. Any automation that runs bili from a clone (agent sandboxes, CI dry-runs) inherits the same contamination.

## Suggested fix

Skip in-place self-update when the resolved install dir is a git working tree — e.g. check for a `.git` entry in `installDir` (npm installs never contain one; `npm pack` strips VCS metadata) and log `running from source checkout — skipping auto-update (use npm install -g)` instead of proceeding. Alternatively or additionally, refuse when `installDir` is not under the global npm prefix. Keep current behavior for genuine non-git install locations (manual extraction).

## Repro

1. Clone the repo, `npm ci && npm run build`.
2. `node dist/index.js start --port <free-port>`.
3. Wait ~10s while a newer version exists on the npm registry (startup check ignores throttle).
4. `git status` → `M package.json` (+ READMEs if they changed between versions).

ranxianglei:
Fixes #580.

## Problem

When bili runs from a **git source checkout** (`node dist/index.js start`), `findInstallDir()` walks up from the running `dist/index.js` until it finds a `package.json` whose name matches — which resolves to the **repo root**. Auto-update then backs that directory up and merge-copies the published tarball over it (`cp(stagingDir → installDir, { recursive: true, force: true })`), silently rewriting tracked files: `package.json` (including the `version` field and pinned devDeps such as the exact `acp-kernel` pin), READMEs, etc. The only pre-flight was a writability check.

The incident already happened once during development (`devlog/2026-08-25_remote-host-bind/WORKLOG.md`: repo-dist server 0.1.52 auto-updated itself to 0.1.53 and rewrote the repo's package.json), and #580 reports it deterministically 3 more times (local 0.1.84 vs registry 0.1.85 / 0.1.87).

## Fix (as suggested in the issue)

Skip in-place self-update when the resolved install dir is a git working tree:

- New `isGitWorkingTree(dir)` in `src/update.ts`: checks for a `.git` entry — directory (normal clone) or file (linked worktree / submodule pointer). npm installs never contain `.git` (`npm pack` strips VCS metadata), so its presence reliably marks a source checkout.
- `checkForUpdate()`: resolves the install dir early and returns before touching the registry, logging `[update] running from a source checkout (… ) — skipping auto-update (use npm install -g billion-context)`. Covers both the periodic auto-update loop and the manual `bili update` command (both go through `checkForUpdate`).
- `installViaTarball()`: structural pre-flight refusal for direct callers of the exported function.

Behavior for genuine non-git install locations (npm global installs, manual extraction) is unchanged. I deliberately did not also adopt the issue's alternative "refuse if not under the global npm prefix": prefix detection is fragile across homebrew/pnpm/yarn/Windows/manual layouts, while the `.git` marker is precise for the reported failure mode.

Known limitation (out of scope): a monorepo sub-package (e.g. `packages/billion-context` inside a larger git repo) has no `.git` at its own root, so it is not covered by this guard.

## Tests

- `isGitWorkingTree`: true for `.git` dir, true for `.git` file (worktree pointer), false when absent.
- `installViaTarball`: refuses when the install dir contains `.git`; asserts tracked files remain byte-for-byte intact (package.json version, README, dist entry).

End-to-end repro verified locally: built `dist` from this checkout, ran `node dist/index.js start`, the startup check logged the skip message, and `git status` stayed clean.

## Pre-flight

- `npm run typecheck` ✅
- `npm test` — 1065/1066 pass; the single failure (`launcher.test.ts: resolveClientCommand: codex/claude resolve to themselves`) is environment-dependent and pre-existing on the pristine tree (this sandbox has `/usr/bin/codex` on PATH; fails identically without this change)
- `npm run build` ✅
- E2E suite not run: change is confined to `src/update.ts` (not part of the request pipeline per AGENTS.md §3) and requires a local upstream that is unavailable here

### DIFF
--- a/src/update.ts
+++ b/src/update.ts
@@ -113,6 +113,20 @@ async function findInstallDir(packageName: string): Promise<string | undefined> 
     }
 }
 
+/** True when `dir` is a git working tree: `.git` present as a directory
+ *  (normal clone) or as a file (linked worktree / submodule pointer). npm
+ *  installs never contain one — `npm pack` strips VCS metadata — so its
+ *  presence marks a source checkout, not an install. Exported for tests.
+ *  (#580) */
+export async function isGitWorkingTree(dir: string): Promise<boolean> {
+    try {
+        await access(path.join(dir, ".git"));
+        return true;
+    } catch {
+        return false;
+    }
+}
+
 /** Read the version from the on-disk package.json (not the startup constant). */
 async function readDiskVersion(installDir: string): Promise<string | undefined> {
     try {
@@ -335,6 +349,18 @@ export async function checkForUpdate(opts: UpdateOptions, force = false): Promis
         }
         await writeLastCheck(now);
         firstCheckDone = true;
+
+        // Source-checkout guard (#580): findInstallDir() walks up from the
+        // running dist/ and lands on the repo root when bili runs from a git
+        // clone (node dist/index.js start). An in-place tarball copy would
+        // silently rewrite tracked files (the version pin, READMEs), so refuse
+        // to self-update here instead of proceeding.
+        const installDir = await findInstallDir(opts.packageName);
+        if (installDir && await isGitWorkingTree(installDir)) {
+            loggerLog("info", `[update] running from a source checkout (${installDir}) \u2014 skipping auto-update (use npm install -g ${opts.packageName})`);
+            return;
+        }
+
         loggerLog("info", `[update] checking npm registry for ${opts.packageName}${sinceLastSec < 0 ? " (startup check)" : sinceLastSec === 0 ? "" : ` (last check ${sinceLastSec}s ago)`}\u2026`);
 
         const url = `${REGISTRY_BASE}/${opts.packageName}/latest`;
@@ -358,7 +384,6 @@ export async function checkForUpdate(opts: UpdateOptions, force = false): Promis
 
         // Read current version from disk (not from startup constant) so that
         // a successful in-place update is detected without a restart.
-        const installDir = await findInstallDir(opts.packageName);
         const diskVersion = installDir ? await readDiskVersion(installDir) : undefined;
         const currentVersion = diskVersion ?? opts.currentVersion;
 
@@ -451,6 +476,14 @@ export async function installViaTarball(
         await access(installDir, constants.W_OK);
     } catch {
         return { ok: false, error: `install dir not writable: ${installDir}` };
+    }
+
+    // Source-checkout guard (#580): a git working tree must never be merged
+    // over by a published tarball — that rewrites tracked files. checkForUpdate
+    // filters these out already; this keeps the refusal structural for direct
+    // callers.
+    if (await isGitWorkingTree(installDir)) {
+        return { ok: false, error: `install dir is a git working tree (${installDir}) \u2014 refusing to overwrite a source checkout (use npm install -g)` };
     }
 
     // Download tarball. Stream into memory with a hard size cap so a corrupt
--- a/tests/update-tarball.test.ts
+++ b/tests/update-tarball.test.ts
@@ -5,7 +5,7 @@ import { mkdtempSync, mkdirSync, writeFileSync, readFileSync, rmSync, existsSync
 import { tmpdir } from "node:os";
 import path from "node:path";
 import * as tar from "tar";
-import { installViaTarball, declaredEntryRelPaths } from "../src/update.ts";
+import { installViaTarball, declaredEntryRelPaths, isGitWorkingTree } from "../src/update.ts";
 
 function integrityField(buf: Buffer, alg = "sha512"): string {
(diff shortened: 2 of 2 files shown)


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 **Reviewed — ✅ approved** (fixes #580). I reproduced the pre-flight independently on the PR head (`6c1eb8e`) rather than trusting the description, and exercised the actual incident path end-to-end.

## What I verified (independently)

- **typecheck** — pass.
- **`tests/update-tarball.test.ts`** — 6/6 pass, including both new tests (`isGitWorkingTree` dir/file/absent; `installViaTarball` git-tree refusal with byte-for-byte integrity asserts on `package.json` version / README / dist entry).
- **full suite** — 1065/1066. The single failure is `launcher.test.ts → resolveClientCommand: codex/claude resolve to themselves`. This PR touches **no** launcher files (`git diff --name-only` confirms only `src/update.ts` + its test), so the failure is independent of this change — it reproduces identically without it and is purely environmental (those binaries happen to be on PATH in this sandbox). CI runners don't have them, which is why CI is green while a local run shows 1065/1066. Not a regression.
- **build** — success, self-contained `dist/index.js`.
- **startup-path guard (the gap the unit tests don't cover)** — I called `checkForUpdate(…, force=true)` from inside a real git checkout. It logged `[update] running from a source checkout (…) — skipping auto-update (use npm install -g billion-context)` and returned **before any registry fetch**, exiting cleanly. This is exactly the #580 incident path, now confirmed fixed at runtime, not just by unit tests.

## Code review findings

- `isGitWorkingTree()` uses an existence-only check (`access`, default `F_OK`) on `.git`, so it correctly covers both a `.git` **directory** (clone) and a `.git` **file** (linked worktree / submodule pointer). npm installs never ship `.git` (`npm pack` strips VCS metadata), so the marker is precise for the reported failure mode. Good call rejecting the fragile "under global npm prefix" alternative.
- Guard placement in `checkForUpdate()` is after the throttle gate and before the registry fetch → covers both the periodic loop **and** manual `bili update` (`force=true` bypasses throttle but still hits the guard). The relocated `installDir` declaration is cleanly reused downstream — no duplicate `findInstallDir()` call, no dangling reference.
- `installViaTarball()` structural refusal sits after the writability check → defense-in-depth for direct callers of the exported function.
- Behavior for genuine non-git installs (npm global, manual extraction) is unchanged, as intended.

- While running from a source checkout, the skip line re-logs every 3 min. That's consistent with existing periodic info logs (e.g. the "up to date" line also repeats each cycle), so I'd leave it — but flagging in case you'd prefer once-per-process.
- The documented monorepo sub-package limitation (a `packages/billion-context` root has no `.git` of its own) is correctly scoped out-of-scope and rare for this project.

E2E suite not run — agreed, justified: the change is confined to `src/update.ts` (not the request pipeline per AGENTS.md §3) and needs a local upstream unavailable here.

Per repo rules, merging is human-only — I won't merge. All checks above are green; ready for you to merge whenever.

PR #610, opened 2026-09-07
fix: salvage single-quoted compress args before hard rejection (#603)

### LINKED ISSUE #603: lenient compress-arg parsing: salvage salvageable model output

ranxianglei:
## What

`parseCompressInput` (src/compress-tool.ts) now retries through a quote-normalization pass when acp-kernel's `parseCompressArgs` recovers zero ranges or reports invalid items, instead of hard-rejecting.

Closes #603 (reported via omp#121).

## Why

The kernel 0.0.54 salvage ladder already covers fences, trailing commas, raw newlines in strings, double-stringified args, and truncated/prose-wrapped content arrays — verified empirically against each class. The one malformation named in the issue that still hard-fails end-to-end is **single quotes** (`{'content':[{'startId':...}]}`), which is exactly what weak local models emit. Today that lands `[Compression FAILED: no valid ranges parsed (kind=malformed-json...)]`, wastes the round, and the FAILED result can trigger tag-echoing.

## How

- `normalizeSingleQuotes`: state machine converting single-quoted strings to double-quoted JSON strings. Apostrophes inside double-quoted values are data and are copied verbatim; control chars inside single-quoted regions become JSON escapes (`\n`/`\r`/`\t`); embedded `"` escaped. Returns undefined when nothing changed, so valid input never triggers a retry.
- Applied to both shapes where quote text survives: raw-string args (OpenAI/Responses function arguments, text-trigger path) and object inputs whose `content` value is a stringified array (non-strict providers).
- Retry wins only when it recovers strictly more ranges than the first pass; salvaged ranges then pass the same ref-validation gate in `applyCompression`, so worst case is a wasted round — never a wrong compression.
- Each recovery logs `[acp-compress-input] quote-salvage: recovered N range(s) after single->double quote normalization (was K, kind=...)` for attribution.

Host-side deliberately (issue option A): no acp-kernel release cycle needed, ships to all three wires at once. Long-term home is the kernel so proxy/omp/pi converge on one parser — follow-up issue filed on acp-kernel.

Note: PR #188 attempted this against the pre-K1 host parser and a kernel salvage API that no longer exists in 0.0.54 (salvage was folded into `parseCompressArgs` itself); it also pins a local file: tarball. This PR supersedes it — left open here for owner's call on closing.

## Pre-flight

- `npm run typecheck` — clean
- `npm test` — 1147/1148 pass; the single failure (`launcher.test.ts resolveClientCommand: codex/claude resolve to themselves`) fails identically on clean master in this environment (sandbox has `/usr/bin/codex` on PATH) — pre-existing, unrelated
- `npm run build` — success
- New tests (tests/basic.test.ts): full single-quoted object args (+topic +callId propagation), mixed single/double quotes with apostrophe preserved in data, truncated single-quoted array recovers complete entry, object input with single-quoted stringified content, prose-with-apostrophes still rejected unchanged, valid double-quoted args untouched by salvage


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 Reviewed the PR and refreshed it against current master — it was `behind` (base predates the acp-kernel 0.0.56 bump from #607 and the exit-matrix work from #608), so I merged `origin/master` into the branch and pushed: `1ccb180..59a9de1` on `2026-09-07_lenient-compress-args`. The merge was conflict-free (master touched none of `src/compress-tool.ts`, `tests/basic.test.ts`, or `CHANGELOG.md` since the base), which also means the retry logic now runs against kernel 0.0.56's `parseCompressArgs` diagnostics — verified below.

**Review notes**
- Retry only wins when it recovers *strictly more* ranges than the first pass, so valid input is never rewritten and worst case stays a wasted round; salvaged ranges then pass the same ref-validation gate in `applyCompression`. Correct as described.
- `normalizeSingleQuotes` is a pure text repair (no structure invented): apostrophes inside double-quoted values copied verbatim, control chars inside single-quoted regions become JSON escapes, returns undefined when nothing changed. The prose-with-apostrophes rejection case is covered by a test and unchanged.
- Host-side placement is fine for shipping now; the kernel follow-up (#211) keeps proxy/omp/pi convergence on the agenda.

**Pre-flight on the merged branch (kernel 0.0.56)**
- `npm run typecheck` — clean
- `npm test` — 1183/1184 pass; all six new `#603` salvage tests in `tests/basic.test.ts` pass. The single failure is `tests/launcher.test.ts:24` `resolveClientCommand: codex/claude resolve to themselves` — environmental: this sandbox has `/usr/bin/codex` on PATH so the launcher resolves the absolute path instead of bare `codex`. Pre-existing on clean master here, unrelated to this change.
- `npm run build` — success

Ready for merge when you are (merge is your call per repo rules). Reminder: PR #188 is superseded by this one and can be closed at your discretion.

ranxianglei:
**Review vs master c8c825b:** ✅ merge-ready. Verified locally: typecheck ✓, `tests/basic.test.ts` 29/29 (incl. the 6 new #603 cases).

- State machine reviewed line-by-line: apostrophes inside double-quoted regions are data (copied verbatim) ✓; `\'` inside single regions collapses to a raw apostrophe ✓; control chars inside single regions become JSON escapes ✓; unterminated single leaves invalid JSON → falls through to the existing reject path ✓; prose guard (`no { and no [`) prevents touching plain sentences ✓.
- Salvage wins only when it recovers **strictly more** ranges than the first pass — valid input is never rewritten, partial kernel recovery is never traded for a worse retry ✓.
- Retry trigger covers both shapes: raw-string args and `{content: "..."}` object inputs.
- Salvaged ranges still go through the same ref-validation gate — worst case stays "wasted round", never "wrong compression" ✓.

Scope note (non-blocking): this covers the **proxy path**. Plugin agents (pi/omp) parse compress args locally via kernel `parseCompressArgs` directly — same malformation class would still hard-fail there. Kernel-side salvage would fix both; host-side is the right first step.

(Closes #603 on merge)

ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 Thanks for the line-by-line review. Confirmed the branch is fully in sync with `origin/master` (0 commits behind) and mergeable. Per repo rules I can't merge PRs myself — please merge when ready: https://github.com/ranxianglei/billion-context/pull/610

On the scope note: agreed, host-side covers the proxy path only; plugin agents (pi/omp) parse locally via kernel `parseCompressArgs`, so the same single-quote class still hard-fails there until the kernel-side salvage lands (tracked in acp-kernel#211). And #603 will close automatically on merge via the `Closes #603` reference.

Issue #583, opened 2026-09-06
compat auto-fix: developer→system retry can still 400 on system-at-index-0 backends (#377 class); add second-chance developer→user learn


ranxianglei:
## Context

#552 / PR#579 shipped a learn-on-failure auto-fix for role-rejecting upstreams: on a 400 like `Invalid role: developer`, bili rewrites the role to `system`, retries **once**, and on success remembers the mapping in `session.metadata.learnedCompatRoles` (session-scoped; config file untouched). If the retry fails, the original 400 is passed through verbatim.

## The edge

An upstream can enforce **both**:

1. unknown roles are rejected (`developer` → 400), **and**
2. `system` messages are only accepted at index 0 (SGLang-style strict placement — the same class as #377).

If the `developer` message sits **mid-list** (codex sends `developer` role messages in Responses `input[]`), the auto-retry rewrites it to a **mid-list `system`** → the retry 400s again (a #377-class placement error, not a role-name error) → the original error passes through verbatim. No loop, no corruption — but this sub-case stays broken and the auto-learn never engages.

## Scope

- **Responses path only.** The chat path is unaffected: the rebuild chain `systemToUser(coreToOpenai(...))` (`src/server.ts:1864`, `src/loop/adapter-openai.ts:193`) already normalizes mid-list `system`/`developer` to `user` before the wire, so the compat rewrite is a no-op there.
- **Existing escape hatch**: manual config `compat.roles {"developer": "user"}` already covers such upstreams.

## Suggested fix

Second-chance learning, mirroring the existing one-retry discipline:

- Teach `detectRoleRejection` (`src/compat-roles.ts`) a conservative pattern for *system placement* errors (e.g. `system message only at index 0`, `multiple system messages`).
- When the first retry (→ `system`) fails with that error class, do **one** more controlled retry with `{role: "user"}`.
- If that succeeds, learn `developer → user` for the session (position-class learning wins over the generic one).
- Hard cap stays at: original attempt + 2 retries, no loops; anything still failing passes the original 400 through verbatim.

Found while answering a maintainer question about the relationship between #377 and #552 after PR#579 merged.

ranxianglei:
> 🤖 ework agent · qwen3.8-27b

Implemented your suggested fix and opened a PR: **https://github.com/ranxianglei/billion-context/pull/585**

## What I verified first (triage)

[...]

PR #612, opened 2026-09-07
test: wire-exit gap cells — preflight-emitter shapes, post-commit JSON errors, abort-storm convergence

ranxianglei:
Refs #588

#608 shipped the exit-matrix registry plus its first behavior cells. This PR fills the cells that were verified uncovered at filing time — each one confirmed by reading every file referenced in `coveredBy` before writing a test:

**Error delivery**
1. `emitPreflightError` unit shape matrix, all 3 protocols × retryable{true,false}. `tests/proxy-stream-error.test.ts` pins `emitStreamError` exhaustively, but the preflight emitter had no pin at all — including the openai-only `retryable` asymmetry.
2. Non-stream preflight failure AFTER early commit → `{error}` JSON on the already-committed 200. The non-stream cell in `tests/preflight-hold.test.ts` only asserts the success path ("then valid JSON"); the registry row claimed failure-shape coverage that did not exist.
3. Upstream non-2xx AFTER early commit (non-stream) → verbatim upstream error body under the committed 200. Asserted nowhere.

**Abort propagation**
4. Repeated-abort convergence storm on proxy-openai-sse (3 rounds): asserts BOTH observables per round — proxy-side `_liveUpstreamTimersForTest()` converging to 0 AND the upstream socket close arriving — polled rather than asserted synchronously. Abort propagation through undici is async (measured t+0 → t+600ms window); a sync assert races it.
5. Client-gone robustness: `emitPreflightError` must not throw when writes fail (all 3 protocols).

`src/exit-matrix.ts`: additive `coveredBy` entries registering the new file on 5 rows (4 ERROR_DELIVERY rows + ABORT_PROPAGATION/proxy-openai-sse). No contract or key changes.

## Pre-flight
- `npm run typecheck`: clean
- New file: 5/5 pass
- Full suite: 1182/1183 — the single failure (`launcher.test.ts` resolveClientCommand) is pre-existing and environment-dependent: it fails identically on clean master in environments where `codex` is installed (resolver returns the absolute path; the test expects the bare name). Unrelated to this diff.

Note: the storm test's upstream handler needs a request-body consumer (`req.resume()`) — the kernel-augmented payload is ~11KB even for a one-word prompt, and without a consumer Node keeps the stream paused, stalling the forward. Documented inline to prevent "cleanup" removal.

### DIFF
--- a/src/exit-matrix.ts
+++ b/src/exit-matrix.ts
@@ -36,22 +36,22 @@ export const ERROR_DELIVERY: Record<WireExitId, ExitCell> = {
     "proxy-openai-sse": {
         implementer: [emitStreamError, emitPreflightError, startServer],
         contract: "mid-stream upstream failure → inline `error` delta + finish + [DONE]; late (early-committed) preflight failure → top-level error object + [DONE] in-band",
-        coveredBy: ["tests/proxy-stream-error.test.ts", "tests/wire-exit-matrix.test.ts"],
+        coveredBy: ["tests/proxy-stream-error.test.ts", "tests/wire-exit-matrix.test.ts", "tests/wire-exit-gap-cells.test.ts"],
     },
     "proxy-anthropic-sse": {
         implementer: [emitStreamError, emitPreflightError, startServer],
         contract: "mid-stream failure → content_block_delta error + message_stop; late preflight failure → event: error payload in-band",
-        coveredBy: ["tests/proxy-stream-error.test.ts", "tests/preflight-hold.test.ts"],
+        coveredBy: ["tests/proxy-stream-error.test.ts", "tests/preflight-hold.test.ts", "tests/wire-exit-gap-cells.test.ts"],
     },
     "proxy-responses-sse": {
         implementer: [emitStreamError, emitPreflightError, startServer],
         contract: "mid-stream failure → full item lifecycle (added → delta → done) + response.failed; late preflight failure → event: error in-band",
-        coveredBy: ["tests/proxy-stream-error.test.ts", "tests/preflight-hold.test.ts"],
+        coveredBy: ["tests/proxy-stream-error.test.ts", "tests/preflight-hold.test.ts", "tests/wire-exit-gap-cells.test.ts"],
     },
     "proxy-json": {
         implementer: [emitPreflightError, startServer],
         contract: "pre-headers failure → 4xx/5xx `{error}` JSON; post-early-commit failure → identical `{error}` JSON body on the already-committed 200",
-        coveredBy: ["tests/preflight-hold.test.ts", "tests/preflight-fail-fast.test.ts"],
+        coveredBy: ["tests/preflight-hold.test.ts", "tests/preflight-fail-fast.test.ts", "tests/wire-exit-gap-cells.test.ts"],
     },
     "plugin-chat-sse": {
         implementer: [pipePluginChatWithStrip],
@@ -74,7 +74,7 @@ export const ABORT_PROPAGATION: Record<WireExitId, ExitCell> = {
     "proxy-openai-sse": {
         implementer: [startServer],
         contract: "client disconnects mid-stream → the in-flight upstream request is destroyed (no orphaned summarization/forward), session lock released",
-        coveredBy: ["tests/wire-exit-matrix.test.ts"],
+        coveredBy: ["tests/wire-exit-matrix.test.ts", "tests/wire-exit-gap-cells.test.ts"],
     },
     "proxy-anthropic-sse": {
         implementer: [startServer],
--- a/tests/wire-exit-gap-cells.test.ts
+++ b/tests/wire-exit-gap-cells.test.ts
@@ -1,0 +1,313 @@
+import assert from "node:assert/strict";
+import http from "node:http";
+import { once } from "node:events";
+import test from "node:test";
+
+process.env.NODE_ENV = "test";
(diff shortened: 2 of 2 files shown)


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

**PR #612 reviewed — all claims verified against source, empirical checks reproduce. LGTM.**

**Duplicate screening**: not a duplicate — this is the planned follow-through of #588 (which proposed the exit-enumeration test matrices); #608 shipped the registry + first cells and left exactly these as verified-uncovered. No other open issue/PR covers any of the five cells.

**Claim-by-claim verification (read every referenced file before running anything):**

1. **emitPreflightError unit shapes** — confirmed zero prior pin: `tests/proxy-stream-error.test.ts` imports only `emitStreamError`, and no other test calls the preflight emitter. The new 3-protocol × retryable{true,false} matrix matches `src/stream-error.ts:84-103` exactly, including the openai-only `retryable` asymmetry (openai embeds the full `err` object; anthropic/responses emit minimal `event: error` payloads without it).
2. **Non-stream preflight failure after early commit** — confirmed gap: `tests/preflight-hold.test.ts:282` (non-stream cell) asserts the success path only; both 429 cells there are stream-only. Gap 3 matches the real branch at `src/server.ts:1600-1609` (`{error}` JSON under the committed 200).
3. **Upstream non-2xx after early commit (non-stream)** — confirmed unasserted: `tests/wire-exit-matrix.test.ts:269` covers post-commit upstream 429 only for `stream: true` (the `emitStreamError` path). Gap 4 matches `src/server.ts:3439-3447` (`res.end(errBody)` verbatim under the committed 200).
4. **Abort-storm convergence** — `wire-exit-matrix` asserts one-shot socket destruction per exit with a fixed 300 ms sleep; timer convergence across *repeated* aborts was unasserted. Gap 5 polls BOTH observables (`_liveUpstreamTimersForTest() === 0` AND upstream close) over 3 rounds — correct design, since undici abort propagation is async and a sync assert would race it.
5. **Client-gone robustness of emitPreflightError** — genuinely new: `tests/proxy-stream-error.test.ts:81` pins write-throws only for `emitStreamError`.

**Registry diff**: additive `coveredBy` on exactly 5 rows (4 ERROR_DELIVERY + ABORT_PROPAGATION/proxy-openai-sse), no contract or key changes — matches the PR body. The meta-test in `wire-exit-matrix.test.ts:175` (fails if any `coveredBy` file goes stale) passes.

**Empirical (branch e221c75):**
- `npm run typecheck`: clean
- `tests/wire-exit-gap-cells.test.ts`: **5/5 pass** (~7 s)
- Full suite: **1182/1183**. The single failure (`launcher.test.ts` "resolveClientCommand: codex/claude resolve to themselves") reproduces **identically on clean master** (2b510af, detached worktree) in this environment — `codex` is installed here, so the resolver returns `/usr/bin/codex` while the test expects the bare name. Pre-existing and environment-dependent, unrelated to this diff. Claim accurate.

**Code quality**: no `as any`/`@ts-ignore`; the inline comments encode load-bearing timing rationale (async abort window, the `req.resume()` drain note) — justified under the "absolutely necessary" bar; file follows the established `preflight-hold`/`wire-exit-matrix` patterns. Test-only change, so the two-compression-mode rule doesn't apply.

**Minor note (not blocking)**: the `launcher.test.ts` codex-path environment-dependence is worth a small separate fix (skip or normalize when `codex` resolves to an absolute path) so suites stay green on dev machines with codex installed. Reporting here only, per convention — no separate issue for something this small.

Verdict: every claimed gap was a real gap, and each new assertion pins actual source behavior rather than an invented contract. Ready to merge — and per repo rules the merge itself stays human-only: https://github.com/ranxianglei/billion-context/pull/612

ranxianglei:
**Review vs master c8c825b:** ✅ merge-ready (test-only + matrix annotations). Verified locally: typecheck ✓, `wire-exit-matrix` + `wire-exit-gap-cells` 12/12.

- The 5 gap tests target exactly the cells #588's survey called out: emitPreflightError channel/payload shape per protocol (gap 1), client-gone-mid-write resilience (gap 2), post-commit structured vs verbatim JSON under the already-committed 200 (gaps 3–4), abort-storm → upstream destroyed + timers converge to zero (gap 5 — same class as #583's timer leak).
- `coveredBy` annotations updated for the 5 affected exits — the matrix's self-consistency test enforces these are real files.
- Zero production-code behavior change (only annotation arrays).

(Refs #588 — partial: the plugin-side dual-path cells remain open by design)

ranxianglei:
> 🤖 ework agent · qwen3.8-27b

Agreed on all counts — my independent pass reached the same verdict (claim-by-claim verification and empirical results are in my review above). One extra data point from the freshness check just now: after re-fetching, `origin/master` (1330fc5) is a **direct ancestor** of the PR head `e221c75` — `git log pr-612..origin/master` is empty, so nothing has landed on master since the branch point. The merge will be conflict-free as-is; no rebase or re-verification needed.

Nothing further outstanding on my side: all five gap cells verified real before and now pinned, registry annotations consistent (the meta-test enforces it), and the sole suite failure (`launcher.test.ts` codex-path resolution) reproduces identically on clean master — environment-dependent, unrelated to this diff. The plugin-side dual-path cells staying open under #588 by design matches what I saw too.

Merge is yours whenever you're ready: https://github.com/ranxianglei/billion-context/pull/612

PR #613, opened 2026-09-07
fix: route updater egress through configured upstream proxy (#609)

### LINKED ISSUE #609: updater egress bypasses configured upstream proxy (registry + tarball fetches)
来源: #606 (https://github.com/ranxianglei/billion-context/issues/606) 分析外部提案 PR #86 (https://github.com/ranxianglei/billion-context/pull/86) 时发现。

### Problem
The auto-updater's two outbound fetches bypass bili's upstream-proxy machinery entirely:
- `src/update.ts:367` — `fetch("https://registry.npmjs.org/<pkg>/latest")` (no dispatcher)
- `src/update.ts:494` — tarball download (no dispatcher)

undici's global fetch does not honor `HTTPS_PROXY`/`HTTP_PROXY` env vars, and no `dispatcher` is passed, so these connections always go direct. Meanwhile every model-traffic path in bili resolves a proxy decision (`src/upstream-proxy.ts`: `resolveProxyDecision()` → `proxyDispatcher()`, precedence per-provider → `BILI_UPSTREAM_PROXY` → web-manual → config `proxy` → env → Windows system proxy → direct, `NO_PROXY` honored).

### Impact
On deployments where npm is only reachable through a proxy (the GFW scenario the README explicitly supports for model traffic), every update check fails with a warn log and auto-update silently never works — users sit on old versions indefinitely, with no visible signal beyond `[update] check failed:` lines.

### Suggested fix
Route both updater fetches through the same decision path: resolve the proxy for the registry origin (and the tarball host) and pass `dispatcher: proxyDispatcher(proxyUrl)` to the `fetch` calls. `startAutoUpdate`/`checkForUpdate` currently take only `{packageName, currentVersion, autoUpdate}` (`src/cli.ts:408-409`), so the proxy inputs (routes, globalProxy, fallback options) need threading — or a small resolver callback. Honor `NO_PROXY` via `matchesNoProxy`.

### Constraints
- Per AGENTS.md §5, any change to `src/update.ts` download logic must be validated with a no-op release before shipping.
- Keep orthogonal to the three-state update-mode discussion (#606): this fix is independent and smaller-risk than adding a new mode.

Origin: bundled in LX666-666's PR #86 alongside a three-state update mode; the mode half is being declined (see #606), the egress half stands on its own merits.

ranxianglei:
## What

The auto-updater's two outbound fetches (npm registry version check, tarball download) now go through bili's upstream-proxy decision chain — the same one model traffic uses — instead of always connecting direct.

## Why

undici's global `fetch` ignores `HTTP(S)_PROXY`/`ALL_PROXY` env vars unless an explicit `dispatcher` is passed. `src/update.ts:367` (registry) and `src/update.ts:494` (tarball) passed none, so on deployments where npm is reachable only through a proxy (the GFW scenario the README supports for model traffic), every check failed with `[update] check failed:` and auto-update silently never worked.

## Changes

- `src/update.ts`
  - `UpdateOptions` gains an optional resolver callback `resolveProxy?: (url: string) => string | undefined` (keeps update.ts decoupled from config details; absent = direct).
  - Both fetches pass `dispatcher: proxyDispatcher(resolveProxy(url))` when the decision yields a proxy. Per-URL resolution, so the registry origin and the tarball host each get their own decision; `NO_PROXY` is honored inside `resolveProxyDecision`.
  - `installViaTarball` gains a trailing optional `dispatcher` param (existing call sites/tests unaffected).
- `src/cli.ts`
  - Both entry points wire the resolver: startup auto-update and manual `bili update`. The closure reads `opts.routes`/`opts.proxy`/`opts.proxyFallback` per call, so web-UI config hot-reload (server.ts mutates opts in place) is picked up live without restart.
  - Latent bug fixed along the way: the `update` command branch ran **before** the flag/env merge loop, so `-F <proxy>` was silently ignored for manual updates. Flags are now applied first; if config loading fails, updater egress goes direct with a visible CLI error (same effective behavior as before, now explicit).
- `README.md` — documents updater egress coverage in the upstream-proxy section.
- `tests/update-proxy.test.ts` (new, 4 tests) — dispatcher presence/absence at both fetch sites, cached-agent identity with `proxyDispatcher`, NO_PROXY-style non-matching hosts.

## Pre-flight

- `npm run typecheck`: clean
- `npm test`: 1181/1182 — the single failure is pre-existing and env-dependent (`tests/launcher.test.ts` "codex/claude resolve to themselves"; reproduced on the pristine tree in this sandbox, which has `/usr/bin/codex` on PATH)
- `npm run build` (tsup): success
- E2E smoke: isolated install dir + local recording proxy on 127.0.0.1, config `{"proxy": "http://127.0.0.1:47999", "upstreamProxyMode": "auto"}` → `CONNECT registry.npmjs.org:443` observed arriving at the proxy (previously went direct); with the proxy socket destroyed the check degrades gracefully to the usual warn log, no crash.

## ⚠ Release gate (AGENTS.md §5)

This PR touches `src/update.ts` download logic, so per §5 it must be validated by a no-op release **before** merging. v0.1.89 is already live on npm (released from master after this branch's base). If a real-world auto-update to 0.1.89 has been observed succeeding, that validates the existing upgrade path and this PR can merge as-is. Otherwise I'll prepare a no-op v0.1.90 release PR to land first — owner's call on sequencing. No PR is merged by the agent either way.

Fixes #609

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,6 +344,9 @@ Rules:
 - Both outbound paths are covered: `/bili/` path-mode (fetch) AND MITM CONNECT
   tunnels (the proxy's connection to the real upstream goes through the HTTP
   CONNECT proxy).
+- The auto-updater's own egress (npm registry check + tarball download) uses
+  the same decision for its hosts, so `bili update` and auto-update work on
+  hosts where npm is only reachable through the proxy (#609).
 
 Env override: `BILI_UPSTREAM_PROXY=http://127.0.0.1:20172` (higher priority than
 the config file). On Windows, common Clash/Mihomo static system proxies are
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -24,6 +24,7 @@ import { loadOptions, ensureConfigTemplate } from "./config.js";
 import { startServer } from "./server.js";
 import { configFile as defaultConfigFile } from "./paths.js";
 import { checkForUpdate, startAutoUpdate } from "./update.js";
+import { resolveProxy } from "./upstream-proxy.js";
 import { runMcpStdio } from "./mcp.js";
 import { PLUGIN_AGENTS, isPluginAgent, pluginInstall, pluginRemove, pluginStatusAll, type PluginAgent } from "./plugin-install.js";
 import { runLaunch, runTestPi, isLaunchClient, type ClientName } from "./launcher.js";
@@ -374,8 +375,24 @@ export async function main(): Promise<void> {
         return;
     }
     if (command === "update") {
-        // Manual one-shot update — bypasses the throttle.
-        await checkForUpdate({ packageName: PACKAGE_NAME, currentVersion: VERSION, autoUpdate: true }, true);
+        // Manual one-shot update — bypasses the throttle. Apply flag
+        // overrides first so `-F <proxy>` reaches loadOptions; the registry
+        // and tarball egress then honor the same upstream-proxy decision as
+        // model traffic (#609).
+        for (const [k, v] of Object.entries(overrides)) {
+            if (v !== undefined) process.env[k] = v;
+        }
+        let updaterResolveProxy: ((url: string) => string | undefined) | undefined;
+        try {
+            const o = loadOptions();
+            updaterResolveProxy = (url) => resolveProxy(o.routes, o.proxy, url, o.proxyFallback);
+        } catch (e) {
+            console.error(`bili update: config load failed (${String(e)}); updater egress goes direct`);
+        }
+        await checkForUpdate(
+            { packageName: PACKAGE_NAME, currentVersion: VERSION, autoUpdate: true, resolveProxy: updaterResolveProxy },
+            true,
+        );
         return;
     }
     if (command === "test") {
@@ -406,6 +423,13 @@ export async function main(): Promise<void> {
     // Start background auto-update after the server is listening so a slow
     // registry check never delays startup or races the listen socket.
     if (opts.autoUpdate) {
-        startAutoUpdate({ packageName: PACKAGE_NAME, currentVersion: VERSION, autoUpdate: true });
+        // Resolver reads opts fields per call, so web-UI hot-reload of proxy
+        // settings (server.ts mutates opts in place) is picked up live (#609).
+        startAutoUpdate({
+            packageName: PACKAGE_NAME,
+            currentVersion: VERSION,
+            autoUpdate: true,
+            resolveProxy: (url) => resolveProxy(opts.routes, opts.proxy, url, opts.proxyFallback),
+        });
     }
 }
--- a/src/update.ts
+++ b/src/update.ts
@@ -28,6 +28,8 @@ import path from "node:path";
 import { fileURLToPath } from "node:url";
 import { cacheDir } from "./paths.js";
 import { log as loggerLog } from "./logger.js";
+import { proxyDispatcher } from "./upstream-proxy.js";
+import type { FetchOptions } from "./fetch-util.js";
 
 const REGISTRY_BASE = "https://registry.npmjs.org";
 const CHECK_INTERVAL_MS = 3 * 60 * 1000;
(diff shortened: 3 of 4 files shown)


ranxianglei:
**Review vs master c8c825b:** ✅ code is correct — but **merge sequencing constraint applies** (see last point). Verified locally: typecheck ✓, `tests/update-proxy.test.ts` 4/4.

- Root cause correct: undici's global `fetch` ignores `HTTP(S)_PROXY` env — updater egress (registry + tarball) always went direct even when model traffic routes through the configured proxy. Behind a corporate/restricted egress this means silent update-check failures.
- Wiring is the right shape: `egressDispatcher` → `proxyDispatcher(resolveProxy(routes, proxy, url, proxyFallback))` reuses the exact decision chain as model traffic (incl. NO_PROXY semantics); resolver closure reads `opts` **per call** so web-UI hot-reload of proxy settings is honored live.
- `bili update` CLI path applies flag overrides to env before `loadOptions()` — consistent with the `cli.ts` override mechanism ✓; config-load failure degrades to direct with a clear error (cli.ts is a legitimate `console.error` site).
- `FetchOptions` (fetch-util.ts:63) already types `dispatcher?: object`; single structural cast, no `as any` ✓.

⚠️ **AGENTS.md §5 hard rule**: this changes `src/update.ts` fetch/install logic → it must NOT be the first hop of an unvalidated upgrade path. Sequence: (1) ship a **no-op release** (e.g. v0.1.90, version-only) and observe one real auto-upgrade succeed, (2) only then merge this and ship it in v0.1.91. Alternatively cite an observed auto-upgrade on the current path if one is already logged. Don't bundle this into the next release directly.

(Closes #609 on merge)

ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 Independent review complete — I re-ran everything locally against `cc74daf` rather than trusting the PR body. Verdict: **code is correct**; for the §5 gate I can now supply evidence for your "cite an observed auto-upgrade" alternative (bottom of this comment).

**Verification (own runs on the PR head):**
- `npm run typecheck`: clean · `npm run build` (tsup): success
- `npm test`: **1181/1182** — the single failure, `resolveClientCommand: codex/claude resolve to themselves` (`tests/launcher.test.ts`), is pre-existing and env-dependent: this PR touches neither `tests/launcher.test.ts` nor `src/launcher.ts`, and it fails here because this sandbox has `/usr/bin/codex` on PATH
- `tests/update-proxy.test.ts`: 4/4 in isolation

**Line-level confirmation of each claim:**
- Root cause holds: undici's global `fetch` without an explicit `dispatcher` ignores `HTTP(S)_PROXY` — both updater fetches (registry check, tarball download) always went direct even when model traffic routes through the configured proxy.
- Wiring is literally the same decision chain as model traffic: `egressDispatcher` → `proxyDispatcher(resolveProxy(routes, proxy, url, proxyFallback))` (`src/upstream-proxy.ts:273`), including NO_PROXY semantics (`matchesNoProxy`, :249); per-URL resolution gives the registry origin and the tarball host independent decisions.
- Safety property worth calling out: **zero behavior change for default-mode users.** Mode defaults to `direct` (`src/config.ts:331`) → `opts.proxy = ""` + `explicitDirect` short-circuits to direct at `src/upstream-proxy.ts:246` before env proxies are consulted. Only deployments that already route model traffic through a proxy see any updater routing change — the updater can never take a route model traffic wouldn't use.
- Hot-reload claim verified: `PUT /__bili/config` mutates the same `opts` object in place (`src/server.ts:829-837`: `proxy` / `proxyFallback` / `routes`), and the cli.ts closure reads those fields per call → live pickup without restart. ✓
- Latent `-F` bug fix confirmed: old `src/cli.ts` ran the `update` branch (:376–378) *before* the flag/env merge loop (:394–395), so `-F <proxy>` never reached `loadOptions()` for manual updates; the new code applies overrides first inside the branch, and config-load failure degrades to direct egress with a visible CLI error (`console.error` is legitimate in cli.ts).
- Cast discipline: `FetchOptions` (`src/fetch-util.ts:63`) already types `dispatcher?: object`; single structural cast in `fetchWithEgress`, no `as any`. Trailing optional `dispatcher` param on `installViaTarball` leaves existing call sites untouched (full suite green).

**§5 gate — recorded auto-upgrade on the current path.** Ran the AGENTS.md "Auto-update testing" procedure in the sandbox against real published artifacts (real npm registry; the upgrading code is 0.1.88's trusted old updater):

```
$ npm install -g billion-context@0.1.88 && bili --version
0.1.88
$ bili start --port 19195
02:44:22Z [update] auto-update enabled (checking every 180s)
02:44:32Z [update] checking npm registry for billion-context …
02:44:32Z [update] new version found: 0.1.88 → 0.1.89, downloading…
02:44:34Z [update] installed 0.1.88 → 0.1.89. Restart to finish.
$ bili --version   # post-install
0.1.89
```

That exercises exactly the plumbing a no-op release would validate — registry reachability, tarball download + integrity, install-dir detection, file copy — entirely under old trusted code. Stated plainly: this is a sandbox observation, not a production-deployment observation.

**Recommendation:** accept your alternative — cite the upgrade above and merge as-is; the fix then ships as **v0.1.90** (next release after live v0.1.89). If you'd rather have belt-and-suspenders, say so and I'll prepare a no-op v0.1.90 release PR first; this then ships as v0.1.91. Your call — merge stays with you either way (#609 closes via the `Fixes #609` link).

Minor (report-only, non-blocking): the captured-fetch tests exercise the tarball site end-to-end; the registry-check site is covered only indirectly (identical 3-line `egressDispatcher` + spread pattern). Optional follow-up: a captured-fetch test around `checkForUpdate`'s registry call.

PR #662, opened 2026-09-09
feat: per-route passthrough option for providers (#661)

ranxianglei:
## What

New per-route option `passthrough: true` on `providers` entries — the scoped sibling of the global `passthrough` setting:

```jsonc
{ "providers": { "mitm://zcode.z.ai": { "passthrough": true } } }
```

Matching requests are forwarded **byte-for-byte**: no kernel round-trip (no message re-serialization, no ACP render tags, no `prompt_cache_key` removal), response piped verbatim, no session state for that route. `mitm://` keys target only MITM (login-client) traffic; a plain `https://` host key covers both MITM and `/bili/` traffic.

## Why

#661: ZCode's upstream anti-fraud returns `405 / 3012` on the kernel-rebuilt `messages` body (field order/whitespace/`prompt_cache_key` + injected render tags), even with `injectTool=false`/`injectNudge=false`. The user's evidence (same MITM TLS chain: un-rewritten paths 200, rewritten path 405) isolates the body fingerprint as the trigger. Per-route passthrough is the shape of the user's proposal A, built on the existing route table instead of a new top-level `bypassHosts` list: zero new matching mechanism (reuses `findRoute` longest-prefix), exact MITM-vs-/bili/ targeting via the existing `mitm://` scheme, discoverable in the existing providers config surface and web UI.

Not proposal B (skip `processTurn` when no active blocks): the first compress call arrives with zero active blocks, and in plugin mode `processTurn` also materializes the agent's compress tool calls, runs orphan GC, and maintains compaction archive + preflight token tracking — skipping it breaks first compression and state maintenance.

Not proposal C (blind tunnel): body-fingerprint evidence makes kernel bypass sufficient; blind tunnel changes the TLS path and stays as a fallback if 405 persists after this.

## Changes

- `src/config.ts` — `ProviderRoute.passthrough?: boolean` + parsing in `parseRouteEntry` (boolean only, non-boolean values ignored)
- `src/server.ts` — `routePassthrough = findRoute(opts.routes, route?.rewrittenUrl)?.passthrough === true`; added to the pipeline gate and the `logUnrecognizedPath` condition; info log when the bypass fires
- `tests/route-passthrough.test.ts` — 6 tests: config round-trip, `mitm://` key matching (incl. lookalike-host negative), e2e byte-identical forward (incl. `prompt_cache_key` + whitespace), e2e control (no-route still kernel-processed: re-serialized, tags, `prompt_cache_key` removed), e2e mixed (both behaviours in one server)
[...]

PR #664, opened 2026-09-09
fix: stop following upstream redirects on the client's behalf (#661)

ranxianglei:
## Problem (issue #661, second symptom)

With bili proxy + CA in front of an upstream that redirects (CDN/WAF), requests failed with **HTTP 405**.

## Root cause

`fetchWithTimeout` (`src/fetch-util.ts`) never set the `redirect` option, so undici's global `fetch` used its default `redirect: "follow"`. Per the fetch spec, following a **301/302/303 on a POST downgrades the request to GET and drops the body**. So when the upstream answered the POST with a redirect, bili re-issued it as a GET at the redirect target — an API endpoint that only accepts POST — which returned **405 Method Not Allowed**.

A forward proxy must not follow redirects on the client's behalf: the 3xx (status + `Location`) must be passed through and the client follows it with its own policy.

## Changes

- `src/fetch-util.ts` — `fetchWithTimeout` now defaults `redirect` to `"manual"` (the proxied request path: `forward`, compat retry, fake-completion retry, compress-loop / preflight / loop re-sends all inherit this).
- `src/server.ts` — the internal `POST /__bili/upstream/test` probe explicitly opts into `redirect: "follow"` (behavior unchanged).
- `src/registry.ts` — the internal models.dev registry fetch explicitly opts into `redirect: "follow"` (behavior unchanged).
- `tests/redirect-passthrough.test.ts` (new) — regression tests:
  - POST: upstream answers 302 → client sees **302 + Location** (not 405), the redirect target is **never contacted**, and the forwarded body still carries ACP render tags → **compression stays fully active**.
  - GET: same pass-through behavior.
  - Verified the tests fail with `405` when the default is reverted to `"follow"` (reproduces the reported symptom exactly).

## ACP compression impact

None. The fix is transport-only: the kernel round-trip (`prepareAnthropic` → `processTurn` → `coreToAnthropic`) still runs before the forward; only the redirect policy of the outgoing fetch changed. The new tests assert ACP tags are present in the forwarded body.

## Hypotheses checked (per issue request)

1. **Redirect changes the method → 405** — ✅ confirmed, this is the root cause (fixed above).
2. Header rewrites (Accept-Encoding / Content-Encoding / Content-Length) — ❌ not the cause: `content-length` is stripped and recomputed from the actual body; `content-encoding` is decoded and stripped before forwarding (consistent). Side finding (minor, not fixed here): the #619 unsupported-encoding path forwards an still-encoded body while `content-encoding` is stripped as hop-by-hop, which would confuse the upstream (400-class, not 405).
3. ACP compression changes method/path/headers — ❌ not the case: method is preserved (`req.method ?? "GET"`); ACP only rewrites the body.

## Pre-flight

- `npm run typecheck` ✅
- `npm test` ✅ 1248/1248 (includes 2 new tests)
- `npm run build` ✅
- E2E codex suite not run here (needs a reachable Responses upstream); unit + integration coverage of the redirect path is provided by the new tests.

Related: PR #662 (per-route `passthrough` for body-fingerprinting upstreams) is an independent, complementary fix for the first symptom of #661.

### DIFF
--- a/src/fetch-util.ts
+++ b/src/fetch-util.ts
@@ -129,6 +129,13 @@ export async function fetchWithTimeout(
             ...opts,
             signal: controller.signal,
             dispatcher: opts.dispatcher ?? directDispatcher(effective),
+            // Forward-proxy correctness: never silently follow a redirect.
+            // undici's default (follow) downgrades POST→GET and drops the body
+            // on 301/302/303, so a redirecting upstream (CDN/WAF) turns a valid
+            // POST into a 405 at the redirect target. Pass the 3xx through to
+            // the client, which follows it with its own policy. Internal bili
+            // fetches that want to follow (registry, upstream test) opt in.
+            redirect: opts.redirect ?? "manual",
         };
         // `fetch` is undici's global; it accepts `dispatcher` at runtime. @types/node
         // types RequestInit.dispatcher as its internal `Dispatcher` interface,
--- a/src/registry.ts
+++ b/src/registry.ts
@@ -109,6 +109,7 @@ async function fetchFresh(): Promise<RegistryShape | null> {
             const { response, clearTimer } = await fetchWithTimeout(REGISTRY_URL, {
                 ...opts,
                 headers: { Accept: "application/json" },
+                redirect: "follow",
             }, 15_000);
             const text = await response.text();
             clearTimer();
--- a/src/server.ts
+++ b/src/server.ts
@@ -872,6 +872,7 @@ async function handle(
         try {
             const result = await fetchWithTimeout(targetUrl, {
                 method: "HEAD",
+                redirect: "follow",
                 ...(proxyUrl ? { dispatcher: proxyDispatcher(proxyUrl, 15_000) } : {}),
             }, 15_000);
             result.clearTimer();
--- a/tests/redirect-passthrough.test.ts
+++ b/tests/redirect-passthrough.test.ts
@@ -1,0 +1,188 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import http from "node:http";
+import { once } from "node:events";
+import { defaultConfig } from "acp-kernel";
+import { startServer } from "../src/server.ts";
+import { SessionStore, _setStoreForTest } from "../src/persist.ts";
+import type { ProxyOptions } from "../src/config.ts";
+import { _setForTest as setRegistryForTest } from "../src/registry.ts";
+
+// #661 follow-up: a forward proxy must not silently follow upstream redirects.
+// undici's default (redirect: "follow") downgrades a POST to a GET and drops
+// the body on 301/302/303, so a redirecting upstream (CDN/WAF) turns a valid
+// POST into a 405 at the redirect target. bili must pass the 3xx through to
+// the client, which follows it with its own policy. ACP compression stays on:
+// the kernel round-trip still rewrites the body before the (now non-followed)
+// forward.
+
+function listen(server: http.Server): Promise<void> {
+    if (server.listening) return Promise.resolve();
+    return once(server, "listening").then(() => undefined);
+}
+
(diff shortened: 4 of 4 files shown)


ranxianglei:

[bot] 🏷 **Review: LGTM — recommend merge.** CI is green (test ubuntu 22/24, windows 22/24, build-artifact, version-guard all pass). I independently re-ran the pre-flight on the PR branch and reproduced the bug both ways.

## What I verified

1. **Root cause is real — reproduced both directions.** On `pr-664` I temporarily reverted the default to `redirect: "follow"`: both new tests fail with exactly the reported symptom — `client must see the upstream 302 passed through; got 405: {"error":"method not allowed"}`. Restored `"manual"`: both pass. The regression tests genuinely pin the bug.
2. **Pre-flight re-run on the branch:** `npm run typecheck` ✅, `npm test` ✅ 1248/1248 (matches your claim), `npm run build` ✅.
3. **Call-site audit — every `fetchWithTimeout` consumer accounted for:**
   - `src/registry.ts:109` (models.dev) and `src/server.ts:873` (`/__bili/upstream/test` probe) — explicitly opt into `"follow"`, behavior unchanged ✅
   - `src/server.ts:3318` (main forward) — inherits `"manual"`; a 3xx lands in the `!upstream.ok` branch (src/server.ts:3449) and is passed through verbatim; `location` is **not** in `UPSTREAM_HOP_HEADERS` (src/server.ts:208-227), so the client gets status + Location ✅
   - `src/server.ts:3390` (compat role-rewrite retry), `src/server.ts:3973` (fake-completion retry) — 3xx handled gracefully ("retry rejected (HTTP 302); presenting original" / original 400 continues) ✅
   - `fetchWithRetry` consumers (`src/preflight.ts:330`, `src/loop/core.ts:213`, `src/compress-loop-responses.ts:218`) — a 3xx now surfaces as `UpstreamHttpError(3xx)` instead of a silently-followed POST→GET: preflight reports a clean failure message; the compress loops emit a protocol error event. Strictly better than before, and a raw 302 couldn't pass through mid-stream anyway (headers already committed) ✅
4. **No coverage gap:** I grepped for direct `fetch(` calls bypassing `fetchWithTimeout` — the rest are local/management-plane only (`__bili/plugin/*`, web UI, npm auto-update, launcher probes). `fetchWithTimeout` is the complete choke point for proxied client traffic.
5. **Both compression modes:** the change is in the shared transport layer, mode-agnostic; plugin-mode re-sends and proxy-mode loop re-sends both inherit `"manual"` and handle 3xx cleanly. ACP impact: none — the new tests assert ACP render tags are present in the forwarded body.

## Layer judgment

The 405 is a symptom; the root cause is the proxy following redirects on the client's behalf (undici's default `redirect: "follow"` downgrades POST→GET and drops the body on 301/302/303). The fix targets the root cause at the single transport choke point with correct forward-proxy semantics (pass 3xx through, client follows with its own policy) — not a symptom patch. The two explicit `"follow"` opt-ins keep internal bili fetches behavior-identical.

- The main forward path now logs a legitimate 3xx as `warn` via the existing non-2xx logging path (src/server.ts:3564, the #174 "always log a non-2xx" policy). Slightly noisier for redirecting upstreams, but consistent with existing behavior — fine as-is.
- I independently verified your #619 side finding: on the unsupported-encoding path (src/server.ts:1001-1007) the still-encoded body is forwarded while `buildForwardTarget` (src/server.ts:2803-2807) strips `content-encoding` via `UPSTREAM_HOP_HEADERS` (src/server.ts:217) — the upstream gets an encoded body with no encoding declaration (400-class), and the "content-encoding header stays intact" comment at src/server.ts:1003-1004 is stale. Correctly deferred; it's documented in the PR description, which satisfies the report-only bar for a minor issue. Happy to file a tracking issue for it if you want one.

Merge is human-only per repo rules — the PR is ready when you are.

PR #665, opened 2026-09-09
fix: Codex preflight summary compatibility (max_output_tokens + responses-lite header)

### LINKED ISSUE #663: fix: Codex preflight summaries fail with unsupported max_output_tokens
## Problem

With billion-context 0.1.95 proxying Codex through `https://chatgpt.com/backend-api/codex/responses`, an over-window request enters preflight compression, but the summary request is rejected with:

```json
{"detail":"Unsupported parameter: max_output_tokens"}
```

Bili then returns HTTP 502 (`preflight compression could not bring it under ... HTTP 400`). Repeated client retries cannot recover the session.

## Reproduction

1. Use the Codex ChatGPT Responses backend with valid authentication (credentials omitted).
2. Trigger proxy-side preflight compression for an over-window conversation.
3. `src/preflight.ts` builds a Responses summary payload with `max_output_tokens: MAX_SUMMARY_OUTPUT_TOKENS`.
4. The backend rejects that parameter with HTTP 400. A minimal authenticated request with this field reproduces the rejection independently of history size.

## Observations

- Ordinary short streaming model requests succeed through the same outbound proxy.
- Existing Bili compression had successfully reduced an earlier turn; context management was active.
- The previous upstream usage was approximately 92k tokens, while the next processed payload was locally estimated at approximately 437k against a configured 272k budget. These are different measurements, not proof of a tokenizer defect.
- The immediate blocker is the unsupported summary parameter. The source of the payload growth is not yet established.
- Existing summary fallback handles `stream must be true`, but not this parameter rejection.

## Expected behavior

Generate a compatible summary request for the Codex ChatGPT backend (or narrowly learn/retry the explicitly rejected optional parameter), while retaining output limits for standard Responses providers and preserving the preflight overflow guard. Cover both summary generation and successful recovery/forwarding in regression tests.

No credentials, session contents, or private host configuration are included. We are preparing a minimal local fix to restore availability.

ranxianglei:
Fixes #663

Two related fixes so Codex (ChatGPT-login) preflight summaries work end-to-end. Both live in `requestSummary` (src/preflight.ts) and only affect the independently-built summary call — the original model request is untouched.

## 1. Learn to drop `max_output_tokens` when the upstream rejects it
The ChatGPT codex backend rejects the Responses `max_output_tokens` parameter outright (400 `{"detail":"Unsupported parameter: max_output_tokens"}`), which killed preflight summaries with a 502 after the #626 stream fix. The proxy now detects that rejection, retries the summarization without the optional parameter, and remembers it per session+upstream+model — so a model/provider that accepts the limit keeps the existing 8192 output cap. Both rejection orders (stream-first, param-first) recover, and the retries are bounded (each capability is learned at most once).

## 2. Drop the Codex Lite header from the summary call only
`requestSummary` copied ALL main-request headers into its independently-built summary body. The main Codex request carries `x-openai-internal-codex-responses-lite` (its body has `reasoning.context: all_turns`), but the summary body has no reasoning field — so the backend rejected the summary with 400 "…requires `reasoning.context` to be `all_turns`". The fix strips that one request-shape-specific header from the summary call only; the original model request keeps the header and its reasoning fields. Auth and routing headers are preserved. The strip is keyed on the Codex header name, so it is a no-op for Anthropic/OpenAI.

## Tests
`tests/preflight-max-output-tokens.test.ts` — 6 e2e tests:
- order A (stream-first) + order B (param-first): both rejection orders learn and recover; second request first-shot compatible
- standard provider: 8192 cap retained, no capability learned
- model scoping: rejection learned for model A keeps model B's cap
- codex lite header: summary drops the Lite header, main request keeps it; auth preserved
- anthropic: header strip is a no-op — summary keeps auth, fold + forward OK

## Pre-flight
- `npm run typecheck` ✓
- `npm test` ✓ (1252/1252)
- `npm run build` ✓

Validated against the real backend by the reporter: an over-window Lite-mode request now completes after preflight folds the ranges, preserving the canary.

### DIFF
--- a/src/preflight.ts
+++ b/src/preflight.ts
@@ -267,6 +267,25 @@ function rememberNoMaxOutputTokens(deps: PreflightDeps): void {
     deps.session.metadata.preflightNoMaxOutputTokens = map;
 }
 
+// #663: request-shape-specific headers that must NOT ride the independently
+// constructed summary call. The main Codex request carries
+// x-openai-internal-codex-responses-lite, which the backend only accepts when
+// the body has reasoning.context: all_turns. The summary body is built
+// independently (no reasoning field), so carrying the header over makes the
+// backend reject it (400 "…requires `reasoning.context` to be `all_turns`").
+// The original model request keeps the header and its reasoning fields; only
+// the side summary call drops it. Auth/routing headers are preserved.
+const SUMMARY_STRIP_HEADERS = new Set(["x-openai-internal-codex-responses-lite"]);
+
+function summaryHeaders(deps: PreflightDeps): Record<string, string> {
+    const headers: Record<string, string> = {};
+    for (const [k, v] of Object.entries(deps.headers)) {
+        if (SUMMARY_STRIP_HEADERS.has(k.toLowerCase())) continue;
+        headers[k] = v;
+    }
+    return headers;
+}
+
 // Extract the summary text from a buffered SSE body (the streaming twin of
 // extractSummaryText). For Responses, prefer the response.completed event's
 // full response object (reuses the JSON extractor); otherwise accumulate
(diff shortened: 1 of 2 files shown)


ranxianglei:

[bot] 🏷 Review in progress — verifying #663 repro evidence, checking the diff against the #626 capability-learning pattern, and running typecheck/tests/build on the PR branch. Will report findings here.

PR #680, opened 2026-09-10
fix: stop a space in the profile from truncating the launched client (#679)

### LINKED ISSUE #679: [bug] Windows: a space in the client or -e path truncates the launch (spawn shell:true, unquoted) — and silently kills compression when only the command is quoted
## Summary

On Windows, every launcher is affected when the resolved client binary path — or the `-e <extension>` path the launcher prepends — contains a space. `runClient` spawns with `shell: true` and passes the command unquoted, so cmd.exe re-splits the line on whitespace.

Wild-caught with the omp launcher (profile path contains a space, which is also where `omp.exe` and the npm-global `billion-context` package live):

```
bili: started proxy at http://127.0.0.1:5165 (MITM domains: ...) (HTTP /bili/ rewrites: 10)
'C:\Users\John' is not recognized as an internal or external command,
operable program or batch file.
```

## Impact

Both halves of the spawn are broken, and only the first one is loud:

1. **The command is truncated** → the launch fails outright (the visible symptom above).
2. **The args are truncated too.** If the command alone is quoted, the process starts but `-e` receives a path cut at the first space (`C:\Users\John`), the extension never loads, and compression silently never activates. This is the failure class behind the CHANGELOG note *"a long omp session ended up at 92% context with zero compressions"* — the user sees no error at all.
3. `shell: true` also makes Node emit `DEP0190` (`Passing args to a child process with shell option true can lead to security vulnerabilities…`) on every Windows launch.

This is not omp-specific: the same `runClient` is shared by the pi / codex / claude / omp / opencode / hermes launches, and each of their resolved commands can carry a space — an npm-installed CLI resolves to a `.cmd` shim, omp resolves to a real `.exe`, and pi falls back to `process.execPath` under `C:\Program Files\nodejs\`.

## Repro

Minimal, without bili:

```js
import { spawnSync } from "node:child_process";
const exe = "C:\\Users\\John Doe\\AppData\\Local\\omp\\omp.exe";
spawnSync(exe, ["-e", "C:\\Users\\John Doe\\ext.js"], { shell: true });
// status=1
// stderr: 'C:\Users\John' is not recognized as an internal or external command
```

Quoting only the command fixes the crash but leaves the silent half:

```js
spawnSync(`"${exe}"`, ["-e", "C:\\Users\\John Doe\\ext.js"], { shell: true });
// status=0, but the child sees argv: ["-e", "C:\\Users\\John"]  ← extension path cut
```

Either shape of the path triggers it — a spaced `.exe`, a spaced `.cmd` shim, or a script/extension path with a space.

## Root cause

`src/launcher.ts`, `runClient`:

```ts
const child = spawnImpl(cmd, args, { stdio: "inherit", env, shell: process.platform === "win32" });
```

With `shell: true`, Node joins `[cmd, ...args]` with spaces and hands cmd.exe `/d /s /c "<that line>"`. cmd.exe strips the outer pair and re-parses the remainder, splitting on whitespace — so a spaced path is cut at its first space. Node cannot fix this for us: the same option also tells Node not to quote the args (hence `DEP0190`).

## Suggested fix

Two changes in `runClient`, both small:

1. **Only use a shell where one is actually required.** Only `.cmd`/`.bat` shims cannot be spawned directly on Windows; a real `.exe`, node, or an extensionless binary spawns fine with no shell, and then the OS quotes the executable path *and* argv itself, spaces included. This removes the shell layer entirely for `bili omp`, which is the reported case. Verified locally: `spawnSync("C:\\Users\\John Doe\\cmd copy.exe", ["/c","echo","hello from spaced arg"], { shell: false })` → `status=0`, argv intact, while the same call with `shell: true` fails with `'C:/Users/John' is not recognized`.
2. **For the `.cmd`/`.bat` case, quote the line and pass it verbatim**:
   `spawn(comspec, ["/d","/s","/c", line], { windowsVerbatimArguments: true })`
   where `line = '"' + [cmd, ...args].map(quoteWinToken).join(" ") + '"'` and `quoteWinToken` only quotes a token that contains whitespace or a quote. The extra outer pair is the documented cmd.exe `/s` form — cmd strips it and parses the inner tokens with their own quoting intact (the same trick `cross-spawn` uses). Because tokens stay bare when they don't need quoting, a space-free launch produces a byte-identical line to today's.

Neither change touches POSIX behaviour: the batch branch is gated on `process.platform === "win32"`, and no shell metacharacters are used anywhere in the launcher argv.

I have a patch with regression tests ready — a pure `buildWindowsCommandLine` test that runs on every platform, plus win32-only real launches that spawn a `.cmd` shim / spaced arg paths in a temp directory whose name contains a space and assert argv survives. Happy to open the PR, or to reshape it if you'd prefer a different approach (e.g. always routing through the explicit comspec form instead of splitting on extension).

## Environment

- Windows 10.0.26200, Git Bash (MSYS2)
- billion-context 0.1.100, omp 18.1.16
- Currently worked around by pointing `BILI_CLIENT_BIN` at a space-free path, which is why the crash is not seen on every launch — the silent arg-truncation half is what actually bites.

CaelumSea:
Fixes #679

## What

On Windows, `runClient` spawned with `shell: true` and an unquoted command, so cmd.exe re-split the line on whitespace. Any profile containing a space — which is where `omp.exe` and the npm-global `billion-context` package live — got cut at the first space:

```
bili: started proxy at http://127.0.0.1:5165 (MITM domains: ...) (HTTP /bili/ rewrites: 10)
'C:\Users\John' is not recognized as an internal or external command,
operable program or batch file.
```

Quoting only the command repairs the crash and leaves the worse half: the **args** are still truncated, so `-e` points at a path cut at the first space, the extension never loads, and compression silently never activates — the "92% context with zero compressions" class, with no error shown. #679 has the minimal repro for both halves.

## The fix

Two changes in `runClient`, both in `src/launcher.ts`:

1. **A shell is now used only where one is required** — a `.cmd`/`.bat` shim (npm-installed CLI). A real `.exe`, node, or an extensionless binary spawns directly, where the OS quotes the executable path *and* argv itself, spaces included. This removes the shell layer entirely for `bili omp`, the reported case.
2. **The `.cmd`/`.bat` case gets an explicitly quoted line, passed verbatim** (`cmd.exe /d /s /c`, `windowsVerbatimArguments: true`), built by the new `buildWindowsCommandLine`. A token is quoted only if it contains whitespace or a quote, so a space-free launch produces a byte-identical line to today's; the extra outer pair is the documented cmd.exe `/s` form that cmd strips before parsing the inner tokens with their own quoting intact.

Verified before writing the code: `spawnSync("C:\Users\John Doe\cmd copy.exe", ["/c","echo","hello from spaced arg"], { shell: false })` → `status=0` with argv intact, while the same call with `shell: true` fails with `'C:/Users/John' is not recognized`.

## Tests

7 new cases in `tests/launcher.test.ts`:

- `buildWindowsCommandLine`: spaced tokens quoted plus the `/s` outer wrap; a space-free line stays byte-identical to the old form; inner quotes doubled rather than dropped. Runs on every platform, so the ubuntu job covers the quoting logic too.
- `runClient`: a `.cmd` shim is routed through cmd.exe with the quoted line and `windowsVerbatimArguments`; a real `.exe` is spawned directly with no shell.
- win32-only real launches: a temp directory whose name contains a space, asserting argv survives both through spaced argument paths (direct spawn — the omp case) and through a real `.cmd` shim (the cmd.exe path).

`npm run typecheck`, `npm run build`, and `tests/launcher.test.ts` (103 → 110 cases) pass on Windows. An earlier version of this patch routed *every* Windows launch through the explicit comspec form and hung the existing `runLaunch` fakes (they identify the client by the `cmd` argument, which then arrived as `cmd.exe`) — the extension-gated design avoids that and keeps every existing Windows test untouched.

Full `npm test` on my machine: 1282 pass / 3 fail, and the identical 3 fail on unmodified `master` (they read the local system-proxy and bili routes config), so they are not from this change.

## Compatibility

The batch branch is gated on `process.platform === "win32"` and only fires for `.cmd`/`.bat`; every other spawn is unchanged on POSIX and, on Windows, loses a `cmd.exe` layer it never needed — along with the per-launch `DEP0190` warning the shell path produced. No shell metacharacters are used anywhere in the launcher argv.


CaelumSea:
The one red check is not from this patch: `build-artifact` builds and uploads the tarball fine, then fails at the final "Comment on PR with install instructions" step with `403 Resource not accessible by integration` — a fork PR's `GITHUB_TOKEN` is read-only, so it cannot post the comment (the same run already correctly skips the npm PR-tag publish for forks). The four `test` jobs (ubuntu + windows, node 22 and 24) and `version-guard` all pass.

ranxianglei:
ework接单

ranxianglei:

## Review — approved, mergeable ✅

**Diff cleanliness (first gate): pass.** Single focused commit vs merge-base `275ce5b`, 3 files (`CHANGELOG.md` +1, `src/launcher.ts` +44/-2, `tests/launcher.test.ts` +113), no stray/generated changes, CHANGELOG entry present under *Fixes* and referencing #679. Every changed line serves the stated purpose.

**Bug is real, root cause correct.** On Windows `runClient` spawned with `shell: true` and an *unquoted* command+args string. Node hands that to `comspec /c <line>` and cmd.exe re-tokenizes on whitespace, so any spaced profile path (`C:\Users\John Doe\…` — exactly where `omp.exe` and the npm-global install live) splits at the first space → `'C:\Users\John' is not recognized`. And the second half is the nastier one: even quoting only the command leaves the **args** truncated, so `-e <ext>` points at a path cut at the first space, the extension never loads, compression silently never activates — no error anywhere. You correctly identified both layers rather than just patching the crash. I verified the mechanism against Node's `shell:true` Windows behavior and confirmed it end-to-end on Linux-CI (see below); the two win32 real-launch tests are gated and I can't execute them off-Windows, but your Windows run + the classic error signature are conclusive.

**Fix is sound.** Routing decision is exactly right:
- `.cmd`/`.bat` shims (the only thing that genuinely needs a shell) → `cmd.exe /d /s /c <quoted-line>` with `windowsVerbatimArguments: true`, via `buildWindowsCommandLine`. Correct `/s` outer-quote-pair handling, tokens quoted only when they contain whitespace/quote.
- Everything else (real `.exe`, node, extensionless binary) → direct `spawn`, no shell — OS quotes exe path *and* argv, spaces included.
- Bonus you may not have called out: pi's node fallback (`command = process.execPath`, e.g. `C:\Program Files\nodejs\node.exe`) was *also* broken before by the same space-splitting of "Program Files"; direct-spawn fixes that too.
- Space-free launches stay byte-identical (tokens stay bare), so no behavior drift for the common case, and the per-launch `DEP0190` warning disappears.

**Tests:** good coverage. Cross-platform `buildWindowsCommandLine` unit tests (spaced tokens quoted + `/s` wrap; space-free byte-identical; inner-quote doubling) plus a `runClient` fake-spawn test asserting the `.exe`→direct/no-shell path and the `.cmd`→cmd.exe verbatim path. Two real-launch tests in a temp dir containing a space cover both the direct-spawn and the real `.cmd` shim paths (win32-gated). I ran them here: `typecheck` clean, `tsup build` OK, `tests/launcher.test.ts` → **0 fail** (unit + cross-platform pass; the 3 win32-only real-launch cases skip on Linux as designed).

**CI red check — your diagnosis is correct.** The only failing job is `build-artifact` dying at the final "Comment on PR" step with `403 Resource not accessible by integration`: a fork PR's `GITHUB_TOKEN` is read-only, so it can't post the artifact comment. All four test jobs (ubuntu+windows × node22+24) and the version-guard pass. Not caused by this patch — it will go green on merge into `master` (same-repo token). Nothing to change here.

---

**Two minor, non-blocking observations** (I did not force-edit your fork branch for these — flagging per review norms so you/the owner decide):

1. **Narrow PATHEXT regression window, no shipped client affected.** `resolveClientCommand` returns a fully-resolved path when `resolveOnPath` finds the binary (it checks `.cmd/.bat/.exe/""`), otherwise it falls back to the **bare name**. A bare name now goes through direct `spawn`, and `CreateProcessW` does *not* apply PATHEXT — so a client reachable *only* via a PATHEXT extension outside `{.cmd,.bat,.exe}` (e.g. a hand-installed `.com`/`.vbs`/`.js` wrapper, or a custom-PATHEXT setup) would fail where the old `shell:true`+cmd.exe succeeded. In practice every supported client (pi/codex/claude/omp/opencode/hermes/dsh/zcode/qoder/trae/codebuddy) is an npm package → `.cmd` shim → resolved by `resolveOnPath` → batch path, so **none are hit**. Optional hardening if you want belt-and-suspenders: route an unresolved bare win32 name through `cmd.exe` too (strictly preserves old behavior; costs a couple of lines + one assertion tweak). Left as-is is defensible given zero real-world exposure.

2. **`quoteWinToken` inner-quote doubling is dead code for real inputs.** NTFS forbids `"` in filenames and the launcher argv carries no shell metacharacters (per your own note), so the `"`-doubling branch never fires for a real path — harmless, and the corresponding unit test documents intent for an impossible case. No action needed; just noting it isn't load-bearing.

Net: correct root cause, correct minimal fix, clean diff, solid tests, red check explained. Ready to merge.

修复了 Windows 下用户目录含空格导致启动命令被 cmd.exe 按空格截断、进而 `-e` 扩展路径失效使压缩静默不生效的问题（根因是 `shell:true` + 未加引号的命令行被 shell 二次切分）：改为仅对确实需要 shell 的 `.cmd/.bat` shim 走显式加引号 + `windowsVerbatimArguments` 的 `cmd.exe`，真实 `.exe`/node/无扩展名二进制直接 spawn，空格路径与参数均完整保留；改动最小、测试充分、红勾系 fork PR 只读 token 无法发评论所致而非本补丁，可以合并。

CaelumSea:
Superseded by #682 — it lands the same fix on current master and additionally routes extensionless paths and unresolved bare names through comspec, covering the PATHEXT window raised in the review. The #679 diagnosis is carried there with repro and tests, so this fork branch has nothing left to contribute. Closing to keep the queue clean.

Issue #681, opened 2026-09-10
[bug] Windows: codex MCP injection -c value loses its quotes through cmd.exe when node lives under a spaced path


ranxianglei:
来源: ranxianglei/billion-context#679 分析 Windows 启动截断问题时发现

## Context

While fixing #679 (runClient's shell:true truncated spaced client/-e paths), I audited every arg class the launcher passes to clients on Windows. One class cannot be transmitted through cmd.exe faithfully by ANY quoting scheme, and it is broken today already.

## What happens

Default `bili codex` injects the native bili MCP tools inline (src/launcher.ts, buildCodexMcpArgs):

```
-c mcp_servers.bili.command=<JSON.stringify(process.execPath)>
```

When node lives under a spaced path (e.g. `C:\Program Files\nodejs\node.exe` — the stock Windows install location), this single argv token contains double quotes, backslashes AND spaces:

```
mcp_servers.bili.command="C:\\Program Files\\nodejs\\node.exe"
```

On Windows codex resolves to an npm `.cmd` shim, so the spawn goes through cmd.exe (before AND after the #679 fix — shims always need cmd). cmd.exe treats `"` purely as a quote-state toggle and has NO escape mechanism for a literal quote inside an argument. However the line is built, cmd strips the token's surrounding quotes before the shim sees it, so codex receives:

```
mcp_servers.bili.command=C:\Program Files\nodejs\node.exe
```

— an UNQUOTED TOML value containing spaces. That is not a valid TOML string.

## Impact (needs Windows verification)

Unverified on real Windows which way codex fails: it either rejects the `-c` assignment (launch error) or drops/mis-parses the `mcp_servers.bili` entry (bili MCP server never registers → codex runs WITHOUT the compress tool → in plugin mode compression silently never activates — the same silent failure class as #679's second half). Either way: Windows users with node under a spaced path + default MCP mode are affected.

Note: this is independent of #679's fix — even with perfect per-token quoting, a token containing both quotes and whitespace cannot survive cmd.exe; no quoting scheme exists for it.

## Suggested direction

Stop encoding absolute paths into inline `-c` values. Deliver the MCP server definition via a FILE instead (a temp config file if codex exposes an alternate-config/profile flag, or a temp profile/home dir) — analogous to claude's existing temp `--mcp-config` file approach (src/launcher.ts, injectMcp branch).

First step: verify actual codex behaviour on Windows (does it reject the `-c`, or start tool-less?) before choosing the fix.

[...]

PR #682, opened 2026-09-10
fix: stop truncating Windows launches at spaced paths (#679)

### LINKED ISSUE #679: [bug] Windows: a space in the client or -e path truncates the launch (spawn shell:true, unquoted) — and silently kills compression when only the command is quoted

ranxianglei:
## What

On Windows every launcher (`bili pi/codex/claude/omp/opencode/hermes/dsh`) spawned the client with `shell: true` and unquoted args, so cmd.exe re-split the line on whitespace:

- space in the **resolved client** path → launch fails outright (`'C:\Users\John' is not recognized...`)
- space in the **`-e` extension path** → process starts but the arg is cut at its first space; the extension never loads and compression silently never activates (the failure class behind the CHANGELOG's "92% context, zero compressions" omp session)
- plus a `DEP0190` deprecation warning on every Windows launch

## Fix (src/launcher.ts — runClient)

- `.exe`, node, and any existing path with an extension now spawn **directly** (no shell): CreateProcess/libuv quote the executable and argv itself, spaces included. Covers the reported omp case and the `pi` fallback under `C:\Program Files\nodejs\`.
- `.cmd`/`.bat` shims (npm-installed CLIs) and unresolved bare names route through an explicit `comspec /d /s /c "<line>"` with `windowsVerbatimArguments: true`: each token quoted only when it contains whitespace, wrapped in one extra outer quote pair that cmd's `/s` strips before parsing — the documented `/s` form, same trick cross-spawn uses. A space-free launch produces a byte-identical line to today's.
- `shell: true` is gone entirely → no more DEP0190.
- POSIX behaviour untouched (branch gated on `platform === "win32"`); the proxy child spawn (`ensureProxyRunning`) was already shell-less.

Pure helpers exported for testability: `quoteWinToken`, `buildWindowsCommandLine`, `planClientSpawn(cmd, args, env, platform?)`.

## Tests (tests/fix-679-windows-launch.test.ts)

- Platform-independent: token quoting rules, command-line construction, spawn planning (.cmd/.BAT/.exe/bare-name/extensionless × posix/win32 × COMSPEC set/unset), runClient plumbing (comspec argv shape, verbatim flag, no shell option, exit-code/signal mapping unchanged).
- Real-launch regressions gated to win32 (skipped elsewhere): a `.cmd` shim in a temp dir whose name contains spaces must receive its spaced `-e` arg intact; a spaced-argv direct spawn through `process.execPath` must round-trip including an arg with embedded quotes.

## Known limitation (pre-existing, out of scope)

cmd.exe has no escape mechanism for embedded double quotes, so tokens containing BOTH quotes and whitespace cannot be transmitted through a `.cmd` shim by any quoting scheme. Such tokens keep today's exact behaviour. This bites one real case — codex MCP injection's inline `-c mcp_servers.bili.command=<JSON>` value when node lives under a spaced dir (broken before this change too): https://github.com/ranxianglei/billion-context/issues/681

## Pre-flight

- `npm run typecheck` — clean
- `npm test` — 1299 tests, 0 fail, 2 skipped (the win32-gated real launches)
- `npm run build` — ok

Fixes #679

### DIFF
--- a/src/launcher.ts
+++ b/src/launcher.ts
@@ -103,7 +103,7 @@ export interface SpawnChild {
 export type SpawnFn = (
     command: string,
     args: readonly string[],
-    options: { detached?: boolean; stdio?: StdioOptions; env?: NodeJS.ProcessEnv; shell?: boolean },
+    options: { detached?: boolean; stdio?: StdioOptions; env?: NodeJS.ProcessEnv; shell?: boolean; windowsVerbatimArguments?: boolean },
 ) => SpawnChild;
 
 export interface LaunchOptions {
@@ -1597,15 +1597,62 @@ export function stopProxy(handle: ProxyHandle): void {
     } catch {}
 }
 
+/** #679: quote one token for cmd.exe's line parser. Only whitespace-bearing
+ *  tokens get wrapped in double quotes, so a space-free launch produces a
+ *  byte-identical line to the old shell:true form. A token containing an
+ *  embedded double quote stays bare: cmd.exe has no escape mechanism for
+ *  quotes, so wrapping would only change how it is mangled (today's behavior
+ *  preserved). */
+export function quoteWinToken(token: string): string {
+    if (!/\s/.test(token) || token.includes('"')) return token;
+    return `"${token}"`;
+}
+
+/** #679: full command line for `comspec /d /s /c <line>` — tokens quoted as
+ *  needed, wrapped in one extra outer pair that cmd's /s strips before
+ *  parsing the inner tokens with their own quoting intact (the documented /s
+ *  form; same trick cross-spawn uses). */
+export function buildWindowsCommandLine(cmd: string, args: readonly string[]): string {
+    return `"${[cmd, ...args].map(quoteWinToken).join(" ")}"`;
+}
+
+/** #679: which spawn form a resolved client needs on Windows. Only .cmd/.bat
+ *  shims and unresolved bare names need cmd.exe — CreateProcess cannot
+ *  execute a batch file, and an extensionless name needs cmd's PATHEXT
+ *  resolution. Everything else (.exe, node, an existing path with an
+ *  extension) spawns directly and the OS quotes the executable and argv
+ *  itself, spaces included. shell:true is never used anymore: no DEP0190, no
+ *  cmd.exe re-splitting of spaced paths at their first space (which truncated
+ *  both the command and its args). */
+export function planClientSpawn(
+    cmd: string,
+    args: readonly string[],
+    env: NodeJS.ProcessEnv,
+    platform: NodeJS.Platform = process.platform,
+): { command: string; args: string[]; windowsVerbatimArguments?: boolean } {
+    if (platform !== "win32") return { command: cmd, args: [...args] };
+    const lower = cmd.toLowerCase();
+    const base = cmd.slice(Math.max(cmd.lastIndexOf("/"), cmd.lastIndexOf("\\")) + 1);
+    const needsCmd = lower.endsWith(".cmd") || lower.endsWith(".bat") || !path.extname(base);
+    if (!needsCmd) return { command: cmd, args: [...args] };
+    const comspec = nonEmpty(env.COMSPEC) ? env.COMSPEC : "cmd.exe";
+    return {
+        command: comspec,
+        args: ["/d", "/s", "/c", buildWindowsCommandLine(cmd, args)],
+        windowsVerbatimArguments: true,
+    };
+}
+
 export function runClient(
     cmd: string,
     args: string[],
     env: NodeJS.ProcessEnv,
-    deps?: { spawnImpl?: SpawnFn },
+    deps?: { spawnImpl?: SpawnFn; platform?: NodeJS.Platform },
 ): Promise<number> {
     const spawnImpl = deps?.spawnImpl ?? (spawn as SpawnFn);
+    const plan = planClientSpawn(cmd, args, env, deps?.platform);
     return new Promise((resolve, reject) => {
-        const child = spawnImpl(cmd, args, { stdio: "inherit", env, shell: process.platform === "win32" });
+        const child = spawnImpl(plan.command, plan.args, { stdio: "inherit", env, windowsVerbatimArguments: plan.windowsVerbatimArguments });
         child.on?.("error", (...rest: unknown[]) => reject(rest[0]));
         child.on?.("exit", (...rest: unknown[]) => {
             const code = rest[0];
--- a/src/launcher.ts
+++ b/src/launcher.ts
@@ -1649,6 +1649,9 @@ export function runClient(
     env: NodeJS.ProcessEnv,
     deps?: { spawnImpl?: SpawnFn; platform?: NodeJS.Platform },
 ): Promise<number> {
+    // #679: never shell:true — besides DEP0190, cmd.exe re-splits the unquoted
+    // line on whitespace and truncated spaced client/-e paths at their first
+    // space; planClientSpawn picks the direct-vs-comspec form instead.
     const spawnImpl = deps?.spawnImpl ?? (spawn as SpawnFn);
     const plan = planClientSpawn(cmd, args, env, deps?.platform);
     return new Promise((resolve, reject) => {
--- a/tests/fix-679-windows-launch.test.ts
+++ b/tests/fix-679-windows-launch.test.ts
@@ -60,6 +60,10 @@ test("planClientSpawn win32: .exe spawns directly even when spaced — the OS qu
         planClientSpawn("C:\\Program Files\\nodejs\\node.exe", ["C:\\Users\\John Doe\\cli.js"], WIN_ENV, "win32"),
         { command: "C:\\Program Files\\nodejs\\node.exe", args: ["C:\\Users\\John Doe\\cli.js"] },
     );
+    assert.deepEqual(
+        planClientSpawn("C:\\TOOLS\\NODE.EXE", [], WIN_ENV, "win32"),
+        { command: "C:\\TOOLS\\NODE.EXE", args: [] },
+    );
 });
 
 test("planClientSpawn win32: .cmd/.bat shims route through comspec /d /s /c with verbatim quoting", () => {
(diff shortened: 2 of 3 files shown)


ranxianglei:

[bot] 🏷 收到，开始 review PR #682（Windows launcher 路径截断修复）。流程：先查重 → 审 diff 整洁度 → 本地跑 typecheck/test/build → 代码走查（含两种压缩模式影响面）→ 发现问题直接在 PR 分支补 commit。结论稍后更新。

ranxianglei:

[bot] 🏷 **Review complete — LGTM, ready to merge.** I verified everything locally rather than trusting the description.

**Duplicate screening** — searched open+closed issues/PRs for the Windows-launch/spaced-path/DEP0190 topic: no duplicates. This PR is the designated fix for #679 (`Fixes #679`); the related codex-MCP `-c` JSON defect is the separate, already-filed #681 (different failure class — quotes+spaces, unfixable by any cmd quoting scheme), not a dup.

**Diff cleanliness (first gate)** — clean. Exactly 2 files, every line on-topic: `src/launcher.ts` (+50/−3, all inside the new helpers + the one `runClient` spawn line) and `tests/fix-679-windows-launch.test.ts` (+207). No lockfile churn, no generated artifacts, no whitespace/reformat noise. Branch was current with master before my commit (single commit on tip); after my push it's 2 commits, still mergeable.

**Pre-flight (re-run on the branch, Linux CI box)**
- `npm run typecheck` — clean
- `npm test` — 1299 tests: 1297 pass / 0 fail / 2 skipped (the win32-gated real launches, as designed)
- `npm run build` — ok; `dist/index.js` self-contained (2.7 MB, acp-kernel inlined)
- Hygiene: `version` untouched (0.1.100), acp-kernel pinned exact `0.0.61`, no `as any`/`@ts-ignore`/`console.error` introduced.

**Code walkthrough (what I checked beyond the happy path)**
- *Classification completeness.* `resolveClientCommand` can only produce `.exe` / `.cmd` / `.bat` / extensionless / bare-name values through built-in paths, so the `needsCmd` split covers every built-in shape. The pi fallback (`process.execPath` + `cli.js` prefixArgs) and the spaced `-e dist/agent/*.js` extension args now ride direct CreateProcess spawn — exactly the two shapes from #679's repro. Uppercase `.CMD`/`.BAT` handled via `lower`; dotfile bases (`.hidden`) correctly fall to the comspec route.
- *Verbatim mechanics.* With `windowsVerbatimArguments: true` libuv appends argv unmodified (program name still quoted if needed), so the command line is literally `comspec /d /s /c "<line>"`; `/s` strips exactly the outer pair and the inner per-token quotes survive to the shim's `%*` — the standard npm-shim flow, same trick cross-spawn ships. Direct-spawn branch passes `windowsVerbatimArguments: undefined` → MSVCRT argv quoting, correct for real exes.
- *COMSPEC.* Node's Windows env access is case-insensitive, so `env.COMSPEC` resolves the system's `ComSpec`; the `"cmd.exe"` fallback (empty/unset) resolves via PATH since System32 is always on it. Worst case of a custom ComSpec being missed = system cmd, still correct.
- *Compat.* `SpawnFn` type extended additively; both existing `runClient` call sites (`runLaunch`, `runTestPi`) unchanged — `platform` defaults to `process.platform`. Exit-code/signal mapping (`signal ? 130 : code ?? 0`) untouched and re-tested. Grep confirms zero remaining `shell:` spawn sites in `src/` — DEP0190 is structurally gone, not just suppressed.
- *Two compression modes (§6).* The change sits entirely in client-process spawn, before any wire traffic — orthogonal to plugin/proxy-mode logic; neither mode's carrier/injection behavior is touched. Both modes benefit identically (the client actually launches).
- *Tests.* Pure-helper matrix (quoting rules, line construction, spawn planning across posix/win32 × COMSPEC set/unset × shim classes), `runClient` plumbing (verbatim flag, no `shell` option, exit/signal mapping), plus win32-gated end-to-end launches (spaced-dir `.cmd` shim receiving a spaced `-e` intact; spaced argv round-trip through `process.execPath` including embedded quotes). Solid.

**Review fix pushed** — `5cc35b4` adds three coverage gaps I found while walking the branches: uppercase `.EXE` direct-spawn (pins the `lower` normalization), spaced-COMSPEC verbatim-as-program, and dotfile base (`.hidden`) → comspec route. Re-ran the suite green after the push.

**Minor observations (report-only, pre-existing, not blocking)**
1. *Script-extension overrides.* A user pointing `BILI_CLIENT_BIN`/`PI_BIN` at a bare `.js` script previously "worked" under `shell:true` only when Windows had the `.js`→node file association registered; now it fails fast with a clear error + exit 1 instead of depending on registry state. Built-in resolution never produces such commands, so this is a niche override case — worth a CONFIGURATION note ("expect an executable") or a follow-up routing script extensions through comspec, only if someone hits it.
2. *Non-whitespace cmd metacharacters.* Tokens containing `&`/`%`/`!`/`<`/`>`/`|`/`^`/parens **without** any space stay bare (unchanged from today — not a regression; quoting wouldn't stop `%VAR%` expansion anyway). cross-spawn quotes on the full metachar set, which would be a reasonable hardening follow-up; deliberately not bundled here to keep the "space-free launch stays byte-identical" guarantee scoped to the bug class in #679. The one real-world victim of the broader quotes+metachars gap (#681) is already tracked.

**Verdict:** diff clean, fix correct and minimal, tests meaningful, pre-flight green on the branch. Merge it.

中文摘要：修复了 Windows 下 launcher 因 `shell:true` + 无引号参数导致路径含空格被 cmd.exe 截断的问题（.exe/node/带扩展名直接 spawn，.cmd/.bat/裸名走显式 `comspec /d /s /c` + 逐 token 引号 + verbatim，彻底移除 shell:true 消除 DEP0190）；本地复核 typecheck/test(1299)/build 全绿，diff 干净、两种压缩模式均不受影响（改动在 wire 之前的进程启动层），review 中已直接在分支补了 3 处缺失的测试覆盖（大写 .EXE、含空格 COMSPEC、dotfile 基名），可以合并。

PR #706, opened 2026-09-11
feat(pi): native plugin mode — self-bootstrapping proxy + fetch interception

ranxianglei:
## What

Bare `pi` — with this package installed via pi's package manager or `bili plugin install pi` — now gets full plugin-mode compression with **no launcher**:

```
pi package install billion-context   # (or: bili plugin install pi)
# 之后裸跑 pi 即可,什么都不用管
```

## How (all verified end-to-end on pi 0.83.6 / Node 25)

At extension load, `dist/agent/pi-native.js` (new tsup entry, now the `pi` manifest target):

1. **Spawns the package's own proxy** via `ensureProxyRunning` — ephemeral port, parent-pid watchdog = the pi process; a healthy compatible instance is **attached, not doubled** (#394/#417 semantics).
2. **Patches `globalThis.fetch`** (`src/agent/native-intercept.ts`) and rewrites model-API requests to `<proxy>/bili/<full-upstream-url>`. The load-bearing assumption from #519 is verified source-level and e2e: pi → `pi-stable-ai` → Anthropic/OpenAI SDKs resolve fetch from the global at **first-request client construction** (`this.fetch = options.fetch ?? getDefaultFetch()`), so a load-time patch always wins; SSE streams pass through untouched.
3. **Exports the existing `createBiliPlugin` factory** with `BILLION_CONTEXT_PROXY` set — native tools, `x-bili-plugin` headers, `/acp`, and the compaction cancel all reuse the launcher-mode code paths unchanged.

### Patch discipline

- Matches **only** model-API URL shapes: `(v\d+/)?(messages|chat/completions|completions|responses|conversations)` suffix — version-unpinned (zhipuai `/v4/chat/completions`), path-prefixed upstreams included (bailian `/apps/anthropic/v1/messages`).
- Never touches the proxy's own `/bili/`·`/__bili/` paths (self-guard) or non-HTTP(S) URLs.
- Pre-ready requests **wait** (≤15s) for the bootstrap; failure/timeout degrades to a direct (uncompressed) send with a one-time warning — the agent never breaks.
- A `TypeError` (network failure) triggers **one respawn + one retry** — the watchdog reaps the proxy when the *owning* pi exits while later sessions still ride it.

### Gating

- Skipped entirely under a `bili` launch (`BILLION_CONTEXT_PROXY` / `BILI_PROVIDER_REWRITES` set) — `bili pi` keeps `-e dist/agent/pi.js` byte-identical.
[...]

PR #709, opened 2026-09-11
fix(launcher): cross-process startup-window race — starting marker prevents double-spawn (#707)

### LINKED ISSUE #707: bug: cross-process startup-window race — two concurrent launches each spawn a proxy (double writers over one sessions dir)
来源: [#706](https://github.com/ranxianglei/billion-context/pull/706) 分析 feat(pi): native plugin mode 时发现

## What

`ensureProxyRunning` (#394/#417 attach semantics) only **attaches** when it finds a *healthy* instance at probe time. Two processes calling `ensureProxyRunning` inside the probe→healthy window (child spawn + health polling, ~1s–several s) both see "no healthy instance" and **each spawn a child**. Result:

- two live proxies sharing ONE sessions dir — exactly the double-writer situation #394/#417 was designed to prevent (that fix covers the *healthy* case only);
- the instance file ends up last-writer-wins;
- the "orphaned" extra proxy is not reaped early: its parent-pid watchdog tracks ITS OWN spawner, so it lives until that client exits.

## Repro

Launch the same client twice within ~1s of each other while the first proxy is still coming up (e.g. `bili codex` ×2 concurrently, or — since #706 — two bare `pi` terminals opened back-to-back, which now self-bootstrap at extension load). Observe two `node dist/index.js start` children and two "proxy started" log entries.

## Why #706 raises the stakes

Before #706 the window was reachable only via deliberate concurrent `bili <client>` launches. Native mode makes self-bootstrap the default path for every bare pi start, so back-to-back terminal opens become a realistic trigger.

## Impact

Concurrent writes over one sessions dir (state corruption risk), duplicated proxy processes, instance-file flapping between the two origins.

## Suggested fix

Publish a "starting" marker BEFORE spawning (e.g. write the instance file with the launchToken up front, or a separate lockfile held across the spawn window). Concurrent probes then find an in-progress bring-up and WAIT for health instead of spawning. In-process callers are already covered by the `singleFlight()` wrapper added in #706's review commit (3cb0331); this issue is the cross-process half.

ranxianglei:
## What

Fixes #707. Two concurrent `ensureProxyRunning` calls inside the spawn→healthy window each spawned their own proxy child: two live proxies over one sessions dir (double writers), last-writer-wins instance file, and the orphaned extra proxy living until ITS OWN spawner exited.

Root cause: cross-launch coordination happened only through `<state>/proxy-origin`, which the child writes *after* its first successful bind (`announceListening`, src/server.ts). Between the launcher's `spawnImpl` and the child's first listen there was zero on-disk evidence of an in-flight bring-up, so concurrent probes found nothing and both spawned. Ephemeral ports (#446) let both children bind different ports cleanly; an explicit port just made the second retry port+1 (#407) — either way, two writers.

## Fix

The starter publishes a "starting" marker **before** spawning:

- New `<state>/proxy-starting` file, claimed with O_EXCL (`fs.openSync(path, "wx")`) — atomic across processes; a rename-based write would leave the same race at the marker level. Content: `{token, pid, host, port, startedAt}`.
- A concurrent caller that finds an **active** marker (owner pid alive AND younger than `SPAWN_WAIT_MS + 30s`) waits for the starter's instance using the same probe as attach (pid alive + `/__bili/health` + instanceId match), bounded by `SPAWN_WAIT_MS`, then attaches. Marker disappearance (starter gave up / crashed) breaks the wait early so the caller spawns immediately instead of burning the full window.
- The starter clears the marker on every terminal path via token-checked clear (`try/finally` around spawn→poll→throw). A hard crash leaves a stale marker that the dead-owner/TTL checks render inert.
- Incompatible-config fall-through preserved: a waited-for instance that isn't compatible with our launch options still spawns (matches today's incompatible-healthy-instance behavior). Explicit `bili start` bypasses the marker entirely (user action, not auto-bootstrap). In-process dedup (`singleFlight`, #706) untouched — this is the cross-process half.

## Verification

- **Repro before fix** — two concurrent `PI_BIN=<fake-client> node dist/index.js pi --port 0` sharing one isolated `XDG_STATE_HOME`: **2** proxy children, two `started proxy` log lines, two registry markers, instance file flapping between origins.
- **Same repro after fix**: **1** proxy child; the second launcher logs `another bili launch is bringing up a proxy — waiting for it instead of spawning a second` → `attaching to running proxy at http://127.0.0.1:44551 (pid …)`; exactly one registry marker; no leftover marker file.
- **New tests** — 7 launcher tests: active-marker wait+attach without spawn; dead-owner stale removal+spawn; expired-marker removal+spawn; early bail on marker disappearance (<2s); starter claims before spawn and clears after success; child-exit failure path clears marker; incompatible bring-up falls through to spawn. Plus 2 marker-mechanics unit tests (O_EXCL exclusivity, token-checked clear, garbage-file degrade path). `tests/launcher.test.ts` now isolates `XDG_STATE_HOME` per run so marker I/O stays hermetic under parallel `node --test`.
- **Preflight**: `npm run typecheck` clean · `npm test` 1376 pass / 0 fail (2 skipped = gated e2e) · `npm run build` ok.
- **e2e note**: the gated `e2e-codex` suite needs a Responses-compatible upstream (default `http://127.0.0.1:8199/v1`), which is not reachable from this environment; the change touches only the launcher bootstrap path (`src/launcher.ts` + `src/instance.ts`), not the request pipeline, and is covered by the unit tests + the empirical concurrent-launch repro above.

中文摘要: 修复了跨进程启动窗口竞态——启动方在 spawn 前用 O_EXCL 发布 `proxy-starting` 标记,并发的启动会等待该实例就绪后 attach,而不是再 spawn 一个代理(消除同一 sessions 目录的双写者),可以合并。


ranxianglei:

[bot] 🏷 Reviewing PR #709 (fixes #707) now — checking diff cleanliness, the O_EXCL marker mechanics in `src/launcher.ts` / `src/instance.ts`, running typecheck + tests + build, and reasoning through both compression modes per AGENTS.md §6. Will report back with findings.

ranxianglei:

[bot] 🏷 **Review: LGTM after two fixes I pushed directly to the branch** (commit `61ae4fd`, now on `2026-09-11_fix-707-startup-race`).

Diff cleanliness first pass: clean — 5 files, all on-topic (`src/launcher.ts`, `src/instance.ts`, 2 test files, CHANGELOG), no drive-by changes. The core design is sound: O_EXCL claim before spawn, active-marker wait with attach, token-checked clear on every terminal path, dead-owner/TTL stale rendering, incompatible fall-through preserved, `bili start` bypass structural (it never goes through `ensureProxyRunning`), and `singleFlight` (#706) correctly left alone. I verified the concurrency invariants hold: a hung owner can't delete our marker (token-checked clear), and unconditional removal of a stale marker can't clobber a newer claimant (O_EXCL bars any new claim while the old file exists).

Two residual double-spawn windows remained, both in the exact invariant this PR establishes — fixed in `61ae4fd`:

**1. Read→claim TOCTOU (the real one).** The pre-spawn `readStartingMarker()` was treated as the coordination decision, but the `claimStartingMarker()` result at `src/launcher.ts` was **ignored** — on EEXIST the code spawned anyway. Two processes that both read "no marker" before either claimed would each spawn: the loser's failed claim changed nothing. The default port-0 path widens the window with `await pickEphemeralPort()` between read and claim (a full listen/close round-trip), so this wasn't theoretical microsecond luck — two near-simultaneous `bili pi` launches could hit it. Now the O_EXCL claim itself is the arbiter: a claim loser re-reads the marker → active holder means wait for its instance (same probe as attach); stale/unreadable file means remove-and-reclaim once; still losing degrades to a plain spawn. Side benefit: an unreadable marker file (crash mid-write leaving partial JSON) previously blocked *every* future claim forever, silently degrading all launches until manual deletion — it now self-heals on the next launch.

**2. Deadline-boundary sliver.** The waiter's deadline coincides with the starter's own poll deadline. A starter succeeding in its last ~200 ms could be missed by the waiter's final poll (200 ms interval + probe latency), which then fell through to spawning a duplicate over a healthy instance. Subtle because the success path *clears* the marker — to the waiter it looks identical to a failure bail. `waitForStarterInstance` now ends with one final probe after the loop exits (cheap: ENOENT/dead-pid short-circuit; only costs a health fetch when a live record exists).

Tests added for both (plus a bounded-wait pin for the hung-starter worst case): garbage-marker self-heal, deadline-boundary attach-instead-of-double-spawn, hung starter waits max twice then spawns. I verified the first two **fail against the pre-review code** (ran them against `2a7e5d1`: the self-heal test sees no re-claimed token; the boundary test hits the double-spawn throw) and pass with the fix.

Preflight on the updated branch: `npm run typecheck` clean · `npm test` **1379 pass / 0 fail / 2 skipped** (was 1376; +3 new) · `npm run build` ok. E2E note stands as you wrote it — the gated `e2e-codex` upstream isn't reachable from my environment either (confirmed), and the diff touches only the launcher bootstrap path, which the unit tests + your concurrent-launch repro cover.

中文摘要:审查发现并直接修复了两处残留的双开窗口——(1) 读标记与 O_EXCL 抢占之间的 TOCTOU(抢占失败被忽略,仍会再 spawn 一个,默认端口 0 路径因 pickEphemeralPort 异步操作窗口更大),现改为以抢占本身为仲裁者,失败方重查后等待或清除重抢,顺带让崩溃残留的垃圾标记文件可自愈;(2) 等待截止时间与启动方自身轮询截止重合,启动方在最后 ~200ms 成功时等待方可能漏探而重复 spawn,现循环结束后补一次终探。两个新回归测试均验证过在修复前代码上失败、修复后通过;全量测试 1379 通过,可以合并。

PR #697, opened 2026-09-10
test: add containerized real-codex E2E (docker) (#686)

ranxianglei:
## What
Containerized **real** end-to-end for codex, per the ask in #686 (“需要真的端到端…docker 起 codex 的那种”). It wraps the existing `tests/e2e/e2e-codex.test.ts` suite -- real `codex` CLI → bili proxy → real Responses upstream → full warmup / load-growth / ACP-compress / purity(/forge) lifecycle -- inside a reproducible Docker image, following the containerized-e2e pattern from Tyan66666/billion-context-dsh#128.

This is "real" in the sense #686 wants: actual model round-trips and compression through the proxy, not just the headless `mcp list` config check from PR #688. The two are complementary (#688 = Windows MCP-registration layer; this = full lifecycle in an isolated container).

## Files
- `scripts/e2e-docker/Dockerfile` -- `node:22-slim` + `npm ci` + `npm run build` + `@openai/codex`; runtime baked into the image, host needs no node/codex.
- `scripts/e2e-docker/run.sh` -- host orchestrator: build image → run the suite in-container, forwarding `E2E_UPSTREAM_URL/KEY/MODEL/FORGE/TMO/CODEX_BIN`, mounting `tmp/docker-e2e` for persistent logs.
- `.dockerignore` -- keeps host `node_modules`/`dist` out of the build context.
- `.github/workflows/ci-e2e-docker.yml` -- `workflow_dispatch` job mirroring `ci-e2e.yml` (secrets-backed upstream); runs preflight then full.
- `package.json` -- `npm run e2e:docker` / `e2e:docker:preflight`.

## Local usage
```bash
# zero-token smoke (boots container, prints codex version + dist + upstream probe)
E2E_UPSTREAM_URL=http://host.docker.internal:8199/v1 E2E_UPSTREAM_KEY=... npm run e2e:docker:preflight
# full lifecycle
E2E_UPSTREAM_URL=... E2E_UPSTREAM_KEY=... npm run e2e:docker
# if the upstream is on the host loopback, share the host netns instead:
E2E_DOCKER_NETWORK=host E2E_UPSTREAM_URL=http://127.0.0.1:8199/v1 npm run e2e:docker
```

## Validation done here (Linux sandbox)
- `bash -n scripts/e2e-docker/run.sh` clean; workflow YAML parses; `package.json` valid JSON, **version untouched**.
[...]

PR #688, opened 2026-09-10
test: add Windows codex overlay MCP-registration check (#686)

ranxianglei:
## What
Adds a `windows-latest` CI job + `scripts/windows-codex-mcp-check.ps1` that closes the **real-machine gap** in #686: it confirms that `codex` accepts the `<CODEX_HOME>-bili` overlay produced by `bili codex` and registers the `bili` MCP server from it.

## Why headless / key-free
`bili codex mcp list` starts the proxy, points `CODEX_HOME` at the generated overlay (win32), and forwards `mcp list` to codex. `codex mcp list` only reads `config.toml` and enumerates MCP servers -- it never calls a model -- so **no `E2E_UPSTREAM_*` secrets are required**. The existing secret-backed ubuntu e2e (`ci-e2e.yml`) still covers the full model round-trip; this adds the Windows MCP-registration layer for free.

## How the script verifies
- Asserts on **codex stdout only** (`2>$null` drops bili's own stderr log lines), so bili's own "injecting native bili MCP tools" line cannot cause a false positive.
- Hard pass = `mcp list` shows `bili`; soft check = `mcp get bili` shows `BILI_MCP_PROXY`.
- Writes a transcript artifact for reporting.

## Pre-flight (run here, on Linux)
- `typecheck` clean, `build` success, `tests/launcher.test.ts` **109/109 pass** (includes #683's overlay tests).
- Confirmed **#679 is intact** alongside #683 (`planClientSpawn`/`quoteWinToken`/`buildWindowsCommandLine` all present in `src/launcher.ts`).
- A fake-`codex` harness proved `bili codex mcp list` **boots the proxy with no upstream/key**, forwards `mcp list`, and sets `HTTPS_PROXY`; the win32 `CODEX_HOME=<overlay>` env-patch was proven separately by driving the real `prepareCodexHome`.
- Python `tomllib` confirmed the delivered `config.toml` is **valid TOML even for a spaced `C:\Program Files\...` node path** (the #681 crux) with exactly one `[mcp_servers.bili]` block.

## Review note
The verify step uses `continue-on-error: true` **deliberately** until one clean run proves codex boots unauthenticated on `windows-latest` (this agent runs on Linux and could not drive real Windows/codex). The transcript artifact shows exactly what happened so it can be flipped to a hard gate with confidence. This does **not** close #686 -- the interactive/compress-tool + concurrent-window acceptance items still need a human on a real box (step-by-step instructions posted in the issue thread).

Refs #686

---
[...]

PR #715, opened 2026-09-11
test: deterministic fake-upstream real-codex compression E2E (#686)

ranxianglei:
## What
The "real regression" requested in #686 — compression **actually happens inside real `codex`**, driven by a **deterministic fake Responses-API upstream**. No model, no tokens, no external secrets, so it can gate every PR.

Distinct from the two earlier #686 follow-ups:
- PR #688 = Windows MCP-registration layer (config-level, headless).
- PR #697 = full lifecycle in Docker against a **real** model upstream (needs secrets).
- **This** = real codex + real in-process compression, but with a **fake** upstream → deterministic and secret-free → runnable on every PR.

## How it works
- `tests/e2e/fake-upstream.mjs` — standalone fake serving the two wire contracts verified against `codex-cli 0.147.0`:
  - `stream:true` → Responses SSE assistant turn;
  - `stream:false` → plain JSON body that bili's summarization call parses (`extractSummaryText` → `json.output[].content[].text`).
  - Recognises bili summarization calls by their exact TASK text in `instructions`; replies with a **faithful summary** preserving the sentinel numbers found in the segment, dropping bulk filler. Every `/v1/responses` request is logged to `FAKE_REQLOG` as the assertion oracle.
- `tests/e2e/e2e-codex-fake.test.ts` — gated by `ACP_TEST_E2E_FAKE=1`:
  - **overflow** (window 12k): warms up planting sentinels, then pushes ~10KB of bulky filler on every load turn via `resume --last`. Asserts: a real compression event fires (log), summarization is called repeatedly (oracle), the forwarded payload **stays bounded** despite heavy injection (compression folds the bulk instead of letting it accumulate), sentinels survive into the final payload, and the `[Compressed conversation section]` block is present.
  - **control** (window 60k): one small turn asserts **no** compression and no summarization calls (no false positives).

## Local usage
```bash
npm i -g @openai/codex@latest   # or set E2E_CODEX_BIN
ACP_TEST_E2E_FAKE=1 npm run test:e2e:fake
```

## Validation (this sandbox, Linux)
Ran against the **real `codex-cli 0.147.0`**: typecheck clean · build ok · **2/2 pass** (~64s). The overflow run shows repeated `preflight compressed` events and a forwarded payload that grows only ~1KB across four ~10KB filler injections — i.e. compression is genuinely folding context inside codex.

## Caveat / review note
The fake encodes the exact SSE/JSON contract of `codex-cli 0.147.0`. If `@openai/codex` changes its wire format on a future `@latest` bump, this test fails loudly (by design) and the fake needs updating — same maintenance trait as the existing `tests/e2e/e2e-codex.test.ts`. Nothing pinned, matching that convention.

## Scope vs #686
Covers the "real compression really happens in codex + assert behaviors" ask. The Windows-specific acceptance items (concurrent-window `livePidHoldsOverlay` warning, interactive compress-tool visibility) still need a human on a real Windows box — tracked under #688 / #683.

Refs #686

---
中文摘要：按你"用 fake api 造数据、让压缩真的在 codex 里发生、并断言一系列行为"的要求，加了一个确定性假上游 E2E（真实 codex + 进程内真压缩，零 token/零密钥），可作每次 PR 的门禁；已在本地对真实 codex-cli 0.147.0 验证 typecheck/build/2 用例全绿。

### DIFF
--- a/.github/workflows/ci-e2e-fake.yml
+++ b/.github/workflows/ci-e2e-fake.yml
@@ -1,0 +1,47 @@
+name: e2e-fake-codex
+
+on:
+  pull_request:
+  workflow_dispatch:
+
+concurrency:
+  group: e2e-fake-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  e2e-fake:
+    name: real codex vs deterministic fake upstream (zero tokens)
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+          cache: npm
+
+      - name: Install codex CLI
+        run: npm i -g @openai/codex@latest
+
+      - name: Install deps
+        run: npm ci
+
+      - name: Build
+        run: npm run build
+
+      - name: Run deterministic real-codex compression E2E
+        run: ACP_TEST_E2E_FAKE=1 node --import tsx --test tests/e2e/e2e-codex-fake.test.ts
+
+      - name: Collect E2E artifacts
+        if: failure()
+        run: |
+          mkdir -p tmp/e2e-artifacts
+          for d in tmp/e2e-codex-fake-*; do [ -d "$d" ] && cp -r "$d" tmp/e2e-artifacts/ || true; done
+
+      - uses: actions/upload-artifact@v4
+        if: failure()
+        with:
+          name: e2e-fake-artifacts
+          path: tmp/e2e-artifacts
+          if-no-files-found: ignore
--- a/package.json
+++ b/package.json
@@ -15,7 +15,8 @@
     "typecheck": "tsc --noEmit --project tsconfig.build.json",
     "test": "node --import tsx --test tests/*.test.ts",
     "start": "node dist/index.js",
-    "test:e2e": "node --import tsx --test tests/e2e/e2e-codex.test.ts"
+    "test:e2e": "node --import tsx --test tests/e2e/e2e-codex.test.ts",
+    "test:e2e:fake": "node --import tsx --test tests/e2e/e2e-codex-fake.test.ts"
   },
   "keywords": [
     "context",
--- a/tests/e2e/e2e-codex-fake.test.ts
+++ b/tests/e2e/e2e-codex-fake.test.ts
@@ -1,0 +1,228 @@
+import { spawn, spawnSync } from "node:child_process";
+import fs from "node:fs";
+import net from "node:net";
+import path from "node:path";
+import { strict as assert } from "node:assert";
+import { test } from "node:test";
+
+const CODEX_BIN = process.env.E2E_CODEX_BIN ?? "codex";
+const DIST = process.env.E2E_BILI_DIST ?? path.resolve(import.meta.dirname, "../../dist/index.js");
+const MODEL = process.env.E2E_MODEL ?? "qwen3.8-27b";
+const TMO = Number(process.env.E2E_TMO ?? 120_000);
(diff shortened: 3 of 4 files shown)


ranxianglei:

[bot] 🏷 Reviewed PR #715 against real `codex-cli 0.147.0` in this sandbox. **Verdict: mergeable** — reproduced green end-to-end and verified every load-bearing assumption against source. No defects found that need a commit.

## What I ran (all green)
- `typecheck` clean · `build` ok · default unit suite **1371 pass / 0 fail / 2 skip**.
- E2E (`ACP_TEST_E2E_FAKE=1`): **2/2 pass in ~64s** — matches your claim.

## The test is not vacuous — oracle artifacts confirm real in-process compression
I inspected the fake's request log after the run:
- **overflow** dir: 10 requests = 5 summarization + 5 forwarded. Forwarded payload sizes `[80420, 80660, 80944, 81228, 81512]` → grew only **~1.1KB across four ~10KB filler injections** (peak/min ratio **1.014**, well under the `1.3` bound). Final payload carries `[Compressed conversation section]` and **all three planted sentinels (4781 / 2903 / 6577)**. So compression genuinely folds the bulk instead of accumulating it.
- **control** dir (60k window): 1 request, **0** summarization calls → no false positive.

If the filler weren't being folded, turn 5 would be ~+40KB and the ratio assertion would fail — so the bound has real discriminating power.

## Load-bearing assumptions verified against source
- `src/preflight.ts:227` — Responses `summaryPayload` puts the summary TASK into `instructions`, which is exactly what the fake's `isSummaryRequest` inspects.
- `src/preflight.ts:312` — the TASK text contains both regex alternatives the fake matches ("must be compressed because the session context exceeds" + "Write a tier-1 compression summary").
- `BILI_LAUNCHER_MODEL_WINDOWS` (`src/server.ts:162`) and `--no-auto-update` (`src/cli.ts:174`) are real; the `/bili/<absolute-url>` base_url matches the documented tunnel convention (README).
- Diff is clean: 4 purpose-related files, `package.json` adds only the `test:e2e:fake` script (**no version bump**), no lockfile/generated churn. Code follows §3 (precise type assertions, no `as any`/`@ts-ignore`).

## Non-blocking notes
- **This becomes a mandatory gate on every PR.** That's sound here because it's secret-free and self-contained (spawns its own fake upstream), unlike `ci-e2e.yml` which needs secrets + an unreachable localhost upstream and stays manual-only. The one real cost is already acknowledged in the PR: `@openai/codex@latest` wire-format drift will fail **all** PRs until the fake is updated. Flagging so you're consciously accepting that maintenance tradeoff for repo-wide gating.
- The `peak <= minLen * 1.3` bound passed with a comfortable margin (1.014); it holds because codex's fixed scaffolding dominates the baseline and the folded filler is small relative to it. Left as-is.
- Scope vs #686 looks right: Windows-specific items are correctly deferred to #688/#683.

## Duplicate screening
Distinct from the other two #686 follow-ups — #688 (Windows MCP-registration, config-level/headless) and #697 (Docker full lifecycle vs a **real** model, needs secrets). This one is the deterministic fake-upstream variant (real codex + in-process compression, zero tokens/secrets). Not a duplicate; keep open.

---
中文摘要：在真实 codex-cli 0.147.0 上复现了 typecheck/build/单测(1371 过)/E2E(2/2 约 64s)全绿，并核对假上游的断言确实有效（溢出用例转发体四次注入仅涨约 1.1KB、保留全部哨兵值，对照用例零压缩），关键假设逐条对源码核实无误，diff 干净无版本号改动，可合并；唯一需你确认的是它成为每次 PR 的强制门禁后，codex 线格式漂移会阻塞所有 PR，这点维护成本请知悉。

Issue #714, opened 2026-09-11
[ACP] No blocks matched 报错


chasel1681:
❌ [ACP] No blocks matched "Newton Agent platform IOP BFF task persistence authorization audit notification TAPD 1000829"

📊 [ACP] acp_status result:
CONTEXT BREAKDOWN
  2.8K tool (53%) | 2.5K text (47%) | 0 summaries (0%)
  Top tools: text (47%), exec (38%), tool (14%)
COMPRESSED BLOCKS
  No compressed blocks.
Tip: buildStatusReport({scope:"uncompressed", view:"messages", tool:"text"}) for per-message listing
Nudge: idle — max compressible 0 < threshold 50000; growth 3093 < floor 22500

❌ [ACP] No blocks matched "Newton Agent platform IOP BFF task persistence authorization audit notification TAPD 1000829"

❌ [ACP] No blocks matched "Newton architecture existing implementation OpenAICompatibleProvider OllamaChatModel"

❌ [ACP] No blocks matched "newton-agent-platform-iop specification"

📊 [ACP] acp_status result:
CONTEXT BREAKDOWN
  2.9K tool (52%) | 2.7K text (48%) | 0 summaries (0%)
  Top tools: text (48%), exec (36%), tool (14%)
COMPRESSED BLOCKS
  No compressed blocks.
Tip: buildStatusReport({scope:"uncompressed", view:"messages", tool:"text"}) for per-message listing
Nudge: idle — max compressible 0 < threshold 50000; growth 3578 < floor 22500

❌ [ACP] No blocks matched "Newton Agent platform IOP BFF task persistence authorization audit notification TAPD 1000829"

❌ [ACP] No blocks matched "newton-agent-platform-iop specification"

chasel1681:
📊 [ACP] acp_status result:
CONTEXT BREAKDOWN
  23.4K tool (79%) | 6.2K text (21%) | 0 summaries (0%)
  Top tools: exec (79%), text (21%)
COMPRESSED BLOCKS
  No compressed blocks.
Tip: buildStatusReport({scope:"uncompressed", view:"messages", tool:"exec"}) for per-message listing
Nudge: idle — max compressible 16364 < threshold 50000; growth 21587 < floor 22500
Compressible ranges (2, oldest first):
  m00001–m00006  6 msgs  6.1K [tool 0% | text 100%]
  m00008–m00011  4 msgs  10.3K [tool 50% | text 50%]




📊 [ACP] acp_status result:
[...]

PR #716, opened 2026-09-11
fix: search_context with zero blocks reports empty state, not a failure (#714)

ranxianglei:
## Problem (#714)

Before any compress runs, `search_context` returns `[No blocks matched "<query>"]`, and `buildVisibilityMarker` classifies that string as a **failure** (`❌ [ACP] No blocks matched ...`). In the reporter's session the model saw repeated ❌ errors while every `acp_status` showed 'No compressed blocks' — and retried the same searches ~8 times. Empty results are being presented as hard errors, driving premature-search retry loops.

## Fix

- **New shared helper** `executeSearchContext(args, core, state)` in `src/decompress-shared.ts` (the existing home of `resolveDecompress` + `ProxyToolCtx`), replacing the three duplicated handlers in `src/loop/core.ts`, `src/stream.ts`, `src/compress-loop-responses.ts`.
- Distinguishes the two empty cases:
  - **zero active blocks** → `[No compressed blocks exist yet — nothing to search.]` (searching is pointless until compress runs)
  - **blocks exist, none matched** → `[No blocks matched "<query>"]` (unchanged wording)
- `buildVisibilityMarker` no longer treats `"No blocks matched"` as a failure: empty results now render with the 🔍 search icon instead of ❌. Real failures (`FAILED` / `not found` / `is required`) still render ❌.

Proxy-mode only by design: plugin-mode agents execute tools in their own process (separate repos); this is where the reported rendering comes from.

## Pre-flight

- `npm run typecheck` ✅
- `npm test`: 1243/1244 pass — the single failure (`launcher.test.ts resolveClientCommand`) is environmental (sandbox has `/usr/bin/codex` installed) and fails identically on clean master
- `npm run build` ✅
- New tests: `tests/search-context.test.ts` (marker icon matrix + helper: missing query / zero blocks / no-match / match) and `tests/loop-core.test.ts` `loop #714` integration test (zero-block round surfaces 🔍 empty-state marker, never ❌)

### DIFF
--- a/src/compress-loop-responses.ts
+++ b/src/compress-loop-responses.ts
@@ -8,7 +8,7 @@ import { lastCompressSuffix, type Session } from "./session.js";
 import { parseCompressInput, PROXY_TOOL_NAMES, MUTATING_PROXY_TOOLS, COMPRESS_TOOL_NAME, ACP_TEXT_OPEN, ACP_TEXT_CLOSE } from "./compress-tool.js";
 import { log as loggerLog } from "./logger.js";
 import { applyRanges } from "./stream.js";
-import { resolveDecompress } from "./decompress-shared.js";
+import { executeSearchContext, resolveDecompress } from "./decompress-shared.js";
 import { buildVisibilityMarker } from "./compress-loop.js";
 import { MAX_LOOP_ROUNDS } from "./loop/index.js";
 import { stripResponsesText } from "./loop/tag-echo-filter.js";
@@ -91,17 +91,7 @@ function executeProxyTool(
         return resolveDecompress(args, ctx);
     }
     if (toolName === "search_context") {
-        const query = typeof args.query === "string" ? args.query : "";
-        if (query.length === 0) return "[search_context FAILED: query is required]";
-        const limit = typeof args.limit === "number" && args.limit > 0 ? Math.floor(args.limit) : 5;
-        const blocks = ctx.core.search(query, ctx.session.state).slice(0, limit);
-        if (blocks.length === 0) return `[No blocks matched "${query}"]`;
-        const lines = blocks.map((b) => {
-            const topic = b.topic ?? "(no topic)";
-            const preview = b.summary.length > 200 ? b.summary.slice(0, 200) + "..." : b.summary;
-            return `${b.blockId} (T${b.tier}) "${topic}"\n  ${preview}`;
-        });
-        return `Found ${blocks.length} block(s) for "${query}":\n\n${lines.join("\n\n")}`;
+        return executeSearchContext(args, ctx.core, ctx.session.state);
     }
     if (toolName === "acp_status") {
         return handleAcpStatus(args, ctx);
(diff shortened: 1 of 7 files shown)


ranxianglei:

Reviewed — **ready to merge** (merge itself stays with you). I also rebased the branch: it was ~20 commits behind master (base was pre-v0.1.106, and both `src/stream.ts` and `src/loop/core.ts` had moved underneath it). Trial merge auto-resolved cleanly, so I rebased directly on the PR branch — new head **590e015**, identical 7-file diff (155+/37−), nothing lost.

**Diff cleanliness**: every file maps to #714 — shared helper + 3 handler call-sites + marker heuristic + 2 test files. No version bump, no lockfile churn, no unrelated edits, no `as any`/`@ts-ignore`.

**Logic verification** (not just reading the diff):
- The zero-blocks guard `state.blocks.some((b) => b.active)` is consistent with the kernel: `core.search()` internally filters through `activeBlocks(state)` (checked in acp-kernel dist), so "no active blocks" exactly mirrors what search can see.
- Grep confirms the three replaced handlers (`src/loop/core.ts`, `src/stream.ts`, `src/compress-loop-responses.ts`) are the only execution sites; the adapters only declare tool names.
- The new empty-state string contains none of the remaining failure substrings (`FAILED` / `not found` / `is required`), so it renders 🔍 while real failures keep ❌ — matches the new marker-matrix test.
- Proxy-mode-only by design holds: plugin-mode agents execute tools in their own process, and this change touches no wire format.

**Pre-flight on rebased head** (fresh install, master pins acp-kernel 0.0.63):
- `npm run typecheck` ✅
- `npm test`: 1382 run, **0 failures**. Note: the `launcher.test.ts` failure cited in your description (environmental `/usr/bin/codex`) no longer reproduces — master's launcher-test refactor landed since the old base, so it's moot now.
- `npm run build` ✅

**E2E caveat**: I could not complete the real-codex e2e suite here — the default local sglang upstream documented in `tests/e2e/README.md` is down (connection refused through the proxy; the proxy itself surfaced it correctly as a structured 502, which is actually a nice incidental check). Recommend running the manual-dispatch CI e2e workflow (`.github/workflows/ci-e2e.yml`, has its own upstream secrets) or re-running locally once that upstream is back, before merging.

一句话中文摘要:修复了 search_context 在零压缩块时把空结果渲染成 ❌ 失败、诱发模型反复重试的问题(新增共享 helper 区分"无块可搜"与"无匹配",标记启发式不再把空结果判为失败);已 rebase 到最新 master,typecheck/test/build 全绿,仅本地 e2e 因上游服务未启动未能跑完(建议合并前在 CI 手动触发一次 e2e),可以合并。

ranxianglei:

Small related observation from the e2e attempt above (minor, report-only): the `E2E_CHECK=1` preflight at `tests/e2e/e2e-codex.test.ts:308-321` only **logs** the upstream `/models` probe result — it never asserts on it — so it exits green even when the upstream is down (which is exactly how I misread it before the full run failed). AGENTS.md describes the preflight as "codex binary + dist + upstream reachable"; either assert `probe !== "unreachable"` in the preflight, or soften that doc line. Not blocking this PR.

PR #718, opened 2026-09-12
fix: strip model-emitted ACP confirmation markers and state the marker contract (#717)

ranxianglei:
## What

A long Kimi Code session (#717) showed a model **writing the proxy's own `📦 [ACP] Compressed …` confirmation-marker format as plain assistant text** — 17 fake "compressions", zero real tool calls — while usage climbed to 89%. Real markers never pass through the model-output path (the proxy injects them itself via `emitMarker`), so any marker-shaped line in upstream model output is by definition forged. This PR makes that forgery impossible to persist:

1. **Filtering (core fix).** New marker-line stripper in `src/loop/tag-echo-filter.ts`, layered onto the existing tag-echo filter (`composeStreamFilters`) at every model-output path: all three compress-loop adapters (anthropic/openai/responses), both plugin passthrough pipes, and the three non-stream JSON rewriters. Stripped lines log `[marker-echo] stripped model-emitted ACP confirmation marker: …`.
   - Shape: line start + exactly one non-ASCII symbol char (every real icon is a single code point) + optional space/tab + literal `[ACP]`. Strictly no leading whitespace — an indented occurrence is quoting the format (code block/docs) and passes through. CJK prose like `注意 [ACP] 标记…` does not match (two chars before the bracket).
   - Multi-line `acp_status` variant: head line stripped only; the body becomes inert unattributed prose (no reliable terminator exists for safe body-swallowing).
   - Streaming-safe: holds ≤ ~8 chars while a line start is still decidable; flush resolves at block/stream end with content preservation. Batch and streaming paths are proven equivalent at every split position (property test).
2. **Prompt contract.** New `withMarkerIntegrityNote` (`src/compress-tool.ts`) states explicitly that markers are proxy-emitted only, the model must never write one, and verification goes through `acp_status`. Appended to every nudge (3 sites) and to the injected philosophy prompt (4 sites, byte-stable constant → prefix-cache safe; rides only where the compress prompt is injected).
3. **Docs.** README (+zh-CN): a visible marker line is not proof of persistence — verify with `acp_status` (block count +1, range start advanced).

Issue's suggested fix #2 (per-session nonce) was evaluated and skipped as redundant: with the filter in place, *any* marker-shaped line in model output is already detectable by construction — a nonce would add plumbing across adapters/core/plugin without extra detection power. Fix #4 (detection) is subsumed by the new warn log.

Both compression modes covered: plugin-mode parity via the passthrough pipes (which also serve skip-injection proxy streams, #460).

## Pre-flight

- `npm run typecheck` — clean
- `npm test` — 1386 pass / 0 fail / 2 skipped (e2e gate)
- `npm run build` — clean
- `E2E_CHECK=1 node --import tsx --test tests/e2e/e2e-codex.test.ts` — 1 pass / 1 skip (zero-token preflight; full ACP_TEST_E2E run needs a live upstream, not available in this sandbox)
- New suite `tests/marker-echo.test.ts` (12 tests): exact observed forgery stripped; ❌/📊/all-icon variants; false-positive guards (indent, ASCII lead, mid-line, CJK); raw-wire gate; streaming↔batch equivalence at every split + arbitrary chunking; onDrop-once; flush content-preservation; compose stats merge; E2E through `runCompressLoop`+anthropic adapter asserting client output is clean and `[marker-echo]` warn fires; all three non-stream rewriters; note wrapper.


ranxianglei:

[bot] 🏷 Review started. Diff is compact and on-topic (13 files, no unrelated churn): core stripper in `src/loop/tag-echo-filter.ts`, wiring in the 3 adapters + 2 passthrough pipes + 3 non-stream rewriters, prompt contract in `src/compress-tool.ts`, docs, and the new test suite. Verifying the #717 claims against the code paths now, then running typecheck/test/build locally before the verdict. Will follow up with findings (and direct fixes for anything small).

ranxianglei:

Review complete — I found **one real coverage gap** and **one precision defect**, both fixed directly on the branch in **`e08e388`**. With those commits, this is ready to merge.

## Defect 1: the passthrough pipes' fast path bypassed the marker filter entirely (the core fix didn't actually cover plugin mode)

The three pipe fast-path gates (`pipePluginChatWithStrip` openai handler `src/plugin.ts:799`, anthropic handler `:851`, `pipePluginResponsesWithStrip` `:1096`) skip the composed filter whenever a chunk can't start a render tag:

```js
if (!mayStartRenderTag(v) && !anyPending()) { /* forward raw */ }
```

But `mayStartRenderTag` only matches `<acp`-ish fragments — a forged marker line contains no `<`, so every marker-shaped delta was forwarded **raw**:

- single chunk: `delta: "…\n📦 [ACP] Compressed m00876–m01100 → 1 block(s), ~54134 tokens saved.\n"` → passes through unchanged;
- split head: `"…\n📦 ["` + `"ACP] Compressed …"` → first chunk fast-pathed (prefix included), second seen fresh at line start with an ASCII lead → also passes.

Scope of the hole: **all plugin-mode streams** (pi/omp/opencode — the acp-native agents where marker-forgery pressure is highest) **and the #460 proxy-mode non-injection streams** (title-gen, `ACP_NO_INJECT_TOOL`, classifier bypass, native-compaction turns). The PR's "plugin-mode parity via the passthrough pipes" claim did not hold. The #717 scenario itself (proxy mode + Anthropic adapter) *is* covered because the compress-loop adapters push every text delta unconditionally — that part verified OK.

Fix: added `mayStartMarkerLine()` to `src/loop/tag-echo-filter.ts` — `s.includes("[ACP]")` or the chunk ends on an undecidable marker-head prefix at line start (`MARKER_TAIL`) — and wired it into all three gate conditions. Soundness: without the literal there is no complete marker line inside the chunk, and the tail probe catches heads split across boundaries; false positives cost one no-op filter pass. Added 4 pipe-level regression tests (responses whole-delta + split-head; chat anthropic whole-delta with no session to mirror the #460 path; chat openai split-head).

## Defect 2: icon class matched any non-ASCII char — silent stripping of legitimate prose

`[^\x00-\x7F]` matches CJK letters too, so a line starting with one CJK char directly before `[ACP]` — e.g. `见[ACP]标记的含义如下` (natural Chinese prose; no space between hanzi and ASCII) — was silently deleted from model output. That contradicts the documented shape ("exactly one non-ASCII symbol char") and turns a forgery filter into a data-loss path for zh-CN users. All six real icons (📦📤🔍📊🫧❌) are Unicode category `So`, so I narrowed `MARKER_HEAD`/`MARKER_LINE` to `\p{So}` — zero real-icon coverage lost.

One subtlety found while testing: `MARKER_HEAD_PREFIX` (streaming hold-back) must stay **broad**. Narrowing it too broke the existing every-split-position property test — a chunk boundary can split an astral icon into a lone surrogate, which `\p{So}` doesn't match, re-opening a leak. Broad prefix = lossless hold across the boundary; the strict `\p{So}` decision happens only in `MARKER_HEAD`. Documented at the regex. Added CJK false-positive guards to the test suite.

## Minor (fixed in the same commit)

- `src/server.ts` `injectSystem`: new line indented 8 spaces where the block uses 4 — diff-cleanliness fix.

## Verified OK

- Adapter push sites unconditional (anthropic/openai/responses), flush-at-stream-end present in both pipes, non-stream rewriters + full-text done-family payloads gated via `containsMarkerLineText`.
- Prompt contract wiring matches the description exactly: 3 nudge sites (`server.ts:2043/:2290/:2489`) + 4 philosophy-prompt sites; byte-stable constant → prefix-cache safe.
- Nonce skip rationale: agreed — with the filter, any marker-shaped model-output line is detectable by construction; a nonce adds plumbing without detection power.
- Diff otherwise clean and on-topic; docs additions correct.

## Minor observations (report only, no action taken here)

- PR body says the suite has 12 tests; `tests/marker-echo.test.ts` actually has 15 (now 20 after my additions).
- Pre-existing on **master**: `package-lock.json` root version is `0.1.105` while `package.json` is `0.1.106` — the v0.1.106 release commit didn't sync the lockfile (AGENTS.md §4 allows "+ package-lock.json if it drifts"). CI evidently tolerates it (v0.1.106 published fine), so nothing to do in this PR; noting for the record. I reverted my local `npm install` churn so the lockfile stays untouched on this branch.

## Pre-flight (after `e08e388`)

- `npm run typecheck` — clean
- `npm test` — **1391 pass / 0 fail / 2 skipped** (baseline 1386 + my 5 new tests)
- `npm run build` — clean
- `E2E_CHECK=1 node --import tsx --test tests/e2e/e2e-codex.test.ts` — 1 pass / 1 skip (zero-token preflight; full `ACP_TEST_E2E` needs a live upstream, not available in this sandbox — same constraint you noted)

修了快路径绕过（三处 pipe 门控只看 render-tag、放行了伪造标记行，导致 plugin 模式与 #460 流实际不受保护，补了 `mayStartMarkerLine` 门控+回归测试）和误杀缺陷（图标类收窄为 `\p{So}`，避免静默删除以单个汉字开头的合法 `[ACP]` 正文；流式保留前缀保持宽匹配以防 astral 图标跨块拆分泄漏），另修了一处缩进——修复后两个压缩模式均真正覆盖，可以合并。

PR #720, opened 2026-09-12
fix: harden rebuilt assistant content against null for OpenAI-compatible backends (#719)

ranxianglei:
## Problem (#719)

DSH plan mode, round 2+: upstream rejects with `400 Invalid assistant message: content or tool_calls must be set` (DeepSeek).

Verified root cause: when an upstream stream is truncated mid-reasoning (no completion event), the assistant turn ends up reasoning-only. `openaiToCore` drops empty text (`if (text)` is falsy for `""`), so BOTH `content:""` and `content:null` inputs rebuild as a reasoning-only core message — and `coreToOpenai` (acp-kernel 0.0.63) emits `content: null` for those turns (`dist/wire/index.js` flush(): `content: pending.text ?? null` / reasoning-only branch `content: null`). DeepSeek requires string content (possibly `""`) or tool_calls for assistant messages; the user empirically confirmed `content:""` + `reasoning_content` → 200 direct.

## Fix

Normalize at the proxy's outbound boundary instead of waiting on an acp-kernel release: new `hardenOpenaiAssistantContent()` in `src/util.ts` maps any assistant message whose `content` is `null`/`undefined` to `{...m, content: ""}` (identity no-op otherwise). Applied at both `coreToOpenai` call sites:

- `src/server.ts` `prepareOpenai` — main request pipeline
- `src/loop/adapter-openai.ts` `buildRequest` — proxy-side compress-loop rounds

Mode-neutral (plugin/proxy): pure outbound normalization after wire rebuild. Deterministic across turns → prefix-cache stable. These messages previously always 400'd on DeepSeek, so nothing working changes shape.

Note: the kernel-side equivalent (flush() emitting `""` instead of `null`) remains worth doing upstream; this PR makes the proxy robust regardless of kernel version.

## Tests

New `tests/openai-assistant-content-harden.test.ts` (6 tests): #719 repro asserting raw kernel output is `content:null` and hardened output is `""`; `content:null` input variant; tool_calls branch normalization with tool_calls preserved; identity no-ops; empty list; end-to-end `adapter.buildRequest` wire check.

## Pre-flight

- `npm run typecheck` — clean
- `npm test` — 1377 pass / 0 fail / 2 skipped (e2e gate)
- `npm run build` — success

Fixes #719

Issue #719, opened 2026-09-12
dsh第二轮plan计划报错问题


thanlife:
<html>
<body>
<h1 id="重建请求时只有-reasoning没有正文没有-tool_calls的-assistant-消息被写成-content-nulldeepseek-直接-400-invalid-assistant-message-content-or-tool_calls-must-be-set" tabindex="-1" data-source-line="1">重建请求时「只有 reasoning、没有正文、没有 tool_calls」的 assistant 消息被写成 <code>content: null</code>，DeepSeek 直接 400 <code>Invalid assistant message: content or tool_calls must be set</code></h1>
<h2 id="环境" tabindex="-1" data-source-line="3">环境</h2>
<ul data-source-line="5">
<li data-source-line="5">billion-context: 0.1.106（npm latest，bili 自检 <code>current=0.1.106 latest=0.1.106 tag=latest (up to date)</code>；找不到可升级版本）</li>
<li data-source-line="6">@deepseek-ai/dsh: 0.1.5-rc.1</li>
<li data-source-line="7">OS: Windows 11（Microsoft Windows NT 10.0.26200.0）</li>
<li data-source-line="8">Node.js: v24.21.0（<a href="http://nodejs.org">nodejs.org</a> 官方 Windows 安装包）</li>
<li data-source-line="9">客户端：DSH Web（3080），<code>bili dsh web</code> 启动，官方 DeepSeek 路由</li>
<li data-source-line="10">会话：<code>pfa-d350d5c128db6f88</code>（标题「我想要开发一款pve9的辅助工具…」）</li>
</ul>
<h2 id="现象" tabindex="-1" data-source-line="12">现象</h2>
<p data-source-line="14">DSH Web 里开启 <strong>plan 模式</strong>后，第一轮正常，<strong>第二轮</strong>起提示：</p>
<div class="code-block-shell"><pre class="cb-numbered"><code data-source-line="16"><span class="cb-line">本轮运行失败 Invalid assistant message: content <span class="hljs-keyword">or</span> tool_calls must be <span class="hljs-built_in">set</span></span></code></pre><button type="button" class="code-copy-button" title="复制" aria-label="复制">复制</button></div>
<p data-source-line="20">该会话在 bili 的统计里 <code>requests=2, inputTokens=0, cachedTokens=0, outputTokens=0</code> —— 两次请求都在上游就被拒了。</p>
<p data-source-line="22">同一报错文本在 bili 日志里精确出现一次：</p>
[...]

Issue #568, opened 2026-09-06
Preflight compression holds the request with zero response bytes; client undici headersTimeout (default 300s) kills it -> repeated `[preflight] summarization aborted: client disconnected` 5-minute death loop, payload never forwarded


shengbuding:
## Environment

- billion-context v0.1.86 (npm), proxy mode (`bili start --port <local>`), Windows.
- Client: dsh (DeepSeek Harness) CLI talking HTTP to the proxy's `/bili/<upstream>` origin; upstream = OpenAI-compatible remote endpoint (deepseek-v4-flash), also reproduced feeding a local llama.cpp server.
- Setups both drive the session well over the model window (e.g. ~954K tokens vs effective window 653,896) so every request enters preflight.

## Symptom

When the session context exceeds the model window, the proxy runs preflight compression **before emitting any response bytes to the client**. Compression of 700K鈥?50K-token histories takes 240鈥?07 seconds (summarization round-trips against the upstream model). The client's HTTP stack (Node undici, default `headersTimeout` = 300,000 ms) aborts the request while preflight is still running; the proxy then records:

```
[preflight] summarization aborted: client disconnected
[<session>] preflight aborted (the client disconnected during preflight compression); not forwarding
```

and never forwards. The client retries; the retry re-enters preflight; same 300s wall. Observed abort cadence is ~5 minutes apart, exactly matching undici's default 300 s `headersTimeout`, for 6 consecutive rounds (one session, 2026-09-06 UTC):

```
02:46:51  [preflight] summarization aborted: client disconnected 鈥?not forwarding
02:52:24  (same)         (+5m33s)
02:57:32  (same)         (+5m08s)
03:02:39  (same)         (+5m07s)
03:14:14  (same)         (+11m35s, compression round was longer)
03:19:22  (same)         (+5m08s)
```

Same failure repeated across three different sessions and both upstream types (remote + local llama.cpp), 10+ occurrences in one day. A successful (client-side, in-session) compression that shrunk the history 88% was the only thing that ended the loop; after that the session ran normally.

Adjacent evidence from the same window: a preflight that *did* finish shows the timescale 鈥?`preflight compressed 2 range(s), ~7435 tokens saved (702031 鈫?694596) in 265882ms; rebuilding payload` 鈥?265,882 ms is within spitting distance of the 300 s client timeout, and that very round is one of the aborted ones (the client died at ~300 s while compression was still running).

## Root cause (as observed, no suggested fix)

[...]

PR #722, opened 2026-09-12
fix: synthesize terminal/error sequence when plugin-mode streams end without one (#721)

### LINKED ISSUE #721: Plugin mode: upstream stream truncated before terminal event → client gets bare cut-off SSE (no completion/error event)
来源: ranxianglei/billion-context#719 分析 DSH plan 第二轮 "Invalid assistant message" 400 时发现

## Problem

In plugin mode (agent-owned compression, e.g. `bili dsh` / `bili omp`), when the upstream model stream terminates abnormally mid-turn (connection drop / timeout — no finish_reason, no [DONE]/message_stop/done-family event), the proxy closes the response right after the last received chunk. The agent client receives a bare cut-off SSE stream with NO completion or error event, and appends the partial assistant turn (e.g. reasoning-only) to its own history as if it were complete.

## Verified code paths (master f85553b, v0.1.106)

- Plugin pipes track terminal events (#498): `sawTerminal` in `pipePluginChatWithStrip` (src/plugin.ts:663+) and `pipePluginResponsesWithStrip` (src/plugin.ts:957+).
- On EOF with `!sawTerminal`, `maybeNoteTruncated` (src/plugin.ts:737-743, responses twin ~L1001) ONLY records server-side: `noteWeakOverflow(session, { reason: "plugin chat passthrough stream ended without a completion event" })`. Nothing is written to the client; `finally { res.end() }` (L921-924) closes the stream.
- Contrast — proxy-mode compress loop handles this properly: src/loop/core.ts:537-551 — `!sawDone || truncatedDone` → `noteWeakOverflow` + `yield adapter.emitError(msg)` (OpenAI: error frame + [DONE], see src/stream-error.ts); plus #413: zero-content truncation retries the fetch once.

## Consequence chain (observed in #719)

1. Upstream stream cuts mid-reasoning (dsh plan mode, round 1).
2. Agent appends the reasoning-only partial turn to its own history (no signal anything was wrong).
3. Next round resends it; bili rebuilds the request → assistant message with `content: null` (kernel coreToOpenai) → DeepSeek 400 "Invalid assistant message: content or tool_calls must be set".
4. PR #720 removes step 3's 400 (normalizes content:null → ""), but steps 1-2 remain: partial turns silently enter context without any abnormal-termination signal — they can confuse the model on later turns and hurt prefix-cache stability.

## Impact

- Agent clients cannot distinguish "model finished" from "stream died mid-turn".
- Partial turns are persisted into session history and replayed verbatim on every subsequent request.
- Only proxy-side log lines / weak-overflow state reveal the truncation; the user sees downstream symptoms (like the #719 400) on the NEXT turn.

## Suggested fix direction (needs owner decision on semantics)

In both plugin pipes, at EOF with `!sawTerminal` and `!res.destroyed`, synthesize a protocol-appropriate terminal sequence BEFORE `res.end()`, mirroring `emitStreamError` (src/stream-error.ts):
- OpenAI chat: optional marker delta + `finish_reason` chunk + `data: [DONE]`
- Anthropic: `message_delta` (stop_reason) + `message_stop`
- Responses: `response.failed` / `response.incomplete` + done

Open questions: (a) error framing vs silent-stop framing; (b) whether to inject a visible marker into the content so the model knows the prior turn was incomplete; (c) whether the zero-content retry from #413 should also apply to plugin pipes.

ranxianglei:
## Problem (#721)

In plugin mode (agent-owned compression: `bili dsh` / `bili omp` / ...), when the upstream model stream terminates abnormally mid-turn — clean EOF with no finish_reason/[DONE]/message_stop/done-family event, or an upstream read failure while the client is still connected — the proxy closed the response right after the last received chunk. The agent client received a bare cut-off SSE with NO completion or error event and persisted the partial turn as if it were complete (the chain that produced #719).

Two shapes, both verified on master f85553b:
- **Clean EOF without a terminal event**: `maybeNoteTruncated` only records server-side (`noteWeakOverflow`), then `finally { res.end() }` closes the stream. Nothing client-visible.
- **Upstream read throw** (RST/timeout/idle abort): the pipe rethrew; the plugin branch in `src/server.ts` has no catch; the top-level handler does a bare `res.end()` when headers are already sent.

## Fix

New `emitUpstreamTruncation(res, protocol, finished, log?)` in `src/stream-error.ts` (sibling of `emitStreamError`/`emitPreflightError`; best-effort, never throws, ends res). Both plugin pipes now call it before closing when `!sawTerminal` and the client is still connected:

- **Case A — finish reason delivered, final terminal byte lost** (OpenAI `finish_reason` chunk / Anthropic `message_delta.stop_reason` seen): synthesize ONLY the missing terminal byte (`data: [DONE]` / `event: message_stop`). Silent completion — the model genuinely finished.
- **Case B — true mid-generation cut** (no finish reason): protocol-native in-band error event, same shapes as `emitPreflightError` (#568): OpenAI top-level `error` data frame `{type:"server_error", code:"upstream_stream_truncated"}` + `[DONE]`; Anthropic/Responses `event: error`. Deliberately NOT a fabricated `response.failed` lifecycle object — the pipe does not track the full output-item list, and strict clients (codex) crash on half-consistent terminal payloads (see the `emitStreamError` comment). Responses always takes the error shape because its terminal events carry the status (no separate finish-reason concept).
- **Throw shape**: both pipes' `catch` now flushes held tag tails (partial prose is never silently lost) and emits the in-band signal instead of rethrowing into the top-level handler's bare close. Client-abort early-return unchanged. The stale rethrow-contract comment at the server.ts plugin branch is updated.

Free coverage: the #460 non-injected proxy-mode branches reuse these same pipes (`src/server.ts:3776`, `3800-3802`) and get the same protection; the `resetAfterSuccess` native-compaction branch now reaches its "rebase NOT scheduled" log path cleanly on truncation instead of throwing into the top level.

Semantics implemented per the recommendation posted in the #721 thread (error framing over silent-stop; no visible content marker). If a different semantic was intended, it is a small change confined to the single helper.

Known residual edge (out of scope): with the opt-in #371 fake-completion backstop enabled, plugin bodies buffer through `resolveFakeCompletion` first, and a read error there still escapes to the top-level bare close. Follow-up candidate: porting the #413 zero-content retry into the plugin pipes (needs re-fetch plumbing; separate PR per the thread discussion).

## Tests

Extended `tests/plugin-truncation-weak-overflow.test.ts` (+11 tests): per-protocol x per-case wire assertions (openai mid-cut → error frame + [DONE]; openai finish-seen → bare [DONE], no error; anthropic stop_reason-seen → message_stop synthesized; anthropic pre-stop cut → error event; responses mid-cut → error event), both throw shapes (mid-generation and post-finish-reason), client-abort suppression (both pipes), and the no-session #460 variant.

## Pre-flight

- `npm run typecheck` — clean
- `npm test` — 1382 pass / 0 fail / 2 skipped (e2e gate)
- `npm run build` — success; verified `upstream_stream_truncated` present in `dist/index.js`
- `E2E_CHECK=1` codex pre-flight — pass (codex binary present). Full `ACP_TEST_E2E` run not possible in this sandbox (no local upstream at 127.0.0.1:8199 — same constraint noted in #720); the changed paths are unit-covered per protocol x case, including both truncation shapes.

Pre-ship verification item from the issue thread: confirm each registered client's mid-stream error handling (pi / omp / opencode / dsh) against a real session before merge — dsh's behavior on OpenAI-wire error frames is the unknown one.

Fixes #721

### DIFF
--- a/src/exit-matrix.ts
+++ b/src/exit-matrix.ts
@@ -54,13 +54,13 @@ export const ERROR_DELIVERY: Record<WireExitId, ExitCell> = {
     },
     "plugin-chat-sse": {
         implementer: [pipePluginChatWithStrip],
-        contract: "byte-faithful passthrough: pre-stream fetch failure → JSON `{error: formatUpstreamError}` before any SSE byte; mid-stream cut → stream simply ends (no synthetic error event — the agent's tool loop must not see fabricated events)",
-        coveredBy: ["tests/issue411-abort-usage.test.ts", "tests/plugin-agent.test.ts"],
+        contract: "byte-faithful passthrough: pre-stream fetch failure → JSON `{error: formatUpstreamError}` before any SSE byte; upstream cut without terminal event → synthesized terminal byte when a finish reason was delivered, else protocol-native in-band error event (#721); client abort → clean end",
+        coveredBy: ["tests/issue411-abort-usage.test.ts", "tests/plugin-agent.test.ts", "tests/plugin-truncation-weak-overflow.test.ts"],
     },
     "plugin-responses-sse": {
         implementer: [pipePluginResponsesWithStrip],
         contract: "same as plugin-chat-sse, responses wire",
-        coveredBy: ["tests/issue411-abort-usage.test.ts"],
+        coveredBy: ["tests/issue411-abort-usage.test.ts", "tests/plugin-truncation-weak-overflow.test.ts"],
     },
     "plugin-json": {
         implementer: [pipePluginJson],
--- a/src/plugin.ts
+++ b/src/plugin.ts
@@ -901,7 +901,17 @@ export async function pipePluginChatWithStrip(
             }
             if (res.destroyed || res.writableEnded) break;
         }
-        if (buf.length > 0 && !res.destroyed && !res.writableEnded) await write(buf);
+        // #721: upstream EOF without a terminal event must not close the
+        // stream bare — the agent would persist the partial turn as complete
+        // (the #719 chain). finished=true when a finish reason was delivered:
+        // only the trailing terminal byte ([DONE]/message_stop) is missing.
+        const truncated = !sawTerminal && !res.destroyed && !res.writableEnded;
+        // A dangling partial event left in buf by a mid-event cut would fuse
+        // with the next complete frame — SSE joins every data line inside one
+        // blank-line-delimited block — corrupting the truncation signal. Drop
+        // it when the signal follows: an unterminated event is unparseable by
+        // the client anyway (same as the pre-#721 bare end).
+        if (!truncated && buf.length > 0 && !res.destroyed && !res.writableEnded) await write(buf);
         const rest = flushTails();
         if (rest.length > 0 && !res.destroyed && !res.writableEnded) await write(rest);
         // Settle BEFORE res.end() in the finally below: the client can issue
@@ -911,11 +921,7 @@ export async function pipePluginChatWithStrip(
         settleUsage();
         maybeNoteTruncated();
         maybeWarnDegenerate();
-        // #721: upstream EOF without a terminal event must not close the
-        // stream bare — the agent would persist the partial turn as complete
-        // (the #719 chain). finished=true when a finish reason was delivered:
-        // only the trailing terminal byte ([DONE]/message_stop) is missing.
-        if (!sawTerminal && !res.destroyed && !res.writableEnded) {
+        if (truncated) {
             emitUpstreamTruncation(res, protocol, finalFinishReason !== undefined, log);
             return;
         }
@@ -929,9 +935,10 @@ export async function pipePluginChatWithStrip(
         // #721: upstream read failed while the client is still connected —
         // deliver the in-band truncation signal instead of rethrowing into
         // the top-level handler, which would close the stream bare. Flush
-        // held tag tails first so partial prose is never silently lost.
+        // held tag tails first so partial prose is never silently lost. The
+        // dangling partial event left in buf is dropped (see EOF path above):
+        // written raw it would fuse with the signal frame.
         try {
-            if (buf.length > 0) await write(buf);
             const rest = flushTails();
             if (rest.length > 0) await write(rest);
         } catch {
--- a/src/plugin.ts
+++ b/src/plugin.ts
@@ -10,6 +10,7 @@ import { executeProxyTool } from "./loop/core.js";
 import { normalizeSseLineEndings } from "./sse-util.js";
 import { containsRenderTagText, createTagEchoFilter, mayStartRenderTag, stripAnthropicText, stripOpenaiChatText, stripResponsesText, type TagEchoFilter } from "./loop/tag-echo-filter.js";
 import { log as loggerLog } from "./logger.js";
+import { emitUpstreamTruncation } from "./stream-error.js";
 import { degenerateTurnWarning } from "./degenerate-turn.js";
 import { noteWeakOverflow } from "./weak-overflow.js";
 import { warnCacheCollapse } from "./cache-warn.js";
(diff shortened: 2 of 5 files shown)


ranxianglei:

[bot] 🏷 On it — reviewing PR #722 now. Duplicate screening done: searched open+closed issues for truncation/stream/terminal variants; this is the only open item on plugin-mode truncation-terminal synthesis. Nearest neighbors are distinct: #719/#720 (symptom 400, orthogonal wire fix), #570/#572 (weak-overflow self-heal, closed), #551 (upstream prefill timeout, closed). No other open PR touches these pipes. Next: line-by-line diff review of the 4 changed files (`src/plugin.ts`, `src/server.ts`, `src/stream-error.ts`, `tests/plugin-truncation-weak-overflow.test.ts`), both-compression-mode verification per AGENTS.md §6, then local typecheck/test/build to confirm the pre-flight claims. Will follow up with the full review verdict.

ranxianglei:

[bot] 🏷 Review of PR #722 complete. Found **one real bug** during review and fixed it directly on the PR branch (two follow-up commits pushed: `765c971`, `c1ea725`).

**Verified**
- Both modes (§6): plugin-mode pipes and the #460 non-injected proxy-mode branches (src/server.ts:3777, 3800-3803) get the protection; session-less calls keep usage accounting off. Non-injected clients receive each wire's documented in-stream error shape (OpenAI top-level `error` frame, Anthropic/Responses `event: error`) — strictly better than a bare close.
- Error frames mirror `emitPreflightError` (#568) exactly for all three protocols — the established precedent strict clients tolerate. Choosing the error channel over a fabricated `response.failed` is correct.
- A/B case split is right: finish reason delivered → synthesize only the missing terminal byte (`[DONE]` / `message_stop`); true mid-cut → in-band error. Responses always `finished=false` (terminal events carry the status).
- Client-abort suppression (`res.destroyed || res.writableEnded`) and the `settleUsage`/`maybeNoteTruncated` ordering are preserved — #411 usage landing and #570/#572 weak-overflow self-heal learning unaffected.
- `#411 upstream cut with client alive` (tests/issue411-abort-usage.test.ts:175) stays compatible: all three of its assertions hold under the new flow.
- The documented residual is accurate: `resolveFakeCompletion` (src/server.ts:4035-4090) returns the buffer unchanged and never synthesizes terminals → no double-signal; read errors during its buffering do still escape to the top-level bare close. Fine as a noted follow-up.

**Bug found & fixed — SSE fusion corrupts the signal (commit `765c971`)**
A cut landing mid-event leaves an unterminated partial event in the chat pipe's `buf`. Both new paths wrote that fragment raw *before* the synthesized signal — and SSE joins every `data:` line inside one blank-line-delimited block, so the fragment fused with the error frame into a single malformed event:

```
data: {...partial...}data: {"error":{"type":"server_error","code":"upstream_stream_truncated",...}}
```

The truncation signal was corrupted in exactly the dominant case (a mid-chunk TCP cut almost always leaves a partial line) — reintroducing, ironically, the half-consistent-payload failure mode this PR exists to remove. The existing tests couldn't catch it because their failing streams end on event boundaries (`buf` empty). Reproduced pre-fix on both the EOF and read-failure paths; post-fix every emitted frame is standalone-parseable. Fix: drop the dangling fragment when the signal follows (it is unparseable by the client anyway — same client-visible semantics as the pre-#721 bare end, and consistent with the responses pipe, which never writes raw `buf`). Two regression tests added asserting every emitted data payload parses standalone.

**Docs fix (commit `c1ea725`)**
`ERROR_DELIVERY["plugin-chat-sse"]` in src/exit-matrix.ts still claimed "mid-stream cut → stream simply ends (no synthetic error event)" — stale after this PR. Contract text refreshed and both plugin-SSE cells' `coveredBy` now point at the #721 test file.

**Validation**: typecheck clean; full suite 1384 pass / 0 fail / 2 skipped (e2e-gated); build success. Full E2E run not possible in this sandbox (no local upstream at 127.0.0.1:8199) — unit coverage follows the #720 precedent.

**Minor observation (non-blocking)**: the catch-path log line says "upstream stream read failed" even when the throw actually came from a client-side write error with the socket half-gone (race where `res.destroyed` hasn't flipped yet). Emission there is harmless best-effort; log wording only.

Verdict: design is sound in both modes, and the fusion bug is fixed on-branch. Ready for human review + merge (`mergeable_state: blocked` = awaiting approval, per branch protection): https://github.com/ranxianglei/billion-context/pull/722

中文摘要：审查发现并修复了一个真实 bug——流在事件中间被切断时，chat pipe 残留的半截 SSE 片段会被原样写出、与新合成的错误帧拼成同一个畸形事件块（SSE 会合并同一事件内的多个 data 行），使截断信号在最常见的切断形态下恰好失效；已改为在发信号前丢弃该片段并补了 2 个回归测试，另更新了 exit-matrix 里过时的契约描述；两种压缩模式、与 #568 错误帧形状的对齐、客户端中止抑制等均验证无误，可以合并。

PR #737, opened 2026-09-13
fix: name the operator-shrunk window in preflight fail-fast errors

### LINKED ISSUE #736: 窗口溢出报错
日志稍后

---
_Mirrored from ework issue #736_

ranxianglei:
## What

When preflight compression cannot bring an over-window payload under the window, the fail-fast error ends with "Raise the model context window or restart the session to recover." — which sends operators to the upstream when the actual wall is bili's own `compress.modelContextLimit` setting.

#736 is exactly that misdirection in the wild: the log shows `[window] source=table-or-registry native=128000 effective=16000` — the model's registered window is 128K and the local upstream demonstrably served it (an earlier turn of the same session reported 82022 input tokens; a sibling session was billed at 16589) — yet every fail-fast only ever said "raise the model context window", while the ceiling was a `compress.modelContextLimit: 16000`-style override edited into the web UI minutes before the brick.

This change threads the resolved native window + shrink reason (`operator` / `codex`) from the per-request window-resolution block into `preflightCompressIfNeeded`, and when `effective < native` the fail-fast message now adds:

- operator case: "Note: bili's effective window 16000 is below the model's full window 128000 — your compress.modelContextLimit setting overrides it; if the upstream actually serves the larger window, raise or remove that setting (hot-reloaded, no session restart needed)."
- codex case: the note names the codex align instead, with the explicit-opt-out pointer.

Also normalizes the doubled period at the detail seam (`...recover..` → `...recover.`), visible verbatim in the #736 log.

## Why not just docs

The message is the only surface an operator sees at the moment of failure; the #736 thread shows how far "raise the model context window" leads when the registry already knows the true window. The hint is emitted only when bili itself shrank the window below the resolved native value, so it can never point at a setting that isn't the cause (output-headroom reservation and self-heal paths leave `windowShrinkReason` unset → no note).

## Tests

New e2e in `tests/preflight-fail-fast.test.ts`: incompressible payload under an operator-shrunk 4k window (model declares 10k) → 502 message names both numbers and `compress.modelContextLimit`; negative control without the override asserts no note. Harness `startProxy` gained an optional compress-overrides param.

Pre-flight: `npm run typecheck` clean · `npm test` 1430 pass / 0 fail / 2 skipped (E2E-gated) · `npm run build` OK.

Fixes #736

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -3131,8 +3131,14 @@ async function preflightCompressIfNeeded(
             : "";
         // #736: when the wall is bili's own shrunken window, say so — "raise the
         // model context window" otherwise sends operators to the upstream when
-        // their compress.modelContextLimit is the actual ceiling.
-        const shrinkNote = resolvedNativeWindow !== undefined && limit < resolvedNativeWindow
+        // their compress.modelContextLimit is the actual ceiling. Gate on
+        // windowShrinkReason (set ONLY by the operator-override and codex-align
+        // paths) — NOT just on limit < resolvedNativeWindow: the per-request
+        // output-headroom reservation (reserveOutputHeadroom) also lowers
+        // reqConfig.modelContextLimit below native for every non-Anthropic turn
+        // with a max_tokens, so comparing alone would emit this note for a
+        // setting the operator never touched (#737 review).
+        const shrinkNote = windowShrinkReason !== undefined && resolvedNativeWindow !== undefined && limit < resolvedNativeWindow
             ? ` Note: bili's effective window ${limit} is below the model's full window ${resolvedNativeWindow} — ` +
                 (windowShrinkReason === "codex"
                     ? `it was aligned down to codex's own window perception; set compress.modelContextLimit explicitly if your upstream serves the larger window.`
--- a/src/server.ts
+++ b/src/server.ts
@@ -1086,6 +1086,12 @@ async function handle(
     // absolute number, a "70%" string of the native window, or unset → native)
     // overrides the table.
     let reqConfig = config;
+    // #736: the resolved NATIVE window (before the compress.modelContextLimit
+    // override and the codex align) plus what shrank the effective window below
+    // it — threaded into preflightCompressIfNeeded so a fail-fast can tell the
+    // operator that their own setting, not the upstream, is the wall they hit.
+    let resolvedNativeWindow: number | undefined;
+    let windowShrinkReason: "operator" | "codex" | undefined;
     // True when the resolved native window came from a low-confidence fallback
     // (built-in table / env default) instead of an authoritative source — such
     // windows get an effective-floor after output-headroom reservation (see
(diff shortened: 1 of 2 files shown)


ranxianglei:

Reviewed PR #737. Diff is clean (single commit, 2 files, focused). I found **one real correctness bug**, reproduced it, and fixed it directly on the branch — please merge the follow-up commit, not just the original.

## Bug: the shrink note fires for output-headroom reductions the operator never made

The note was gated on `limit < resolvedNativeWindow`:

```ts
const shrinkNote = resolvedNativeWindow !== undefined && limit < resolvedNativeWindow
    ? ` Note: bili's effective window ${limit} is below the model's full window ${resolvedNativeWindow} — ... your compress.modelContextLimit setting overrides it ...`
```

But `limit` (= `reqConfig.modelContextLimit`) is lowered below `native` by the **per-request output-headroom reservation** (`reserveOutputHeadroom = window - max_tokens`, applied at `src/server.ts:1548-1562`) for *every non-Anthropic turn with a max_tokens* — no operator setting involved. So comparing alone emits the "your `compress.modelContextLimit` setting overrides it" note for a reduction the operator never made. That contradicts the PR's own invariant: *"it can never point at a setting that isn't the cause (output-headroom reservation and self-heal paths leave `windowShrinkReason` unset → no note)."*

**Reproduced under the original gate** (OpenAI-chat endpoint, model declared 10k, no override, `max_tokens: 4000`, incompressible protected tool result):

```
context ~15032 tokens exceeds the model window 6000 (model=gpt-small) ... Raise the model context window or restart the session to recover.
Note: bili's effective window 6000 is below the model's full window 10000 — your compress.modelContextLimit setting overrides it; if the upstream actually serves the larger window, raise or remove that setting ...
```

The operator never set `compress.modelContextLimit`; the 10000→6000 drop is pure headroom reservation. The message sends them to edit a setting they never touched.

## Root cause / layer

The window can fall below native through several independent mechanisms. Only **two** mean "bili deliberately told you this is your wall": the operator override and the codex align — and those are exactly what `windowShrinkReason` records. Headroom reservation and self-heal are internal accounting, not an operator wall. The original code conflated "window is smaller than native" with "an operator/codex decision made it smaller."

## Fix (pushed — commit `2da365b`)

Gate the note on the deliberate-shrink signal instead of the raw comparison:

```ts
const shrinkNote = windowShrinkReason !== undefined && resolvedNativeWindow !== undefined && limit < resolvedNativeWindow
```

`windowShrinkReason` is set only by the operator-override path and the codex-align path, so the note now fires exactly when bili shrunk the window on purpose — matching the documented intent, in **both** compression modes (this shared `failFast` path serves proxy and plugin/launcher alike).

Added a regression test (`tests/preflight-fail-fast.test.ts`): an OpenAI-chat incompressible payload under a headroom-shrunk window asserts the reported window is `6000` (proves the headroom path ran) and that the message contains neither `full window` nor `compress.modelContextLimit`. The existing Anthropic-only #736 test couldn't catch this — Anthropic is exempt from headroom reservation, so its negative control never exercised the buggy branch.

## Pre-flight (on the fixed branch)

- `npm run typecheck` — clean
- `npm test` — 1431 pass / 0 fail / 2 skipped (E2E-gated)
- `npm run build` — OK
- Verified old gate reproduces the false note; new gate silences it.

Everything else in the original PR looks sound: `resolvedNativeWindow = native` is captured at the right point (fully-resolved native before any reduction), the codex/operator branch split is mutually exclusive, and the `detail.replace(/\.\s*$/, "")` period normalization correctly fixes the `...recover..` seam seen in the #736 log without over-stripping other `failFast` details.

---

中文摘要:修复了 #736 提示误报——原实现仅按"有效窗口 < 原生窗口"判断就提示"你的 compress.modelContextLimit 覆盖了窗口",但非 Anthropic 请求的输出预留(headroom)也会把窗口压低、与算子设置无关,导致误导用户去改一个没动过的配置;已改为仅在 bili 主动缩窗(operator/codex)时提示并补了 OpenAI 回归测试,已推到分支(commit 2da365b),可合并。

PR #763, opened 2026-09-14
fix: close the #762 strict-echo residual 400 class

### LINKED ISSUE #762: DeepSeek thinking mode: residual "reasoning_content must be passed back" 400 on the first post-fold request (0.1.109 / acp-kernel 0.0.69) — the forwarded body is not persisted
## Symptom

One request in a session was rejected by the upstream with

```json
{"error":{"message":"The `reasoning_content` in the thinking mode must be passed back to the API.","type":"invalid_request_error","param":null,"code":"invalid_request_error"}}
```

It happened **216 ms after a `compress` fold**, on the first request that saw the new projection. The next request of the same turn (client retry, 29 → 30 messages, `fold=new`, cache hit 64 %) was accepted, and the session then ran for many more turns. So this is a single-shot failure, not a wedged session.

This is the #684 signature (closed as fixed by acp-kernel #245/#247 plus the proxy gate, learn-on-400 and exit sentinels from #690), reproduced on 0.1.109 / acp-kernel 0.0.69. Your own review of #690 states the sentinels should be unreachable with the kernel gate in place, so this looks like a residual path.

## Measured evidence

From `$TMPDIR/bili-proxy-57852.log` (session `01a09d05-5b17-738d-996a-ac7f9aad1d85`, client omp 18.1.20, upstream `api.deepseek.com`, model `deepseek-v4-flash`, thinking enabled):

```
23:39:47.528  usage input=116041
23:39:47.557  compress requested 3 range(s)
23:39:47.612  processTurn 29 msgs
23:39:47.612  warning reasoning-pair-violated: 2 assistant tool-call message(s) lack
              reasoning_content while 4 carry it — strict-echo upstreams (DeepSeek
              thinking mode) will reject the request (#684)
23:39:48.302  ← upstream 400  (reasoning_content … must be passed back)
```

The client's pre-proxy body for that request is dumped at `~/.omp/logs/http-400-requests/1789342788308-2r0bbrc3az1ov.json` (`prompt_cache_key` identifies session `01a09d05…`, 424 messages, 6 assistant tool-call messages with `reasoning_content: \"\"`).

Two observations that matter:

- **The sentinel does not discriminate.** It kept firing on *every* later request of the same session with the identical `2 lack` count (carry count grew 4 → 20) and none of those requests failed. Its test is `typeof msg.reasoning_content === \"string\" && msg.reasoning_content.length > 0`, so it cannot tell a blank string (accepted by DeepSeek) from an absent field (rejected).
- **The warning is chronic on this host.** `reasoning-pair-violated` fired 38× for the affected session while it ran healthily for hours, and 26×, 96×, 101× in three other concurrent sessions, plus two `pfa-…` sessions. As it stands the warning cannot serve as an alarm.
- **Learn-on-400 did not fire** for this session: `metadata` in its persisted record holds only `pluginAgent`, `outputBudgetHighWater`, `effectiveContextLimit`, `lastTurnWindow`, `effectiveAbsorb`, `systemPromptTokens` — no `strictReasoningEcho`.

## Request

**1. Persist the forwarded body when the upstream returns 4xx.** The failing projection is the only datum that can explain this, and it is not retained anywhere: the client's dumps are pre-proxy, the session envelope stores the fold *state* rather than the rendered wire body, and the proxy logs the request only in summary form. A bounded dump (env-gated, size-capped, holding the forwarded `messages` array) under `$TMPDIR` or the instance state dir would make this class diagnosable from real data instead of requiring a token-billed replay.

**2. Consider normalizing instead of warning on strict-echo paths.** For a strict-echo session, an assistant message that carries `tool_calls` but no `reasoning_content` field can be sent with `reasoning_content: \"\"` — DeepSeek accepts the blank form. That turns the class into something the projection repairs rather than something the sentinel reports. Precedent: hermes-agent PR #15527 (referenced from openclaw#71455) injects the blank field both when creating such messages and when replaying an already-poisoned history, with a provider-detection helper scoped to DeepSeek/Kimi thinking mode.

**3. Optional:** if the intent stays warn-only, the sentinel could count *absent* rather than *blank* (`\"reasoning_content\" in msg` instead of `.length > 0`), which is the actual condition the provider rejects and would remove the chronic noise.

## Not claimed

I could not reproduce the failure on demand — it occurred once in ~180 requests, immediately after a fold, and a replay experiment would cost real tokens. This report deliberately does not claim a root cause inside the fold; it reports the correlation (post-fold request, 216 ms), the identical #684 signature, and the missing data needed to localize it.

ranxianglei:
## What

DeepSeek thinking mode accepts a **BLANK** `reasoning_content` echo but rejects an **ABSENT** field on assistant tool-call turns (`reasoning_content ... must be passed back`). The kernel round-trip drops blank echoes (an empty string carries no core message — `acp-kernel` wire codec only materializes non-empty reasoning), so any rebuild can ship absent fields into a thinking session. That is the residual 400 of #762.

This PR closes the rejection class on all four fronts requested in the issue:

1. **Normalize instead of warn (ask 2).** New `normalizeStrictEchoReasoning()` in `src/server.ts`: when strict-echo applies to the session (static `/deepseek/i` origin or learned flag), every assistant message carrying `tool_calls` but no `reasoning_content` field gets `""` injected before the request leaves. Blank form is what DeepSeek accepts — precedent hermes-agent PR #15527 (openclaw #71455). Runs on both compression modes (proxy + plugin); it touches neither tool calls nor summary carriers. Applied in `prepareOpenai` *before* the sentinel, so a repaired body never fires its own canary.
2. **Sentinel precision (ask 3).** `warnReasoningPairs` now counts **presence**, not non-emptiness. A blank echo is accepted by DeepSeek; counting it as absent made the sentinel fire on every turn of every healthy thinking session (38× in the reported session alone, plus 26/96/101× in three others).
3. **Persist the rejected body (ask 1).** New `src/error-dump.ts`: on upstream 4xx, the exact forwarded bytes are written to `<ACP_DUMP_DIR|stateDir/dumps>/err-<ts>-<sid>-<status>.json`. Env-gated (`BILI_DUMP_4XX=1`), size-capped (`BILI_DUMP_4XX_MAX_BYTES`, default 2 MiB, truncation marker), **off by default** (conversation bodies must not leak to disk uninvited — #276). Armed at BOTH failure sites: the `forward()` main path (client-originated requests) and the compress-loop re-request (`src/loop/core.ts`).
4. **Bonus root-cause fix for the missing flag.** The learn-on-400 learner existed ONLY in the compress loop — it never saw client-originated 400s. That is exactly why the reported session's persisted metadata lacked `strictReasoningEcho` even though its rejection was textbook strict-echo. The main path now learns the flag too (400 + `reasoning_content` in the error body), mirroring the loop's behavior.

## Tests

- Unit: normalization (patch scope, reference preservation, no mutation, log line), sentinel precision (blank counts present, absent still warns), dump gating/cap/sanitization (`tests/error-dump.test.ts`).
- E2E harness (`tests/reasoning-echo-400.test.ts`): real proxy + fake upstream; one 400 → main-path flag learned → second request's tool-call message arrives with a blank echo → sentinel fires exactly once (pre-repair only) → rejected body persisted byte-faithful (absent field visible in the dump) → dump stays off without the env gate.

Pre-flight: `npm run typecheck` clean · `npm test` 1476 pass / 0 fail / 2 skipped · `npm run build` green.

Fixes #762


ranxianglei:

[bot] 🏷 Reviewing PR #763 now — checking out `2026-09-14_762-strict-echo-residual`, running pre-flight (typecheck/test/build), and auditing the diff per review rules. Will report findings here shortly.

ranxianglei:

Reviewed the diff line-by-line, ran the full pre-flight, and stress-tested every claim. **Diff cleanliness: clean** — 6 files, +517/−3, every hunk relevant, no version bump, no drive-by changes. **Pre-flight verified independently:** `npm run typecheck` clean · `npm test` 1476 tests / 1474 pass / 0 fail / 2 skipped · `npm run build` green (dist/index.js 2.81 MB).

**The four fronts check out:**
1. *Normalization* — correct placement: `prepareOpenai` applies it to `rebuiltMessages` **before** `warnReasoningPairs`, covering both the kernel-rebuilt path and the catch-fallback raw-client path, and it's mode-independent (proxy + plugin). Same-array-ref return when nothing patched is a nice touch.
2. *Sentinel precision* — `hasRc` is now `typeof reasoning_content === "string"` (presence), matching DeepSeek's actual acceptance of blank echoes. Verified against the code, not just the description.
3. *Error dump* — gated/capped/sanitized as documented, off by default (#276 respected). I traced `wireBody` through its init/reassignment/forward/dump sites and confirmed the dumped bytes are exactly what went out the door. Both failure sites armed.
4. *Main-path learner* — mirrors the loop learner, and I verified the consumers (`stripReasoning`/`withReasoningDrop`) actually consult the learned flag in both paths, so learning changes behavior rather than just logging.

**However — one real coverage hole in the class this PR claims to close, now fixed directly on the branch (commit `8d84cb6`, pushed):**

The normalization lived only in `prepareOpenai`. But the **compress-loop re-request never goes through `prepareOpenai`**: proxy-mode streaming requests are wrapped in `runCompressLoop` whenever compression is injected, and the re-request body is built by `adapter.buildRequest(coreMessages, ...)` straight from the kernel view. Since the kernel round-trip drops blank echoes (the established codec behavior from #762 floor 3), a rebuilt re-request could still ship an **absent** `reasoning_content` on assistant tool-call turns into a strict-echo session — the identical DeepSeek 400 class. Learning alone doesn't close it: the loop learner sets the flag, but nothing in the loop normalizes bodies afterwards. Reachability: proxy mode (e.g. a launcher pointed at `api.deepseek.com`) any time the model issues a `compress` call or preflight fires. The reported incident itself was plugin-mode main path (which your PR does cover) — this is the residual's other door.

I proved it empirically before touching code: a unit-level probe drove `openaiToCore` → `coreToOpenai` → `createOpenaiAdapter(...).buildRequest(...)` and the outgoing wire carried the tool-call message with the field absent.

**The fix:**
- Moved `isStrictReasoningEcho` + `normalizeStrictEchoReasoning` verbatim into a new leaf module `src/strict-echo.ts` (necessary: `server.ts` imports `./loop/index.js`, so loop code cannot import back from `server.ts`; the module is type-only imports, zero runtime deps). `server.ts` re-exports both, so existing imports/tests are untouched.
- Added `normalizeStrictEchoBody()` (body-level twin; same-ref when disabled / no messages array / nothing to patch).
- Applied it in `runCompressLoop` immediately after `adapter.buildRequest(...)`, gated by the same predicate (learned flag on `ctx.session.metadata` OR static `/deepseek/i` origin derived from `requestOptions.url`), and placed **before** the `ACP_DUMP_BODY` dump so dumps show the repaired wire. Log lines carry the `[acp-loop]` tag so you can tell the two paths apart.

**New tests:** 4 unit cases for the body-level helper (disabled/no-op identity, key preservation, no input mutation, log line) plus `tests/loop-strict-echo.test.ts` — a real-proxy E2E mirroring your harness style, streaming variant: request 1 ships the absent field → 400 → main-path learner arms the flag → request 2 normalized (sentinel silent) → model issues a `compress` call (empty ranges → FAILED, history intact) → **re-request arrives with the blank echo on the split turn** → `[acp-loop]` repair logged → sentinel fired exactly once across the whole arc. One hermeticity note: the store is injected as `SessionStore({ enabled: false })` like your sibling harness — passing a string path silently falls back to the default state dir, which makes this kind of test stateful across runs.

**Post-fix pre-flight:** `npm run typecheck` clean · `npm test` 1481 tests / 1479 pass / 0 fail / 2 skipped (+5 new) · `npm run build` green. E2E verified stable across repeated runs.

One-line Chinese summary: 审查发现本 PR 的 strict-echo 修复只覆盖了 prepareOpenai 主路径，而压缩循环的重发请求绕过该路径、仍会把缺失的 reasoning_content 发往 DeepSeek（同一类 400），已在 PR 分支直接补上循环路径修复并新增单测+E2E（typecheck/test/build 全绿），可以合并。

PR #770, opened 2026-09-14
fix: normalize bili compaction markers in Codex native fallback

### LINKED ISSUE #769: Codex native compaction fallback forwards private bili markers and fails with HTTP 400
A native Codex compaction fallback can forward bili's local compaction item to the upstream, causing HTTP 400 and preventing compaction from completing. This was observed in an automatic approval-review session with billion-context 0.1.109 on Windows.

The upstream error was (identifier sanitized):

```
Invalid 'input[6].id': 'fc_bili_<uuid>'. Expected an ID that begins with 'cmp'.
```

The proxy logged `codex compaction_trigger request not intercepted (gate preconditions not met) — forwarding verbatim (no preflight, no rebuild, no window clamp)` immediately before that response.

### Reproduction

1. Create a local forged compaction response using `buildTriggerForgeBody()`. Its item has an `fc_bili_` ID and `bili:acp:` summary.
2. Echo that item in a Codex Responses request, followed by a final `compaction_trigger`.
3. Let the native fallback run, either because interception gate preconditions fail or because `BILI_CODEX_COMPACT=pass`.
4. An upstream that validates native compaction IDs rejects the local item.

A focused HTTP regression reproduces the failure at `input[6]` in both modes without credentials or an external model.

### Root cause and proposed fix

The trigger fallback in `src/server.ts` forwards the original request buffer before `prepareResponses()`, bypassing its `replaceBiliCompactionItems()` handling. Ordinary turns already convert the local item into a normal user message carrying the summary.

Apply that existing conversion only to the native trigger fallback's original wire input. Keep the original Buffer when no local items are present, preserve real `cmp_*` items and their opaque content, retain the final trigger, and avoid preflight or rebuilding the client's history. Parse the original buffer rather than reusing `parsed`, because a failed prepare/forge attempt may already have mutated it.

Renaming `fc_bili_` to `cmp_` would be incorrect: bili's plaintext summary is not an upstream-native encrypted compaction record.

### Local verification

The two new strict-upstream regressions fail with HTTP 400 before the fix and pass after it. Existing byte-identical fallback and native-compaction completion/state tests pass; TypeScript checking and the bundle build pass. A fix PR will follow.

This issue concerns the confirmed trigger fallback. It does not claim that earlier capacity/time-out failures have the same cause or that changing approval policy is necessary.

z0d0dOr:
Fixes #769

Codex can echo a bili-forged `fc_bili_*` compaction item in a later native compaction request. When interception falls back to the upstream, the raw item currently bypasses normal summary conversion and fails native ID validation with HTTP 400.

Normalize only bili compaction items in the original fallback buffer using `replaceBiliCompactionItems()`. Preserve the summary at its original position, native compaction blobs, the final trigger and other request fields. Requests without bili items remain byte-identical. Reading the original buffer also avoids forwarding unrelated mutations left by a failed prepare/forge attempt.

Validation:

- Two strict HTTP upstream regressions reproduce the original `input[6].id` failure before the change and pass afterward, covering gate failure and pass mode.
- `node --import tsx --test tests/codex-compact-e2e.test.ts tests/native-compact-terminal.test.ts`: 13 passed, including existing raw-buffer and state/completion invariants.
- `npm run typecheck`, `npm run build`, and `git diff --check`: passed.

No new dependencies or version change. The real-client/four-phase upstream E2E required before merging has not been run; this PR's runtime evidence uses the existing loopback HTTP harness. The running local proxy and approval settings were not changed.


ranxianglei:

[bot] 🏷 Review in progress — initial findings:

**Bug confirmed (pre-fix repro).** I applied only this PR's new tests onto `origin/master`: both new regression tests fail with exactly the #769 error — HTTP 400 `Invalid 'input[6].id': 'fc_bili_…'. Expected an ID that begins with 'cmp'.` — while all 10 pre-existing codex-compact-e2e tests still pass. On the PR branch the same file passes 13/13. So the new tests are a genuine regression of the reported failure, not a contrivance.

**Approach verified.**
- Reusing `replaceBiliCompactionItems()` — the same function the normal prepare path already uses at `src/server.ts:2385` — scoped to the trigger-form fallback is correct: native `cmp_*` blobs, the final `compaction_trigger`, and all other request fields are preserved; bili items are replaced **in place** with a user message carrying the summary (so prefix context survives the native compaction instead of being silently dropped); requests without bili items stay byte-identical (`forwardBody === bodyBuffer`).
- Parsing from `bodyBuffer` instead of reusing `parsed` is correct: `bodyBuffer` is never mutated after decode (`src/server.ts:972`/`:1011`), while a failed `runPrepare()` may have mutated `parsed` (e.g. `src/server.ts:2389`). The re-parse also cannot throw — the identical buffer was already successfully parsed at `src/server.ts:1059` before this point.
- Endpoint form (`/responses/compact`) is unaffected: `prepareResponsesCompact` already strips bili items on every forwarding path (`src/server.ts:2685`), so the trigger-form fallback was the only gap.
- Compression modes: this path is codex-wire specific (plain codex through the proxy, i.e. proxy mode); plugin-mode agents don't emit `compaction_trigger` items, so they're untouched.

**Checks so far (PR branch):** targeted suite 13/13 ✓ · `npm run typecheck` ✓ · `npm run build` ✓ · `git diff --check` ✓. Diff is minimal (2 files, +54/−6), no version change, no new dependencies.

Full suite + real-client E2E status to follow shortly.

ranxianglei:

[bot] 🏷 Review complete — **LGTM, with one pre-merge gate outstanding** (see below).

**Verification results** (all on branch `pr-770`, commit `a3d30e3`):

| Check | Result |
|---|---|
| Pre-fix repro (new tests applied to `origin/master`) | ✅ both new tests fail with exactly `Invalid 'input[6].id': 'fc_bili_…'. Expected an ID that begins with 'cmp'.` (HTTP 400); all 10 pre-existing e2e tests still pass |
| Targeted suite post-fix (`codex-compact-e2e` + `native-compact-terminal`) | ✅ 13/13 |
| Full `npm test` | ✅ 1461 pass / 0 fail / 2 skipped (1463 total) |
| `npm run typecheck` · `npm run build` · `git diff --check` | ✅ all clean |

**Code review.** The change is correctly minimal and correctly scoped:

- Reusing `replaceBiliCompactionItems()` gives automatic parity with the normal prepare path (`src/server.ts:2385`) — no new conversion semantics to diverge. In-place replacement (not stripping) is the right choice for the trigger form specifically: after a prior forge, codex's local history is `[fc_bili_item, tail…]`, so the item *is* the prefix record — dropping it would make the upstream's native compaction lose that context, while replacing it with the summary-bearing user message lets the prefix survive (this matches the documented rationale at `src/codex-compact.ts:65-71`). The endpoint form needed nothing: `prepareResponsesCompact` already strips bili items on every forwarding path (`src/server.ts:2685`), confirming the trigger fallback was the only gap.
- Reading from `bodyBuffer` (not `parsed`) is correct and safe: the buffer is never mutated after decode (`src/server.ts:972`/`:1011`), a failed `runPrepare()` can mutate `parsed` (`src/server.ts:2389`), and the re-parse cannot throw because the identical buffer already parsed successfully at `src/server.ts:1059`.
- Byte-identity guarantee holds: when no bili items are present, `forwardBody === bodyBuffer` (the original buffer object, untouched). When present, `{...original, input: items}` preserves key order and every other field, including native `cmp_*` blobs and the final trigger.
- Both compression modes considered: the path requires codex UA + `compaction_trigger`, i.e. plain codex through the proxy (proxy mode); plugin-mode agents never emit this shape.
- Diff cleanliness: 2 files, +54/−6, every line on-topic, no whitespace churn, no version bump, no new dependencies. Not a duplicate — this PR is the direct fix for #769.

**Outstanding before merge:** AGENTS.md requires the real-client four-phase E2E (`tests/e2e/e2e-codex.test.ts`, `ACP_TEST_E2E=1`) before merging request-pipeline changes like this one. It cannot be run in my sandbox — the codex binary (0.147.0) and built dist are present, but no Responses-compatible upstream is reachable here (zero-token preflight confirms). Please run it locally against your upstream (or via the manual-dispatch CI workflow with its secrets), then merge. Nothing else blocks.

中文摘要：修复了 Codex 原生压缩回退路径把 bili 私有 `fc_bili_*` 压缩项原样转发给上游导致严格后端 400（#769）的问题——在原始请求缓冲上复用现有 `replaceBiliCompactionItems()` 做原位归一化，已验证修复前可精确复现原报错、修复后 13/13 定向测试与全套 1461 项测试通过，diff 干净最小；唯一待办是合并前补跑一次真实客户端四阶段 E2E（本沙箱无可用上游，需本地或 CI 执行），跑完即可合并。

Issue #774, opened 2026-09-14
Context-window lookup misses provider-prefixed model ids (vLLM/relay 'prefix/name') — stale static table wins over fresh models.dev data


ranxianglei:
来源: #736 (https://github.com/ranxianglei/billion-context/issues/736) 分析窗口溢出问题时发现

## Symptom (from #736 user log, bili v0.1.107, local vLLM)

```
[acp-registry] loaded models.dev (395 models, direct)
[window] model=qwen/qwen3.8-27b source=table-or-registry native=128000 effective=16000 launcher=none configured=none peek=none fallback=false
```

The models.dev registry loaded FRESH (395 models, fetched live) yet `peek=none`: the request's model id `qwen/qwen3.8-27b` (vLLM serves HF-style `org/model` names) never matches the registry entry `alibaba/qwen3.8-27b`. The resolved `native=128000` therefore came from the built-in static table (`/^qwen/i → 128_000`) — stale against models.dev's 262144 for the same model.

## Root cause

1. `registryLookup()` (src/registry.ts) builds candidates as: exact key, then keys ending in `/` + FULL name (cross-provider scan). For `qwen/qwen3.8-27b` the scan looks for keys ending in `/qwen/qwen3.8-27b`, which never exists — the BARE name `qwen3.8-27b` is never tried.
2. The server.ts resolution order (beta → plugin → launcher → configured → peek → **table** → async registry only when `!native`) means once the static table hits, the async registry is never consulted — a fresh registry can never correct a stale table value for prefixed ids.

## Impact

Any relay/vLLM deployment serving models under non-vendor prefixes gets the stale static-table window forever:
- under-reported window → preflight compression + nudge bands fire early (wasted summarization calls);
- over-reported window → guaranteed upstream 400s until the learned-limit self-heal pays for itself with failed turns.

## Suggested fix (implemented, linked PR)

Try the bare basename after the full name in both `lookupContextLimit()` and `registryLookup()`; an exact listed prefixed id still outranks the bare fallback.

ranxianglei:
> 🤖 Powered by [ework](https://github.com/ranxianglei/ework-aio) · qwen3.8-27b

[bot] 🏷 Triage done — root cause confirmed in code, implementing the suggested fix now.

**Verified:**
[...]

PR #775, opened 2026-09-14
fix: resolve context windows for provider-prefixed model ids

ranxianglei:
## What

Relay/vLLM deployments serve models as `prefix/name` (e.g. vLLM's `qwen/qwen3.8-27b`) while models.dev and the built-in table key by vendor id (`alibaba/qwen3.8-27b`, `/^qwen/i`). Neither lookup ever tried the bare basename, so such ids silently fell through to the stale static table — even with a fresh models.dev snapshot resident.

Log evidence (#736, same minute):
```
[acp-registry] loaded models.dev (395 models, direct)
[window] model=qwen/qwen3.8-27b source=table-or-registry native=128000 ... peek=none fallback=false
```
Registry had the model at 262144; the user got the table's 128000.

## Fix

- `src/config.ts lookupContextLimit()`: try roots `[full name, bare basename]` against the table patterns.
- `src/registry.ts registryLookup()`: build candidate names as full-name variants first, then bare-name variants (exact-key and cross-provider suffix scan each run per name, order preserved) — so a genuinely listed prefixed id still outranks the bare fallback.

With this, #736's scenario resolves `native=262144` straight from the registry/snapshot instead of the stale table.

## Tests

- `tests/routing.test.ts`: table matches `meta-llama/Llama-4-Maverick`, `qwen/qwen3.8-27b`; unknown prefixed ids still undefined.
- `tests/registry.test.ts`: exact #736 shape (`alibaba/qwen3.8-27b` listed, `qwen/qwen3.8-27b` requested) resolves via bare basename, with and without host; listed prefixed id outranks bare-scan fallback; bundled snapshot resolves the #736 name offline.

Pre-flight: `npm run typecheck` clean · `npm test` 1493 pass / 0 fail / 2 skipped (E2E-gated) · `npm run build` OK.

Fixes 774


ranxianglei:

[bot] 🏷 Review complete — bug verified at code level, diff is clean, pre-flight independently reproduced. No issues found; ready to merge (merge stays human-only per repo rules).

**Triage (bug class — verified before endorsing)**
1. Reproduced the mechanism in code: for `qwen/qwen3.8-27b`, old `registryLookup()` (src/registry.ts:289) tried only the exact key plus keys ending in `/qwen/qwen3.8-27b` — the listed `alibaba/qwen3.8-27b` (262144) was unreachable, so `peek=none` and the table's `/^qwen/i → 128_000` won, exactly matching the #736 log line. ✔
2. Layering: the reported symptom *is* the root cause (neither lookup ever tried the bare basename) — not a surface effect of something deeper. No server.ts reorder needed: peek (registry, pre-warmed from the bundled snapshot) already outranks the table in the resolution chain, so making peek resolve the bare name fixes it end-to-end.
3. Fix evaluation: full-name-first candidate order makes the change **purely additive** — any input that matched before returns the identical value; only previously-unmatched prefixed ids gain a hit. A genuinely listed prefixed id keeps precedence over the bare fallback (pinned by the "outranks" test). Consistent with the existing cross-provider max-scan semantics.

**Diff cleanliness**: 4 files, +62/−3, every line on-topic (two lookups + their tests). Single conventional commit on current master tip, no lockfile drift, no version touch, no stray files.

**Pre-flight (run independently on `pr-775`)**
- `npm run typecheck` — clean
- `npm test` — 1493 pass / 0 fail / 2 skipped (E2E-gated)
- `npm run build` — OK
- Functional spot-check against the real bundled snapshot: `bundledSnapshotLookup("qwen/qwen3.8-27b")` → **262144** (exact #736 shape, offline), `lookupContextLimit("meta-llama/Llama-4-Maverick")` → 128000, `unknown-org/unknown-model` → undefined.

**Minor observations (non-blocking, report-only)**
- Multi-segment ids (`a/b/c`) only try the final segment as root, not intermediate `b/c`. Within the stated scope and vanishingly rare; fine as-is.
- Known-provider hosts give the bare basename its exact-key chance but no cross-provider scan — consistent with the documented "known providers don't cross-scan" policy, just noting it.

**Duplicate screening**: no other open issue/PR covers this fix; `Fixes 774` linkage is correct (#774 itself carries the required source marker back to #736).

修复了 relay/vLLM 前缀模型 id（如 `qwen/qwen3.8-27b`）在静态表和 models.dev 注册表中均查不到、被陈旧表值（128K）永久遮蔽的问题——两处 lookup 增加裸 basename 回退且全名优先，行为纯增量；diff 干净、预检独立复现全过（typecheck 干净 / 1493 测试通过 / 构建成功），可以合并。

PR #776, opened 2026-09-14
fix: resolve context windows for relay 'org/model' ids via bare basename (#774)

### LINKED ISSUE #774: Context-window lookup misses provider-prefixed model ids (vLLM/relay 'prefix/name') — stale static table wins over fresh models.dev data

ranxianglei:
## What

Relay/vLLM deployments serve HF-style `org/model` ids (e.g. `qwen/qwen3.8-27b`) that never matched the models.dev registry key (`alibaba/qwen3.8-27b`), so window resolution fell through to the stale static table — and once the table hit, the async registry was never consulted.

Two lookup sites now retry the **bare basename** after the full id:

- `registryLookup()` (src/registry.ts): per variant name, after the full-name exact keys + cross-provider scan, retry the same candidate logic on the basename (`qwen/qwen3.8-27b` → scan `*/qwen3.8-27b` → hits `alibaba/qwen3.8-27b`). An exactly-listed prefixed id still outranks the bare fallback (full id tried first). Known-provider hosts keep their no-cross-scan invariant for the bare name too.
- `lookupContextLimit()` (src/config.ts): mirror bug — table rows are `^`-anchored on bare names, so `alibaba/qwen3.8-27b` matched NO row while its basename would. Same full-id-first, then-basename order.

No server.ts reorder needed: the registry cache is pre-warmed from the bundled snapshot, so a resolving `peekRegistryContext` already outranks the table on every request (peek sits above table in the resolution chain at src/server.ts:1134-1150).

## Why the old behavior was wrong

For `qwen/qwen3.8-27b`, the cross-provider scan looked for keys ending in `/qwen/qwen3.8-27b` — which never exists; the BARE name was simply never tried. Result from #736's user log (v0.1.107, local vLLM):

```
[window] model=qwen/qwen3.8-27b source=table-or-registry native=128000 effective=16000 launcher=none configured=none peek=none fallback=true
```

## Verification

- `npm run typecheck` clean; `npm test`: 1435 pass / 0 fail / 2 skipped (e2e-gated); `npm run build` OK.
- New unit tests: prefixed id resolves via bare basename (relay host, no host), exactly-listed prefixed id outranks bare fallback, `-thinking` variant strips before the bare fallback, known-provider host still does not cross-scan the bare name, bundled snapshot resolves the exact #736 id offline; config-table tests for the mirror bug (incl. full-id-outranks-basename and unknown-family-still-misses).
- End-to-end against the real server path with the bundled snapshot (no registry seeding, no configured/plugin/launcher signals) — before/after for the exact #736 scenario:

```
# before (user log, v0.1.107)
[window] model=qwen/qwen3.8-27b source=table-or-registry native=128000 effective=16000 ... peek=none fallback=true
# after (this branch)
[window] model=qwen/qwen3.8-27b source=registry-peek native=262144 effective=262144 ... peek=262144 fallback=false
```

Fixes #774


ranxianglei:

[bot] 🏷 **PR #776 review — logic correct & green, but it's a duplicate of open PR #775; please consolidate before merging either.**

## ✅ What I verified (against current master)

I merged #776 onto `origin/master` in a scratch checkout and ran the full gate:

- **typecheck**: clean (`tsc --noEmit`)
- **tests**: `1495 pass / 0 fail / 2 skipped` (the 2 skips are the e2e-gated suite). Your body said 1435 — that was your older base; master has simply accumulated ~60 more tests since. No failures.
- **build**: OK.
- **traced every new test against the actual code** — all pass and encode the real behavior:
  - relay id resolves via bare basename (`qwen/qwen3.8-27b` → cross-scan `/qwen3.8-27b` → `alibaba/qwen3.8-27b`), both with a relay host and with no host.
  - exactly-listed prefixed id outranks the bare fallback (full id tried first).
  - `-thinking` variant strips at the outer `modelVariants` level *before* the bare fallback kicks in.
  - known-provider host still does **not** cross-scan the bare name (your `-custom` fixture correctly prevents a false positive from `minimax/qwen3.8-27b-custom`).
  - bundled snapshot resolves the exact #736 id offline.
  - config-table mirror: full-id-outranks-basename (`claude-anything/x`→200k) and unknown-family-still-misses.
- **confirmed the "no server.ts reorder needed" claim.** The resolution chain at `src/server.ts:1149-1154` is `betaWindow ?? pluginWindow ?? launcherWindow ?? configuredWindow ?? peekWindow ?? lookupContextLimit(...)` — the registry peek sits *above* the static table, so a now-resolving `peekRegistryContext` wins on every request. And `nativeFromFallback` clears whenever `peekWindow` is set (`src/server.ts:1161`), which is exactly why the log flips to `source=registry-peek … peek=262144 fallback=false`. The runtime cache is pre-warmed from the bundled snapshot at module load (`src/registry.ts:50-53`), so the "after" state is reachable fully offline — not just a unit-test artifact. The before/after logs are internally consistent with this code path.
- **diff is clean**: single commit, 4 files, all in-scope, no version bump, no lockfile churn. It auto-merges onto current master without conflict (only `src/config.ts` overlaps, and it resolves).

Minor, non-blocking notes on #776 itself: for relay hosts the cross-provider scan now runs up to 2× per variant (once for the full name, once for the basename) — bounded and negligible. The nested restructure re-indents the whole existing cross-scan block into an inner loop (churn with no functional change), which widens the review surface a bit.

## ⚠️ Blocking finding: #776 duplicates open PR #775

There is **another open PR solving the exact same bug (#774 / #736)** with the same bare-basename approach and the same observable behavior:

| | **PR #776** (this one) | **PR #775** |
|---|---|---|
| Branch | `2026-09-14_fix-774-prefixed-model-id-window` | `2026-09-13_registry-prefixed-model-names` |
| Base | older master `564bf93` (needs rebase) | **current master `4ac36e7`** (no rebase) |
| `src/registry.ts` | +79/−36 (wraps cross-scan in a nested `[name, base]` loop) | +18/−1 (builds `roots=[model, base]`, flattens+dedupes variants, leaves the existing cross-scan block untouched) |
| `src/config.ts` | +13/−2 | +12/−2 |
| New tests (registry+routing) | 43 | 41 |
| Verified here | typecheck ✓ · full suite 1495/0 ✓ · build ✓ | typecheck ✓ · targeted 41/41 ✓ |

Both try the same candidate set `{full-id variants} ∪ {bare-name variants}` and return on first hit, so they are **behaviorally identical** — only structure differs. They cannot both merge (they rewrite the same lines of `registryLookup`/`lookupContextLimit`).

## Recommendation

Consolidate to **one**. My lean: **keep #775 as the base and port #776's 2 extra tests into it**, then close the other as `重复`. Rationale: #775 is already on current master (zero rebase/conflict risk) and its minimal diff leaves the proven cross-provider scan byte-for-byte intact — lowest regression surface for a production proxy. #776's only advantage is 2 extra tests, which are trivially portable (in particular the known-provider-bare-basename invariant test with the `-custom` fixture and the offline bundled-snapshot test). If you'd rather keep #776 instead, just say so and I'll rebase it onto current master and drop #775.

I have **not** merged or closed anything — pick the winner and I'll execute the port/close. Please don't merge #776 independently while #775 stays open; they'll conflict.

---

中文摘要：审核了 #776（relay `org/model` 上下文窗口按裸名回退修复）——逻辑与测试均正确、可在当前 master 干净合并且全绿；但发现它与未合并的 PR #775 是**同一问题（#774/#736）的重复实现**，行为一致、仅代码结构不同，二者不能同时合并。建议二选一：倾向保留更干净且已在最新 master 上的 #775，把 #776 多出的两个测试移植过去后关闭另一个；请先处理重复再合并。

PR #727, opened 2026-09-12
fix: recover preflight from empty SSE summaries + dead-end cooldown (#726)

ranxianglei:
Fixes #726

## Problem

After the #626 (stream) and #663 (max_output_tokens) compatibility retries, a ChatGPT-backend summarization call can still return **HTTP 200 SSE with zero extractable summary text** — an in-stream error event or truncated stream that `extractSummaryFromSse` silently discarded. The only trace was `summary too short (0 chars)`, one viable range got skipped whole, and every codex auto-retry re-ran the identical doomed walk (~8 large summary calls per retry), producing the stable `502 retryable=false` loop from #726.

## Changes

**`src/preflight.ts`**

1. **Diagnose empty summaries** — new `diagnoseEmptySummary(text, json?)` scans a no-text body for terminal error signals: SSE `error` / `response.failed` / `response.error` events (`response.error.code` + `message`), `response.incomplete` (status + optional message), bare JSON `{error: {...}}`, plus fallbacks for empty bodies, event-carrying streams without text, and non-SSE bodies. The diagnosis goes into the warn log and is threaded into the fail-fast 502 message (`Last unusable summary: …`), so the operator/client finally sees WHY the summary was empty.
2. **Shrink-and-retry** — the chunk pass became a worklist: a chunk whose summary comes back unusable is halved (oldest half first) and retried down to a floor (`2×MIN_CHUNK_TOKENS` tokens / `2×minCompressRange` chars) before the range is skipped. This recovers size-driven rejections — the prime suspect in #726, where the first chunk was `CHUNK_FRACTION × window` ≈ 630K tokens of summarization input — while staying bounded by the existing `MAX_SUMMARY_CALLS_PER_PREFLIGHT` budget. Skip-on-exhaustion semantics for genuinely unsalvageable ranges are preserved.

**`src/server.ts`**

3. **Dead-end cooldown** — when preflight fails with `compressedRanges === 0` (zero progress under unchanged state = deterministic dead-end), the failure fingerprint (`model\u0000limit`) + status + full message is cached in `session.metadata.preflightDeadEnd` for a cooldown (default 5 min, `BILI_PREFLIGHT_DEAD_END_COOLDOWN_MS`). Matching requests fail fast with the cached diagnosis and **zero upstream calls**; any non-failure outcome clears the marker; aborted failures never arm it. The 502 message gains a recovery hint (restart the session). This stops the quota-burning retry loop from the issue's logs (8 identical walks in 48s).

[...]

PR #784, opened 2026-09-14
fix: preflight SSE extraction rejects truncated and failure streams (#780)

### LINKED ISSUE #780: [Bug] preflight 摘要抽取层：被截断的 SSE 会被当成完整摘要，delta-less 终端会被当成空摘要（#726 的恢复链路仍未闭环）
## 背景

#726（preflight SSE 摘要为空 → 持续 502）已由 PR #727 关闭，其修复是**诊断 + 缩小重试 + 死锁冷却**（`diagnoseEmptySummary` / `src/server.ts` / `tests/preflight-empty-summary.test.ts`）。

但决定「这段文本是不是一份可用摘要」的**抽取函数本身没有改**，仍是朴素版。

## 当前实现（v0.1.111 / `bfb0b68`，`src/preflight.ts:311` `extractSummaryFromSse`）

```ts
for (const line of text.split("\n")) {
    if (!line.startsWith("data:")) continue;
    ...
    if (protocol === "anthropic") { if (o.type === "content_block_delta") { /* d.type === "text_delta" */ } }
    else if (protocol === "openai") { /* choices[0].delta.content */ }
    else { if (o.type === "response.output_text.delta") out += o.delta;
           else if (o.type === "response.completed" && ...) { const full = extractSummaryText(...); if (full) return full; } }
}
return out;
```

即：只认 `data:` 行；每协议只认单一 delta 类型；循环结束**无条件** `return out`。缺少四件事：

1. **没有 SSE 帧完整性检查。** 网关把流截断（半行 `data: {...`、JSON.parse 失败被 `continue` 跳过、缺尾部空行）时，前面已累积的文本会被当成**完整摘要**返回 → 一份被截断的 tier-1 摘要被持久化进会话。这比「空摘要」危险得多：空摘要会走 #727 的诊断/重试链路并报错，截断摘要看起来完全正常，只是永久少了一段内容。#764（`QODER_MODEL_TRANSPORT=http` 下网关输出损坏 SSE：无 `[DONE]`、`data:` 行为半截 JSON、偶发缺 `finish_reason`）证明这种流在真实网关里存在。
2. **没有 `event:` 帧头类型兜底。** 当 `data:` payload 里没有可用的 `type` 字段时无法判型。仓库其它地方都有这个写法（`src/loop/adapter-anthropic.ts:54`、`src/loop/adapter-responses.ts:177`、`src/stream-terminal.ts:18` 的 `l.startsWith("event:")`），只有 preflight 没有。
3. **没有 delta-less 终端兜底。** 只发 `response.output_text.done`（文本在 `text`）或 `response.output_item.done`（文本在 `item` 内）而不发 delta 的上游，抽取结果恒为 `""` → 每次摘要都空 → 触发 fail-fast/降级，看起来像「上游不返回摘要」。
4. **没有失败终态判定。** `response.incomplete` / `response.failed` / `response.error` 不被视为无效；只要失败前吐过一点文本，那点文本就会被当成摘要接受。

## 本地加固（自托管实例自 2026-09-13 起在跑，已验证）

* `const properlyFramed = text.endsWith("\n\n");`（responses 协议要求流以完整帧收尾）
* 解析循环先取帧头：`if (line.startsWith("event:")) { eventType = line.slice(6).trim(); continue; }`，判型改为 `const type = typeof o.type === "string" ? o.type : eventType;`
* `response.output_text.done` → `terminalText = o.text`；`response.output_item.done` → `extractSummaryText(protocol, { output: [o.item] })` → `terminalText`
* `response.completed` 不再立即 `return full`，改为置 `completed = true` 并记录 `terminalText`（与 delta 文本二者取一，非空优先）
* `response.incomplete | response.failed | response.error` → `invalidTerminal = true`
* 收尾：

  ```ts
  if (protocol === "responses") return completed && properlyFramed && !invalidTerminal ? out || terminalText : "";
  return !invalidTerminal ? out || terminalText : "";
  ```

效果：被截断的流不再产出「半份但看起来正常」的摘要，而是返回 `""`，从而进入 #727 已经建好的诊断 + 缩小重试 + 冷却链路——#726 的恢复链路这才真正闭环。

## 建议

按上面 4 点加固抽取层（约 30 行，无协议语义变更：完整流的行为完全不变），`""` 继续交给 `diagnoseEmptySummary` 处理。

`Refs #726`、`Refs #764`

ranxianglei:
## What

Harden the preflight summary extraction layer (`extractSummaryFromSse`, `src/preflight.ts`) per #780. The naive accumulator conflated *"the stream delivered a complete summary"* with *"the stream died mid-delivery"*: a gateway truncation (#764 shape — half-line `data:` JSON, no final event) left a partial accumulator that was **persisted as a complete tier-1 summary**. That is silently worse than an empty summary, because #727's diagnosis + halving + cooldown chain only fires on empty results — a truncated summary looks normal and permanently loses content.

## Changes (`src/preflight.ts`)

1. **Frame integrity.** A `data:` line that fails to parse is corruption (`badFrame`), not noise to skip. Anthropic/openai bodies must end on a frame boundary (`\n\n`, CRLF-tolerant). Deliberately NOT requiring `finish_reason`/`[DONE]`/`message_stop` — #764 proves real gateways omit these occasionally; requiring them would turn one corrupt-gateway class into systemic 502s.
2. **`event:` frame headers** type frames whose payload lacks a `type` field, matching the convention already used in `src/loop/adapter-anthropic.ts`, `src/loop/adapter-responses.ts`, `src/stream-terminal.ts`. Two fixes over the local patch from the issue: the header value is reset at every frame boundary (SSE semantics — a sticky value leaks across frames), and precedence on completion is `terminalText || out` (the completed event's full response object stays authoritative over delta accumulation, as before).
3. **Delta-less terminals.** `response.output_text.done` (text in `text`) and `response.output_item.done` (text inside `item`) now carry their text instead of extracting `""` forever.
4. **Failure terminals invalidate.** `response.incomplete | response.failed | response.error`, generic `error` events, and bare `{error}` objects mark the result invalid — text emitted before the failure is not accepted as a summary.
5. **Responses requires `response.completed`** (spec-mandatory terminal). Once seen it is trusted as-is — no framing check stacked on top, so gateways that close right after the final event without a trailing blank line are safe.

Rejected streams return `""` and flow into the existing `diagnoseEmptySummary` + halving + cooldown chain (#726/#727) — which also got a sharper message: it now names truncation explicitly (`incomplete data line(s)` / `no final frame terminator`) instead of the generic "carried N events but no summary text", so logs say what actually happened.

No protocol-semantics change for well-formed streams: complete streams extract exactly what they did before (verified by the pre-existing #727 fixtures, including the shape where `completed` carries an empty `output` and extraction falls back to accumulated deltas).

## Tests

- New `tests/preflight-sse-extraction.test.ts`: 23 unit tests across all three protocols — mid-frame truncation, framed-but-missing-completed, delta-less done/item-done, delta+done coexistence (no duplication), multi-part done concatenation, every failure terminal, event-header typing + cross-frame leak guard, CRLF bodies, garbage-after-completed still accepted, diagnose messages.
- Extended `tests/preflight-empty-summary.test.ts`: mock upstream that serves the #764-shaped truncated summary stream; full-pipeline test asserts 502 `preflight_compress_failed` with the truncation diagnosis, bounded summary calls, nothing forwarded, dead-end marker armed, cooldown spends zero upstream calls.

## Pre-flight

- `npm run typecheck` — clean
- `npm test` — 1535 pass / 0 fail / 2 skipped
- `npm run build` — ok
- E2E (`tests/e2e/e2e-codex.test.ts`): not run from this sandbox — it needs a reachable local upstream port, which the sandbox network policy forbids. The change is confined to pure string processing in `src/preflight.ts` plus its tests; CI's manual-dispatch e2e workflow covers the full client path if desired.

Fixes #780. Refs #726, Refs #764.


ranxianglei:
review 了吗

ranxianglei:

Reviewed — line-by-line diff review plus full verification at head `f298320`. Verdict: **ready to merge** (merge itself stays human-only per repo rules).

## What I checked

**Diff cleanliness.** Single commit, 3 files, all on-purpose: `src/preflight.ts` (+86/−16), new `tests/preflight-sse-extraction.test.ts` (23 unit tests), `tests/preflight-empty-summary.test.ts` (+57, one integration test). No version bump, no lockfile drift, no unrelated churn. Every line serves the PR's stated goal.

**Correctness against #780's four points.** All four gaps are closed exactly as specified in the issue:
1. Frame-integrity check (`framed`, CRLF-tolerant) + unparseable `data:` lines counted as `badFrame` instead of silently skipped → truncated streams return `""` and fall into the existing #727 diagnose/halving/cooldown chain instead of persisting a half summary. Verified end-to-end through `requestSummary` → `diagnoseEmptySummary` → halving worklist → budget cap → fail-fast with the diagnosis carried into the 502 message.
2. `event:` header fallback present, and correctly **reset at frame boundaries** (the sticky-type leak from the issue's local patch is not present).
3. Delta-less terminals handled: `response.output_text.done` (`text`) and `response.output_item.done` (via `extractSummaryText` on the item), with `terminalText || out` precedence keeping `response.completed`'s full text authoritative — matching the original immediate-return semantics for complete streams.
4. Failure terminals (`error`, `response.incomplete`/`.failed`/`.error`, bare `{error}` objects) invalidate accumulated text across all three protocols.

Edge cases I traced deliberately: CRLF bodies; OpenAI chat frames without a top-level `type` still route correctly (empty type falls through to the choices branch); responses streams trusted once `completed` is seen even if bytes follow or the final terminator is missing (gateway "close right after completed" shape); anthropic/openai intentionally do NOT require `message_stop`/`finish_reason`/`[DONE]` (#764 gateways omit them) but DO require clean framing — the fail-closed tradeoff is documented in comments and its cost is bounded by the retry chain. The two deviations from the reporter's local patch noted in the #780 triage are both implemented as decided.

**Verification run at head:**
- `npm run typecheck` — clean
- `npm test` — 1535 pass / 0 fail / 2 skipped
- `npm run build` — OK
- `git merge-tree` against current `origin/master` (`467d8d9`) — zero conflicts, clean merge

## One correction: why e2e can't signal here

The description says the sandbox network policy blocks the e2e upstream. That's not what I found — I ran both the zero-token preflight (passes) and the full suite, then a **baseline run of the same suite on the PR base `bfb0b68`**: both fail identically at phase 1 warmup because the default local upstream (`127.0.0.1:8199`) is simply **not running** (ECONNREFUSED on every forward; confirmed still down now). So this is an environment gap, not a PR regression — but it also means e2e gave no signal here at all. Since this change touches the preflight path (in scope for the "run e2e before merge" rule in AGENTS.md), please dispatch the manual CI e2e workflow (`.github/workflows/ci-e2e.yml`) before or shortly after merging — it has the repo secrets and a real upstream.

Minor non-blocking nit: `diagnoseEmptySummary` labels mid-stream parse failures as "(stream appears truncated…)" too — a corrupted-but-complete stream isn't strictly truncated. The "appears" hedge makes it acceptable as-is; no action needed.

---

**中文摘要：** 已逐行审查并全量验证（typecheck 干净、1535 测试全过、build 通过、与当前 master 无合并冲突），实现完整覆盖 #780 的四点加固且两处对本地补丁的修正均已落实；本沙箱因本地上游服务未启动导致 e2e 无法出信号（基线 bfb0b68 与 PR 头同样失败，属环境问题非回归），建议合并前后用 CI 手动跑一次 e2e 补信号。**可以合并。**

PR #789, opened 2026-09-14
fix: name entries dropped by plugin install/remove; neutral /acp no-proxy warning (#788)

### LINKED ISSUE #788: pi plugin: `bili plugin install pi` silently drops the `npm:billion-context-pi` entry; /acp "no proxy detected" warning misleads non-proxy users
来源: 分析 [billion-context-pi#445](https://github.com/ranxianglei/billion-context-pi/issues/445)（用户报告 `Warning: bili: no proxy detected (run via \`bili <client>\` or set a /bili/ baseURL)`）时发现

Real user confusion surfaced via billion-context-pi#445: a pi user who followed the documented pi path (`pi install npm:billion-context-pi`) started seeing the thin-plugin warning and believed pi needed `bili` running. Two defects in this package contribute:

## 1. `bili plugin install pi` silently removes the `billion-context-pi` entry

`piInstall()` (src/plugin-install.ts) filters **every** `packages` entry matching `isPiEntry()` out of `~/.pi/agent/settings.json` before appending its own root — and `isPiEntry()` deliberately matches `npm:billion-context-pi` / `node_modules/billion-context-pi` paths (comment calls it "legacy"). The command output only reports `pi: installed -> … packages += <root>`; nothing is printed about what was removed.

Repro:
1. `pi install npm:billion-context-pi`
2. `npm install -g billion-context && bili plugin install pi`
3. `cat ~/.pi/agent/settings.json` → the `npm:billion-context-pi` entry is gone, no notice was shown.

Impact: the user is left with only the thin proxy-client plugin. Plain `pi` launches then have **no** context compression at all, and `/acp` warns about proxy mode — even though both READMEs' "Which do I need?" tables still recommend `billion-context-pi` (in-process) as *the* pi path. The removal contradicts the documented recommendation without telling the user.

Suggested fix: report removed entries explicitly in the command output (at minimum when they match `billion-context-pi`), e.g. `pi: replaced existing entry npm:billion-context-pi`.

## 2. Misleading `/acp` warning for users who never intended proxy mode

src/agent/pi.ts registers `/acp` unconditionally (whenever `registerCommand` exists). When `detectProxyBase(ctx.model.baseUrl)` is undefined it notifies:

```
bili: no proxy detected (run via `bili <client>` or set a /bili/ baseURL)
```

For anyone who does not intend to use proxy mode (i.e. everyone following the pi → billion-context-pi recommendation), the actionable fix is *not* "run via `bili <client>`" — it is "remove this plugin" (`bili plugin remove pi`). The current text assumes proxy intent and steers users into a different usage mode of the same product family, which is exactly what confused the #445 reporter.

Suggested fix: make the message neutral and offer both exits, e.g. "bili plugin is loaded but this model is not routed through a bili proxy. Launch via `bili <client>` to use proxy mode, or remove the plugin (`bili plugin remove pi`) if you use billion-context-pi or don't want a proxy."

Happy to send PRs for either if the direction looks right.

ranxianglei:
Fixes #788

## What

Two user-facing defects found via billion-context-pi#445 (a pi user who installed `npm:billion-context-pi` per the docs then ran `bili plugin install pi` lost the in-process plugin silently, and `/acp` steered them toward proxy mode they never wanted):

1. **Silent entry removal.** `piInstall()` filtered every `isPiEntry()` match out of `~/.pi/agent/settings.json` — deliberately including `npm:billion-context-pi` — and only reported the addition. Now the output names every dropped entry:
   ```
   pi: installed -> ~/.pi/agent/settings.json (packages += <root>)
   pi: replaced existing entries: npm:billion-context-pi
   ```
   `piRemove()` likewise reports what it drops (`pi: removed entries: ...`). The removal *set* is unchanged — replacing every match is intentional (exactly one bili plugin must stay live, or both double-register the ACP tools); only the silence was fixed.
   - src/plugin-install.ts
2. **Proxy-assuming `/acp` warning.** The unconditional `/acp` command now warns with neutral, agent-aware wording offering both exits:
   > bili: no proxy detected — run via `bili pi` (or set a /bili/ baseURL) to use proxy mode, or remove this plugin (`bili plugin remove pi`) if you use billion-context-pi or don't want a proxy

   The `billion-context-pi` mention is pi-only; omp gets the same shape without it. dsh-acp.ts keeps its own separate message.
   - src/agent/pi.ts

## Tests

tests/plugin-agent.test.ts:
- fresh install asserts NO replaced line; first remove asserts it names the dropped root;
- legacy-replacement fixture asserts all five dropped entries are named and unrelated entries are not;
- `/acp` test asserts the dual-exit wording; new omp variant test asserts agent-awareness + pi-only billion-context-pi hint.

## Pre-flight

- `npm run typecheck`: clean
- `npm test`: 1512 pass, 0 fail, 2 skipped (pre-existing gated skips)
- `npm run build`: success
- E2E suite not run: change touches neither the server request pipeline (server.ts / src/loop/* / adapters / preflight) nor compression behavior — CLI output strings + one client-side `ui.notify` message only.

No version bump (content branch). devlog/2026-09-15_788-plugin-removal-notice included.


ranxianglei:

[bot] 🏷 PR reviewed — diff clean, all pre-flight claims independently re-verified on the PR head. Ready for merge (merge itself is human-only).

**Diff cleanliness (first gate)**
- Single commit `f12bc07` directly on `master` (`bfb0b68`): 5 files — `src/plugin-install.ts`, `src/agent/pi.ts`, `tests/plugin-agent.test.ts`, devlog REQ/WORKLOG. Every line serves #788; no unrelated files, no whitespace churn, no version bump (correct for a content branch).

**Independent verification (ran locally at PR head)**
- `npm run typecheck`: clean.
- `npm test`: **1512 pass, 0 fail, 2 skipped** — matches the stated numbers exactly. New tests present and green: `/acp` dual-exit wording (pi), new omp variant asserting agent-awareness + pi-only `billion-context-pi` hint, roundtrip test asserting fresh-install emits NO replaced line, first-remove names the dropped root, legacy fixture names all five dropped entries and leaves `/home/x/other-package` untouched.
- `npm run build`: success.
- E2E skip justified: the diff touches neither the request pipeline (`server.ts`/`src/loop/*`/adapters/preflight) nor compression behavior — CLI output strings + one client-side `ui.notify`. Both compression modes unaffected.

**Layer check (why the fix hits the root cause, not the symptom)**
- Defect 1's root cause is *observability*, not the removal set: replacing every `isPiEntry()` match is load-bearing (exactly one bili plugin live, or ACP tools double-register). The PR keeps the set byte-identical (`kept`/`removed` partition unchanged in both `piInstall()` and `piRemove()`) and only makes the side effect visible — correct layer. Idempotency paths (`already installed` / `not installed`) preserved and asserted.
- Defect 2's root cause was proxy-intent assumption in the wording. New message interpolates the real agent name (`bili pi` / `bili omp`) instead of the literal `<client>` placeholder — more actionable than the issue's suggested text, while keeping both exits. `agent` is in scope from the factory (`const agent = agentName(agentOverride)`, src/agent/pi.ts:254); `agentName` yields only `"pi"|"omp"`, so the third `removeHint` arm is unreachable-but-harmless defensive code.
- Scope completeness: `dsh-acp.ts` keeps its own separate message (untouched, as claimed); `opencode.ts` registers no `/acp` command, so the pi/omp factory is the full surface. No stale references to the old wording anywhere else in the repo (README/docs checked).
- Scoping the entry-reporting to pi (not omp/codex/opencode) is right: only pi's `packages[]` can hold a *different documented package* (`npm:billion-context-pi`) that the user chose themselves; the others remove their own single key/file.

No small issues found to patch on the branch.

中文摘要：审查了 PR #789——diff 干净（单 commit、5 个文件、无版本变更），我在 PR 分支上独立复跑了 typecheck（clean）、全量测试（1512 过 / 0 挂 / 2 跳过，与 PR 声明一致）、build（成功），并确认修复打在了根因层（删除集合不变、只补可见性；警告改为中性双出口且按 agent 区分措辞），未发现需补修的小问题，可以合并。

Issue #670, opened 2026-09-09
fix(ci): PR Build Artifact publish step fails with ENEEDAUTH on fork PRs


ranxianglei:
来源: #668 (https://github.com/ranxianglei/billion-context/pull/668) 分析 preflight 修复 PR 时发现

## Problem

The `PR Build Artifact` workflow (`.github/workflows/pr-artifact.yml`) triggers on every `pull_request` to master, including PRs from forks. Its "Publish to npm with PR tag" step authenticates with `NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}`. GitHub does not pass repository secrets to workflows triggered by fork PRs, so on fork PRs the token is empty and `npm publish` fails with `ENEEDAUTH`.

## Repro

1. Open a PR to master from a fork (observed on #668).
2. The `build-artifact` job builds successfully, then fails at "Publish to npm with PR tag" with `ENEEDAUTH` (npm requires login for `npm publish`).

## Impact

- The PR shows a failed CI job even though the source is fine, muddying the checks signal on fork PRs.
- All steps after the publish step are skipped: tarball creation, artifact upload, and the PR comment with install instructions — fork PR authors lose the entire artifact feature.

## Suggested fix

Gate the publish step to non-fork PRs:

```yaml
- name: Publish to npm with PR tag
  if: github.event.pull_request.head.repo.full_name == github.repository
```

and make the "Comment on PR" step drop the "Option A — Install from npm PR tag" section for fork PRs so the comment does not advertise a tag that was never published. The tarball, artifact upload, and comment steps should still run for fork PRs.

## Notes

- No credential or workflow permission change is needed; this is a workflow-gating issue.
- Verified on #668: build succeeded, publish failed with ENEEDAUTH, job stopped before the artifact steps.

ranxianglei:
> 🤖 Powered by [ework](https://github.com/ranxianglei/ework-aio) · qwen3.8-27b

[bot] 🏷 Picked up. Verifying the claim against #668's actual workflow run (checking the failed step + error), then I'll implement the gating fix. Will follow up with a PR link.

ranxianglei:
> 🤖 Powered by [ework](https://github.com/ranxianglei/ework-aio) · qwen3.8-27b

Verified and fixed. PR: https://github.com/ranxianglei/billion-context/pull/671

## Verification (triage)

[...]

PR #797, opened 2026-09-15
fix(launcher): read opencode JSONC config (comments + merge) like opencode itself

### LINKED ISSUE #796: opencode launcher/discovery ignores opencode.jsonc — provider discovery comes back empty
## Problem

opencode accepts JSONC (comments, trailing commas) in **every** config file, and its preferred global config file is `opencode.jsonc` — it even seeds a near-empty one (just `$schema`) when no config exists yet. It also **merges** `config.json` → `opencode.json` → `opencode.jsonc` when reading (later wins).

billion-context's readers hardcode the opposite assumptions:

- `resolveOpencodeConfigFile` (`src/client-config.ts`) always returns `~/.config/opencode/opencode.json`
- `readOpencodeConfig` uses strict `JSON.parse`
- `prepareOpencodeHttpRewrite` (`src/launcher.ts`) uses strict `JSON.parse`

## Impact (for a `.jsonc` user, e.g. me)

- **Provider discovery is empty** — `readOpencodeConfig` returns `{providers:{}}`, so no model context windows are collected and no HTTP `/bili/` rewrites are planned: plaintext-HTTP upstream providers bypass the proxy's URL rewrite entirely (HTTPS upstreams still ride cert-MITM, which is why this can go unnoticed).
- **`prepareOpencodeHttpRewrite` emits a plugin-only temp copy** — the parse failure is caught, so the temp `OPENCODE_CONFIG` carries only the plugin + `compaction.auto:false`, and the rewritten-provider copy path is silently lost.

## Suggested fix

Mirror opencode's own discovery + merge semantics:

1. Tolerant parse (`parseConfigText`): try `JSON.parse`, fall back to `jsonc-parser` (devDep, inlined via tsup `noExternal` so `dist` stays self-contained).
2. `readOpencodeConfigRoot(env)`: `OPENCODE_CONFIG` read directly; otherwise merge `config.json` → `opencode.json` → `opencode.jsonc` in opencode's order — the merge is *required*, not optional: since opencode seeds a near-empty `opencode.jsonc`, a single-file `.jsonc`-preference read would miss the real config living in `opencode.json`.
3. `resolveOpencodeConfigFile`: first **existing** file among `["opencode.jsonc", "opencode.json", "config.json"]` (mirrors opencode's `globalConfigFile`).
4. `prepareOpencodeHttpRewrite` takes the parsed merged root (deep-cloned, so the baseURL rewrite cannot mutate the caller's merged read — the old file-reading form was accidentally safe only because it re-read the file on every call).

PR incoming with tests (JSONC comments/trailing commas, merge order, seeded-jsonc-shadowing, OPENCODE_CONFIG bypass, non-mutation of the caller's root).

jensenojs:
Fixes #796

## What

opencode accepts JSONC (comments, trailing commas) in every config file, prefers `opencode.jsonc` (seeding a near-empty one when no config exists), and **merges** `config.json` → `opencode.json` → `opencode.jsonc` when reading. bili's readers hardcoded `opencode.json` + strict `JSON.parse`, so `.jsonc` users got empty provider discovery and a plugin-only temp `OPENCODE_CONFIG` copy.

## Changes

- `parseConfigText` (`src/client-config.ts`): JSON → JSONC-tolerant parse via `jsonc-parser` (devDep, inlined through tsup `noExternal`; `dist` stays self-contained, +33 KB).
- `readOpencodeConfigRoot(env)`: `OPENCODE_CONFIG` read directly; otherwise merges the three global files in opencode's own order. The merge is required, not optional — a seeded near-empty `opencode.jsonc` must not shadow the real config in `opencode.json` (see the test covering exactly that).
- `resolveOpencodeConfigFile`: first **existing** file among `opencode.jsonc` / `opencode.json` / `config.json` (mirrors opencode's `globalConfigFile`); all-missing falls back to the `.jsonc` path.
- `parseOpencodeProviders`: the provider/model-window extraction, factored out of `readOpencodeConfig` so both single-file and merged readers share it.
- `prepareOpencodeHttpRewrite` (`src/launcher.ts`) now takes the parsed merged **root** instead of a config file path, and deep-clones it: the old form only avoided mutating caller state because it re-read the file on every call — with a shared merged root the baseURL rewrite would otherwise leak into the caller (covered by a regression test).
- README: one line documenting `.jsonc`/merge support in the `bili opencode` entry.

## Tests

New/updated in `tests/launcher.test.ts`:
- `readOpencodeConfig` parses JSONC (comments + trailing commas)
- `readOpencodeConfigRoot`: merge order (later file wins), seeded-empty-`.jsonc` does not shadow `opencode.json` providers, `OPENCODE_CONFIG` bypasses the global dir, empty dir → `undefined`
- `resolveOpencodeConfigFile`: `.jsonc` preferred, `.json` fallback, `config.json` last, `OPENCODE_CONFIG` wins
- `prepareOpencodeHttpRewrite`: JSONC root → rewritten `/bili/` copy, original file untouched, plugin injection, missing root still emits plugin-only copy, and the caller's root object stays unmutated across successive calls
- `pluginDirMode` test updated to the new root parameter

## Pre-flight

- `npm run typecheck` ✅
- `npm test`: 1602 pass / 1 fail — the single failure (`runLaunch codex: budget args injected for MITM mode (built-in table window)`) reproduces identically on clean master in this environment (stashed and re-ran), so it is pre-existing and environment-dependent, not introduced here
- `npm run build` ✅
- Live check against a real commented `opencode.jsonc`: `readOpencodeConfigRoot` now surfaces all six providers (previously `{}`)


ranxianglei:

[bot] 🏷 Acknowledged — reviewing PR #797 now. Fetched the branch (`jensenojs:2026-09-15_opencode-jsonc-config`, single commit `164b075` on top of current master). Diff is tight: 7 files, all scoped to the opencode JSONC/merge change. Running typecheck / full test suite / build locally and checking CI status; will report findings shortly.

ranxianglei:

[bot] 🏷 Review complete — pulled the branch (`2026-09-15_opencode-jsonc-config`, `164b075`) and verified everything below.

**Local pre-flight** (on PR head): `npm run typecheck` ✅ · full suite `node --import tsx --test tests/*.test.ts` → **1603 pass / 0 fail / 2 skipped** (the codex MITM-budget failure you noted did not reproduce here) · `npm run build` ✅ (dist stays self-contained, `jsonc-parser` inlined via tsup `noExternal`).

**CI**: all four *required* `test (...)` checks green + zero-token e2e check green. The red `build-artifact` is **not a code failure**: checkout/install/build/pack/upload-artifact all succeeded — only the final "Comment on PR with install instructions" step threw `HttpError: Resource not accessible by integration`. For fork PRs the `GITHUB_TOKEN` is forced read-only regardless of declared `permissions:`, so the bot cannot comment (the publish step already guards against forks per #670; the comment step doesn't). It hits every fork PR identically and is not a required check, so it doesn't block merge. Minor infra gap — suggested fix: skip or fail-soft that step when `head.repo.full_name != repository`. Happy to file a separate issue if you want it tracked.

**Parity vs real opencode source** (`packages/opencode/src/config/config.ts`): file preference (`opencode.jsonc` → `opencode.json` → `config.json`, first existing, fallback `.jsonc`), near-empty config seeding, and merge order (later file wins) — all match exactly. ✅

**One substantive finding — shallow vs deep merge.** opencode merges the three global files with remeda `mergeDeep` (**deep**); this PR uses a shallow `{...root, ...next}` spread. Consequence: if a user splits providers across files (provider A in `opencode.json`, provider B in `opencode.jsonc`), opencode sees both, but bili's merged root drops A's `provider` map — A goes undiscovered, so bili misses its budget alignment and baseURL rewrite, and that provider silently bypasses the proxy. Narrow edge case (single-file configs and the seeded-empty-`.jsonc` case are fine either way), but the PR's stated goal is parity with opencode's own merge, so I recommend fixing before merge (~10 lines, no new dep) — illustrative patch:

```ts
function mergeRoot(t: Record<string, unknown>, s: Record<string, unknown>): Record<string, unknown> {
    const out: Record<string, unknown> = { ...t };
    for (const [k, v] of Object.entries(s)) {
        const p = out[k];
        out[k] = isPlainObj(p) && isPlainObj(v) ? mergeRoot(p, v) : v; // arrays/scalars replaced, like remeda default
    }
    return out;
}
// readOpencodeConfigRoot: root = root ? mergeRoot(root, next) : next;
```

**Minor nits** (non-blocking): the merge loop hardcodes `["config.json", "opencode.json", "opencode.jsonc"]` instead of deriving from `OPENCODE_CONFIG_FILES` (two sources of truth for one order); `parseConfigText` ignores jsonc-parser's error array, so a partially-garbage parse is still used — acceptable (arguably better than discarding), just noting it.

Diff is clean and tightly scoped; the new tests cover exactly the right cases (seeded-shadow, merge order, OPENCODE_CONFIG bypass, caller-root mutation regression). Since I can't push to your fork, please update the branch with the deep-merge change and I'll re-verify.

**中文摘要**:该 PR 让 bili 像 opencode 一样读取 JSONC 配置并按 opencode 自身的顺序合并三个全局配置文件,代码/diff 干净,本地与 CI(必需项)全部验证通过;唯一实质问题是三文件合并用了浅展开而 opencode 是深合并(remeda `mergeDeep`),providers 跨文件拆分时会被漏发现导致绕过代理,建议补上约 10 行的深合并逻辑后即可合并。

ranxianglei:

Reviewed the diff against opencode's own config loader (checked current opencode source) and pushed two fixes directly onto the branch — see commit `d50652c`.

**What was right from the start:**
- `parseConfigText` matches opencode exactly (`JSON.parse` → jsonc-parser with `allowTrailingComma: true`, same as opencode's `ConfigParse.jsonc`).
- `resolveOpencodeConfigFile` mirrors opencode's `globalConfigFile()` (first existing of `opencode.jsonc` → `opencode.json` → `config.json`).
- The rewrite path taking a parsed, cloned root is right; `structuredClone` keeps it non-mutating.

**Two bugs fixed on the branch (`d50652c`):**

1. **Shallow merge dropped split-file providers.** The three-file merge used top-level spread (`{ ...root, ...next }`). Whenever both `opencode.json` and `opencode.jsonc` defined entries under a top-level `provider` key, every provider from the earlier file was silently dropped — exactly the split-file layout this PR exists to support. (The seeded-empty-`.jsonc` case alone survives a shallow merge, which is why it was easy to miss.) opencode deep-merges its config files (`mergeDeep`: plain objects recurse, arrays/primitives replaced by the later file). Added `mergeConfigDeep` with those exact semantics.
2. **`OPENCODE_CONFIG` replaced globals instead of layering over them.** `readOpencodeConfigRoot` early-returned a direct read of `$OPENCODE_CONFIG`. opencode loads the global files first and merges the explicit file *on top of them* — so with the env var set, providers living only in the global files were invisible to discovery and to the rewritten temp overlay. Now the merged root is `config.json` → `opencode.json` → `opencode.jsonc` → `$OPENCODE_CONFIG`.

**Test updates:** cross-file provider coexistence + same-provider deep-merge assertions; `OPENCODE_CONFIG` layering assertions; a hermeticity fix in the `prepareOpencodeHttpRewrite` test (its root read was picking up the host user's real opencode config — pinned `XDG_CONFIG_HOME` to an empty dir); non-mutation assertion on the caller's root.

**Verification:** typecheck ✓ · full suite 1603 pass / 0 fail / 2 skip · build ✓ · `dist` self-contained (jsonc-parser inlined via tsup, zero runtime deps).

**CI note:** the `build-artifact` job fails at its final "Comment on PR" step on any fork PR (`Resource not accessible by integration` — the GITHUB_TOKEN is read-only for fork-triggered workflows). Pre-existing (it fails identically on the original push) and it is not a required check, so it doesn't block merging. All four required test jobs plus the fake-upstream E2E are green on `d50652c`.

> 中文摘要：review 修复了两个配置合并的语义 bug（顶层浅合并会在两个文件都定义 provider 时丢掉先读文件的条目；OPENCODE_CONFIG 应叠加在全局配置之上而非替代它），已在 PR 分支直接补了 commit `d50652c` 并加了相应测试，本地 typecheck/test/build 全过、四项必需 CI 全绿，可以合并（按仓库规则合并需人工操作）。

PR #799, opened 2026-09-15
fix: reserve Codex preflight headroom and avoid fork artifact comment failures

### LINKED ISSUE #798: Codex preflight can stop above the native compaction takeover threshold
Codex Responses requests in intercept mode can leave ordinary preflight above the health threshold required to intercept the next native compaction request. This is reproducible on master `da6cdb7d1538b280cd600f66ef156cad1b26d382` (v0.1.113).

The ordinary preflight trigger and stopping condition use the full configured window, while `codexCompactGatePre()` rejects `lastInputTokens >= effectiveLimit * 0.9`. For a 200,000-token bili budget, preflight can report `215252 -> 198404` as success, but the next native trigger requires less than 180,000. It consequently takes the native fallback, which deliberately skips preflight. In the observed long approval-reviewer session, that upstream native compaction failed with a context-window error. This is a general Codex intercept integration gap; it is not specific to a terminal host or the approval model. The budget here is bili's resolved budget, not a claim about the upstream model's actual capacity.

This differs from #769 / #770: that patch correctly normalizes private bili compaction markers in the native fallback; it did not change capacity thresholds. It remains intact.

Deterministic reproduction: send a detected Codex Responses request with approximately 18.5K tokens under a 20K budget, with compressible old messages and protected recent messages. Preflight currently makes no room. With a shared best-effort 90% target, the request is folded below 18K and the following native trigger is intercepted. The same issue occurs in plugin mode. An already overflowing request must also continue folding after first fitting the hard window.

Proposed fix: share the existing 90% constant and use it as the ordinary Codex-intercept preflight target. Keep hard overflow, chunk sizing, protected-content relaxation, and fail-fast decisions tied to the real configured limit. Missing headroom alone must not reject an otherwise fitting request. Preserve the native fallback ordering and raw-request invariants.

The accompanying PR also addresses the fork artifact workflow failure observed in #770: build/upload succeeded, but automatic commenting failed with `Resource not accessible by integration`. Write install instructions to the Actions summary for every PR and skip comment API calls for forks; no token permission escalation is needed.

z0d0dOr:
Codex preflight could stop just below the full configured window while native compaction interception requires usage below 90%. For a 200K bili budget, `215252 -> 198404` passed preflight but forced the following native request onto the upstream fallback. Long approval-reviewer histories exposed the gap, but it affects detected Codex Responses clients generally in intercept mode.

This change shares the existing 90% health threshold and uses it as the ordinary preflight target, including plugin mode. The hard window still controls failure decisions, chunk sizing, and relaxation of recent-content protection. Hard-fitting protected payloads and cooldown cases continue forwarding when additional headroom is unavailable. Native fallback byte preservation and #770 marker normalization remain unchanged. Non-Codex and pass-mode requests keep their existing target.

A separate CI commit fixes the artifact-comment failure seen in #770: all PRs receive install instructions in the Actions summary; fork PRs skip comment API calls. Same-repository PR comments still create/update normally. Build, upload, publishing guards, and token permissions are unchanged.

Validation:
- Six new HTTP regressions: proxy/plugin headroom followed by native interception; continued compression after fitting the hard window; protected content; pass-mode and non-Codex controls. Baseline: three failed and three passed; patched: all six passed.
- 49 existing related regressions passed, including native fallback/private-marker, preflight, hold, fork estimate, and empty-summary coverage. One random ephemeral-port `fetch failed` / `bad port` occurred before proxy interaction; that single unchanged case passed on retry.
- Typecheck and build passed; dependency and version files are unchanged.
- Extracted Actions script passed fork, same-repository create, and same-repository update simulations.

No live paid-model/native-client long-history compaction E2E was forced. This prevents the ordinary-preflight threshold gap; it does not claim to revive an already stuck native request or resolve upstream service overload.

Fixes #798


ranxianglei:

[bot] 🏷 Reviewing PR #799 now: fetching the branch, checking diff cleanliness, CI status, and running the new regressions locally before reporting back.

ranxianglei:

[bot] 🏷 **Review complete — approved, ready for human merge.**

**Diff cleanliness** — 2 commits, 5 files (+183/−17): `.github/workflows/pr-artifact.yml`, `src/codex-compact.ts`, `src/preflight.ts`, `src/server.ts`, `tests/preflight-codex-headroom.test.ts`. Every line serves the stated purpose; no version/dependency drift, no unrelated churn.

**Root cause check (vs #798)** — confirmed the reported layering is correct: ordinary preflight triggered/stopped at the full window while `codexCompactGatePre()` requires `lastInputTokens < limit * 0.9`, so a payload compressed to just under 100% (e.g. `215252 → 198404` on a 200K budget) left the next native compaction request above the 180K gate → upstream fallback. The fix attacks the root cause (shared threshold), not the symptom.

**Code review highlights**
- `CODEX_COMPACT_HEALTH_RATIO = 0.9` (`src/codex-compact.ts`) is now the single source of truth used by both the compact gate and the preflight target — the gap cannot reopen via divergent literals.
- Target scoping is exact (`src/server.ts` `preflightCompressIfNeeded`): 90% only when `protocol === "responses"` + Codex UA + `intercept` mode; plugin mode included via session binding; pass-mode and non-Codex keep the full-window target (control tests confirm).
- Hard-window semantics preserved (`src/preflight.ts`): `target` is clamped to `limit`, and resets to `limit` once recent-content protection relaxes — chunk sizing and failure decisions still track the configured window (protected-content test confirms no relaxation-for-headroom).
- Three adjacent changes I scrutinized specifically, all load-bearing and correct:
  1. No-compressible-ranges forward now uses `payloadFitsWindow`; for unknown-baseline sessions fit is proven with the conservative upper bound (`estimateCoreMessagesUpper`) instead of the optimistic estimate — addresses the #553 concern while letting headroom-triggered-but-fitting payloads forward instead of failing fast.
  2. Dead-end cooldown (#726) now passes through payloads that fit the hard window — correct, since no summarization call is needed and the overflow failure mode the cooldown guards against doesn't apply.
  3. The image-dominated forward-once escape (#496) now additionally requires `payloadEstimate >= limit` — without this, a 90–100%-band payload would skip compression entirely via that path and the headroom goal would silently fail.
- CI commit: `isFork` is defined before use; `core.summary.addRaw(body).write()` runs for every PR, `if (isFork) return;` precedes all comment-API calls (fixes the #770 `Resource not accessible by integration` failure); same-repository create/update paths and token permissions untouched. Valid API for `actions/github-script@v7`.

**Validation (independently reproduced)**
- Local: typecheck ✅ · tsup build ✅ · 6/6 new HTTP regressions ✅ · 94/94 related existing regressions (preflight compress/responses/fail-fast/hold/fork-estimate/empty-summary/max-output/multi-range/sse-extraction/stream-summary, codex-compact, native-compact-terminal, marker-echo, fallback-window-floor, fork-nudge-trigger) ✅.
- CI on head `000c8c7`: all 7 checks green — test ubuntu 22/24, windows 22/24, version-guard, build-artifact, and the real-codex-vs-fake-upstream zero-token E2E.
- Not run locally: the full paid/native-client long-history E2E (needs a live local upstream); the author discloses this explicitly, and CI's real-client zero-token check covers the wire behavior. No objection.

No issues found requiring code changes. Merging stays human-only per repo rules.

---
中文摘要：修复了 Codex intercept 模式预检只压到满窗口线、导致后续原生压缩请求越过 90% 健康门限而走上游回退的问题（阈值改为共享常量，预检默认目标 90%，硬窗口仍管失败判定与保护松弛）；同时修复 fork PR 的 artifact 评论失败（安装说明写入 Actions summary，fork PR 跳过评论 API）。本地复跑 6 条新回归 + 94 条相关旧回归、typecheck/build 全绿，CI 全绿，可以合并。

PR #805, opened 2026-09-15
fix: propagate in-band OpenAI stream errors

amirzamli:
Propagates upstream overload/errors that arrive inside successful HTTP 200 SSE streams instead of converting them into empty successful stops.

Changes the OpenAI adapter to surface choices-less error frames as errors and prevents the gateway from synthesizing [DONE] after an in-band Responses failure.

Validation: targeted adapter tests, typecheck, build, and gateway tests passed.


amirzamli:
Follow-up: added an English language toggle for the web UI with persisted browser preference, translated navigation/configuration/status text, and regression coverage. Validation: typecheck and web UI tests passed.

ranxianglei:

[bot] 🏷 Review complete — the underlying bug is real and the fix approach is sound, but the PR needs one structural change before it can merge.

**Bug verification (confirmed on master)**
In `src/loop/adapter-openai.ts`, a parsed frame without `choices` is silently dropped (`continue`, ~line 327) unless it carries `usage`. An in-band `{error:{...}}` frame therefore vanishes, the following `[DONE]` yields `done` with `finishReason: "stop"`, and the loop emits a successful empty completion — exactly the reported symptom. I also checked the Responses adapter: it already handles `response.failed` / `response.error` (`src/loop/adapter-responses.ts:500`), so scoping the fix to the chat adapter + shared core is correct.

**Fix assessment (commit c779d71)**
- New `error` event kind flowing adapter → `runCompressLoop` → the existing `!sawDone` terminal path (`yield adapter.emitError(...)`) replaces the synthetic empty stop. Correct layering — the error rides the same channel as a hard truncation, and the message now carries the upstream detail instead of the generic "no completion event".
- Bonus behavior worth noting: when the error arrives with zero bytes forwarded, the round takes the existing #413 invisible-retry path first, so a transient overload gets one free retry before surfacing as an error. That's a good fit.
- CI on the head commit is fully green: full test matrix (ubuntu/windows × node 22/24), version-guard, build-artifact, and the real-codex-vs-fake-upstream e2e.
- Duplicate screening: no existing issue/PR covers in-band stream errors (or the web UI i18n topic).
- GitHub shows `mergeable_state: blocked`, but that's stale — I verified locally that current master is an ancestor of your head (fast-forward possible) and a simulated merge produces no conflicts.

**Blocker: unrelated commit bundled in**
Commit `b161343` ("feat: add English translation to web UI") is a separate feature riding in a bug-fix PR. Repo rules require every line in a PR to relate to its stated purpose, so please split it out: rebase this PR down to just `c779d71`, and open the i18n work as its own PR. Two observations for that follow-up PR (not blockers here): the `translations` object has duplicate keys (`"客户端接入"`, `"设置"` appear twice), and dynamically-set strings (toasts, busy labels like `"处理中…"`) bypass `translateText()` entirely, so English mode still shows Chinese toasts.

**Minor points on the fix itself (non-blocking, your call)**
1. An in-band error still lands in `noteWeakOverflow` via the `!sawDone` path (`src/loop/core.ts:595-602`) — but an upstream *overload* is not an overflow signal. Repeated overloads would pollute the weak-overflow heuristic that drives preflight compression. Consider excluding explicit stream errors from that accounting (or at minimum logging them distinctly, which you already do).
2. Test coverage is adapter-level only. A loop-level test (mock upstream yielding an error event mid-stream through `runCompressLoop`) asserting a protocol error is emitted and no completion synthesized — plus the zero-bytes-retry firing — would pin the actual regression this fixes.
3. Nit: the `as ParsedStreamEvent` cast at the new yield site is now unneeded since the union includes the error variant (harmless; matches file style).

Happy to prepare a clean replacement branch (master + cherry-picked `c779d71` only) if splitting is easier that way — just say the word. Otherwise, once this PR contains only the fix commit, it's ready for a human merge.

中文摘要：已验证 in-band 流错误被静默丢弃导致空成功 stop 的 bug 真实存在，修复方案（错误帧走统一 error 通道、零字节时复用 #413 重试）正确且 CI 全绿；但 PR 混入了无关的 Web UI 英文翻译提交，需拆分为独立 PR 后方可合并。

ranxianglei:
你直接拆分

amirzamli:
Addressed review follow-ups on the bug-fix branch:

- Removed the unrelated web UI/i18n commit from fix-stream-error; it now contains c779d71 plus the loop regression coverage.
- Excluded explicit in-band stream errors from weak-overflow accounting.
- Added a loop-level regression test covering the zero-byte retry and protocol error output.

Validation: npm run typecheck and OpenAI adapter/loop tests pass.

ranxianglei:

[bot] 🏷 Verified the updated head (`479032a`) — all three follow-ups landed correctly:

1. **Split confirmed**: branch is now master + `c779d71` + `479032a` only; the i18n commit is gone (diff touches exactly `src/loop/adapter-openai.ts`, `src/loop/core.ts`, `tests/openai-toolcall-finish-reason.test.ts`). The i18n work lives in your new PR #807.
2. **Weak-overflow exclusion**: correct placement — the guard `&& streamError === undefined` at `src/loop/core.ts:616` keeps genuine truncations counted while explicit in-band errors no longer pollute the preflight heuristic.
3. **Loop-level test**: covers exactly the regression — asserts the zero-byte invisible retry fires once (`retryFetches === 1`), the upstream detail propagates into the error output, the protocol error marker is emitted with `[DONE]` termination, and the generic "no completion event" text is absent. Good use of a real kernel core/session via `makeLoopCtx()`.

My validation on the new head: typecheck clean, full suite **1613 tests / 1611 pass / 0 fail** (2 skipped = e2e gate), build OK. CI is fully green as well (all 7 checks incl. the real-codex e2e). The earlier "blocked" merge state was stale; the branch is a clean fast-forward of master.

**#805 is ready for a human merge.** I left two small residual notes on #807 (duplicate translation keys; dynamic toast/busy strings bypassing translation) since that PR carries the i18n code now.

中文摘要：已验证更新后的 #805 仅含修复+循环级回归测试，weak-overflow 排除与零字节重试断言均正确落地，本地全量测试与 CI 全绿，可以合并；i18n 部分已在 #807 单独处理。

amirzamli:
Im still seeing some issues with agents unexpectedly stopping, will review this and update if I find the issue.

ranxianglei:
先合并 后续有问题可以直接提交 ework帮我翻译

PR #808, opened 2026-09-15
fix: request-scoped idle watchdog, malformed-body 400 guard, stale-install reminder dedupe (#806)

### LINKED ISSUE #806: Accepted request is never dispatched and never answered — client hangs forever (no request-scoped deadline)
## Summary

A request can be accepted, logged as `forward POST → …`, and then **never dispatched upstream and never answered**. The client socket stays `ESTABLISHED` forever. No timeout fires anywhere, so the client session is bricked until the proxy is restarted. Observed twice on two independent sessions (53 min and 21 min of total silence), on a proxy instance that kept serving *other* sessions normally in between.

The forensic signature is unambiguous: **the proxy process has zero outbound sockets while holding the client's request**.

## Environment

- `billion-context` — wedged process was running **v0.1.112**; v0.1.114 was already installed on disk (see §3, that is part of the bug).
- macOS 15 (Darwin 25.6.0), arm64, Node 24 (`/usr/local/bin/node`).
- Mode: long-lived `bili start --host 127.0.0.1 --port <P>`, shared by several clients.
- Config: `{"upstreamProxyMode":"manual","upstreamProxy":"http://127.0.0.1:<Q>","providers":{}}` (`<Q>` = a local Clash-style HTTP proxy).
- Client: Claude Code 2.1.272 with `ANTHROPIC_BASE_URL=http://127.0.0.1:<P>/bili/https://api.anthropic.com` (`/bili/` prefix mode), upstream = `api.anthropic.com`.

## 1. Main bug — request accepted, never dispatched, never answered

Log for a wedged session (the last line for that session, ever):

```
10:17:28.665Z [info] [<session-B>] nudge INJECT T1: usage=84% (167755/200000) …
10:17:28.665Z [info] forward POST → https://api.anthropic.com/v1/messages
<nothing — for 21 minutes, until the proxy was restarted>
```

Meanwhile the proxy was perfectly healthy for other traffic: an unauthenticated probe through the same instance returned `401` in 0.6 s while these two requests were stuck.

### Evidence that the request never left the process

```
$ lsof -nP -a -p <proxy-pid> -i TCP
node  <pid>  15u  TCP 127.0.0.1:<P> (LISTEN)
node  <pid>  16u  TCP 127.0.0.1:<P>->127.0.0.1:aaaaa (ESTABLISHED)   <- client A
node  <pid>  20u  TCP 127.0.0.1:<P>->127.0.0.1:bbbbb (ESTABLISHED)   <- client B
```

**No socket to the upstream proxy, none to the upstream host.** For comparison, the freshly restarted instance serving the same two clients shows exactly what you'd expect:

```
node  <newpid>  16u  TCP 127.0.0.1:<P>->127.0.0.1:ccccc (ESTABLISHED)  <- client
node  <newpid>  17u  TCP 127.0.0.1:ddddd->127.0.0.1:<Q> (ESTABLISHED)  <- upstream
node  <newpid>  18u  TCP 127.0.0.1:<P>->127.0.0.1:eeeee (ESTABLISHED)  <- client
node  <newpid>  20u  TCP 127.0.0.1:fffff->127.0.0.1:<Q> (ESTABLISHED)  <- upstream
```

### Evidence that nothing is pending

`sample <proxy-pid> 3` on the wedged process: main thread parked in `uv__io_poll`/`kevent`, all four `libuv-worker` threads in `_pthread_cond_wait`, all V8 workers idle. Not a blocked DNS lookup, not a busy loop, not a threadpool exhaustion — the process is simply **idle while owing two clients a response**. Timers still fire normally throughout (the 180 s update check keeps logging on schedule), so the event loop is alive; the request promise was just dropped.

On `SIGTERM` the old process could not drain: `SIGTERM received — flushing sessions… / shutdown grace window elapsed; forcing exit`.

### Why no timeout saves it

In `src/fetch-util.ts` (0.1.114 on disk), `fetchWithTimeout` arms its deadline **inside** itself, i.e. only once `fetch` is actually entered:

```js
async function fetchWithTimeout(url, opts, timeoutMs, externalSignal) {
  const effective = timeoutMs ?? upstreamTimeoutMs();
  ...
  let timer = armTimer();          // only from here on is there a deadline
  const raw = await fetch(url, finalOpts);
```

If the handler never reaches `fetch` (or the awaited promise is dropped between `log("forward …")` and dispatch), **no timer is ever armed**, `res` is never ended, and the client waits forever. `BILI_UPSTREAM_TIMEOUT_MS` cannot help, because it only parameterises a timer that was never created.

### Suggested fix

Arm the deadline **when the request is accepted**, not when `fetch` starts — a request-scoped watchdog in the HTTP handler that guarantees every accepted request terminates in a response or a socket close. Something like:

```js
const deadline = setTimeout(() => {
  if (!res.writableEnded) {
    log('error', `[${sid}] request watchdog: no response after ${ms}ms — failing closed`);
    res.writeHead(504, {'content-type':'application/json'});
    res.end(JSON.stringify({type:'error', error:{type:'timeout_error', message:'bili: request watchdog fired'}}));
  }
}, requestWatchdogMs);
res.on('finish', () => clearTimeout(deadline));
res.on('close',  () => clearTimeout(deadline));
```

A 504 the client can retry is strictly better than an infinite hang: Claude Code retries a pre-first-byte failure, but has no way to recover from a socket that stays open and silent.

## 2. Secondary — `TypeError` on a malformed body becomes a 502

Probing the same instance with `POST /v1/messages` and body `{}`:

```
10:35:06.472Z [info] mitm api.anthropic.com:443 tunnel established (TLS terminated locally)
10:35:06.534Z [error] TypeError: Cannot read properties of undefined (reading 'find')
```

Looks like an unguarded `.find()` on `body.messages` (or similar) when the field is absent. Minor on its own — a validation guard returning 400 would be clearer than a 502 — but it shows the request path has at least one place where an exception escapes into a generic failure.

## 3. Root cause of the long outage — auto-update never restarts, and never escalates

The reason a buggy build served every session for **4+ hours**: auto-update replaced the on-disk install, but the running process kept executing the old code and only warned:

```
[warn] [update] running v0.1.112 but v0.1.114 is installed — restart bili to activate
  (auto-update replaced the on-disk install; this process is still on the old code)
```

repeated every 180 s, in a log nobody watches. Everything downstream inherits a known-buggy proxy indefinitely.

Suggestions, roughly in order of preference:

1. **Surface it to the client** — a one-time in-band notice on the next response, so the human actually sees "proxy is running stale code, restart it".
2. **Opt-in self-restart** — e.g. `--auto-restart-on-update`, restarting when there are no in-flight requests (the instance is single-port and re-binds cleanly; existing clients pointed at the same port reconnect by themselves — verified: two hung sessions recovered on their own after restart).
3. At minimum, escalate the log level and stop repeating identically forever.

## What I verified while debugging (in case it narrows it down)

- DNS is not involved: `dns.lookup('api.anthropic.com')` returns in 10–18 ms.
- The upstream path is not involved: the same request straight through the upstream proxy returns `401` in 0.53 s, at the same moment the proxy is holding two requests.
- Not a global wedge: the instance kept serving other sessions normally between the two hangs.
- Not the client: another client of the same version bypassing the proxy entirely ran for over an hour of heavy traffic with zero interruptions; only proxy-routed sessions hang.
- After restarting onto 0.1.114: 19 forwarded requests, 0 errors, both previously-hung sessions resumed on their own.

I cannot prove the hang still exists in 0.1.114 — the wedged process was 0.1.112 and the failure is not deterministically reproducible. But the missing request-scoped deadline (§1) is visible in 0.1.114's `fetch-util.ts`, so the guard seems worth having regardless of whether the specific drop was fixed in 0.1.113/0.1.114.

ranxianglei:
Fixes #806

## What changed

**1. Request-scoped IDLE watchdog (main bug)** — `src/server.ts`
An accepted request that produces no output for the budget now ends in a `504 gateway_timeout` instead of hanging the client forever. Armed at accept time (before prepare), re-armed on every response write, cleared on close. Default budget = 2x the upstream idle timeout (24 min); override/disable with `BILI_REQUEST_WATCHDOG_MS` (consistent with the existing env-knob conventions).

Design notes:
- **Idle, not total-time**: healthy marathon streams keep writing bytes; worst-case legitimate client-facing silence is one segment bounded by the upstream idle timeout, so 2x gives margin without killing long generations.
- **Firing also aborts the upstream fetch.** Closing the response alone does NOT cancel the fetch — the `res.on("close")` abort handler is gated on `!writableEnded` (correct for normal end-of-stream). The watchdog registers each request's `AbortController` in a WeakMap at both `clientAbort` sites and aborts it on fire, so the per-session lock (`withSessionLock`/`acquireInFlight`) releases promptly instead of being held until the upstream timeout fires.
- Covers `/bili/` and MITM paths alike (same createServer handler).
- Residual: undici keeps an aborted pre-response socket in its pool until `headersTimeout` (verified against plain Node fetch — baseline behavior, identical to the pre-existing IDE-cancel path), bounded by `BILI_UPSTREAM_TIMEOUT_MS`.

**2. Malformed-body guard (secondary bug)** — `src/server.ts`
Reproduced locally: `POST /v1/messages` with `{}` → `TypeError: Cannot read properties of undefined (reading 'find')` → 502. Root cause is outside this repo: acp-kernel's `conversationSignalAnthropic`/`conversationSignalOpenai` (`dist/wire/index.js:172/:406`) read `body.messages.find(...)` unguarded when computing the conversation fingerprint, and bili's pipeline gate only checks `typeof parsed === "object"` (so top-level arrays crash too). Bili now rejects parseable bodies missing the required `messages` array with a wire-native `400 invalid_request_error` before reaching the kernel; unparseable bodies keep their raw-forward behavior. Hard-failure logs now include a capped stack trace. A separate issue will be filed in acp-kernel for the missing guard there.

**3. Stale-install reminder dedupe (tertiary bug)** — `src/update.ts`
The `[update] running vOLD but vNEW is installed — restart bili to activate` warning repeated identically every 180s check cycle. It now logs once per version pair. Decision extracted to `reportNotNewer()` (unit-testable) because `checkForUpdate`'s source-checkout guard blocks in-repo e2e testing of that path.

Not included (needs owner decision): in-band update notification and opt-in self-restart — feature territory touching the load-bearing updater; discussed in the issue thread.

## Tests
- new `tests/fix-806-request-watchdog.test.ts`: env parsing; silent-upstream wedge → 504 within budget; chunked stream longer than the budget survives (pins idle-not-total semantics); fast response untouched
- new `tests/fix-806-malformed-body.test.ts`: 400 shapes for both protocols; top-level array body; valid bodies reach upstream; non-JSON raw forward preserved
- extended `tests/update-restart-reminder.test.ts`: +4 cases for warn-once / re-warn-on-new-pair / info-path-key-reset / newer-latest

Pre-flight: `npm run typecheck` clean, `npm run build` ok, full suite green (1526 tests, 0 fail, 2 skipped = ACP_TEST_E2E-gated).


ranxianglei:

[bot] 🏷 Reviewing this PR now. Plan: verify each of the three bugs against the code paths (not just the description), check diff cleanliness, run `typecheck` / full test suite / build on the branch head, and confirm CI status. Findings will follow in this thread.

ranxianglei:

[bot] 🏷 **Review complete — verified, mergeable.** I did not take the bug reports on faith; each was checked against the code, and the pre-flight was re-run locally on the branch head (`674a89f`).

## Diff cleanliness (first gate)
Clean. Exactly 5 files, +576/−11, all on-topic: `src/server.ts`, `src/update.ts`, 2 new test files, 1 extended test file. No lockfile drift, no version bump (version-guard CI confirms), no unrelated reformatting. Every line serves one of the three fixes.

## Pre-flight (re-run locally on branch head)
- `npm run typecheck` — clean
- `npm test` — **1526 tests, 0 fail, 2 skipped** (ACP_TEST_E2E-gated) — matches your reported numbers exactly
- `npm run build` — ok
- GitHub CI on head: all 7 green, including **real codex vs deterministic fake upstream** (this is the mandatory e2e for request-pipeline changes per AGENTS.md, so the accept-path change has real-client coverage despite the local sglang upstream being down)

## Bug-by-bug verification

**1. Request-scoped idle watchdog** — the structural defect is real and correctly diagnosed: `fetchWithTimeout` arms its deadline *inside itself* (`src/fetch-util.ts:91`), so a request that never reaches `fetch` has no deadline anywhere — exactly the wedge signature in #806. Verified against the implementation:
- Armed at accept (first line of the `createServer` handler), so prepare/preflight/lock-hold/forward all sit inside the deadline.
- Idle-not-total math holds: worst-case legitimate client-facing silence is one upstream LLM segment bounded by `upstreamTimeoutMs()`, and the budget is 2× that — and because both derive from `BILI_UPSTREAM_TIMEOUT_MS`, operators who raise the upstream timeout for slow local-model prefills scale the watchdog proportionally instead of getting false kills.
- The abort-on-fire is load-bearing and correctly wired: controllers registered at both `clientAbort` sites (`forward()` and `preflightCompressIfNeeded()`), and lock release happens in the outer `finally` (`src/server.ts:1816-1818`), so an aborted fetch promptly frees `withSessionLock`/`acquireInFlight`. Without the explicit abort, closing `res` alone would not cancel the fetch.
- Fire path handles both cases sensibly: pre-headers → `504 gateway_timeout`; mid-stream → truncate (best available once headers are committed). Double-fire and late-write races are guarded (`timer = undefined` first, `writableEnded/destroyed` checks).
- The `res.write` patch is transparent (bound original, same boolean return → backpressure intact) and nothing else in the codebase patches or bypasses it (checked for socket-level writes too).
- Test coverage pins the right semantics: silent-upstream wedge → 504 within budget, and the 3.2s-stream-vs-1.5s-budget case proves idle-not-total.
- One thing I checked that you didn't list: **blind CONNECT tunnels** go through a separate `connect` handler (`src/mitm.ts:94`) outside the watchdog — but their bili-controlled phase (establishment) already has a 15 s deadline (`src/mitm.ts:152-159`), and afterwards bili is a transparent pipe, so there is no uncovered wedge class.

**2. Malformed-body 400 guard** — reproduced the root cause in the installed kernel: acp-kernel 0.0.69 `dist/wire/index.js:172` (`conversationSignalAnthropic`) and `:406` (`conversationSignalOpenai`) do `body.messages.find(...)` unguarded — your line references are exact. Two things worth recording:
- I verified the guard's blast radius: `protocol` is derived from URL suffixes only (`/messages` → anthropic, `/chat/completions`|`/llm_raw_chat` → openai, `src/server.ts:1075-1085`), so embeddings, `/responses`, passthrough, and GET/model-discovery traffic are untouched; unparseable bodies keep raw-forward (preserving #619 semantics). 400 envelopes match each wire's native error shape.
- **Master moved while this PR was open**: `acp-kernel` bumped 0.0.69 → 0.0.72. I pulled 0.0.72 and confirmed the unguarded `find` is *still there* — so the bili-side guard remains load-bearing post-merge, and the kernel-side fix (acp-kernel#298, PR #299) is still open. Good thing this wasn't left to the kernel alone.

**3. Stale-install reminder dedupe** — `reportNotNewer()` preserves the original branch logic exactly (verified line-for-line against the removed block), dedupes warn per version pair, re-warns on a new pair, resets the key on the info path. The four added unit tests cover all of that. Note: #806 §3 ranked in-band client notice and opt-in self-restart *above* mere dedupe — you correctly parked those as feature territory needing an owner decision rather than bundling them into a load-bearing-updater fix. That decision item stays open on #806.

## Nits (non-blocking, no action requested)
- The 504 body is OpenAI-shaped (`{error:{type:"gateway_timeout"}}`) even when the request was on the Anthropic wire. For a transport-level 504 clients treat it generically and retry pre-first-byte, so this is cosmetic — flagging only for completeness.
- Residual (also yours to decide, not a blocker): if the watchdog fires while a request is still queued on the session lock (no controller registered yet), `handle()` continues after the client got its 504 and burns one upstream round-trip against a dead `res` before the fetch's own idle timer bounds it. Bounded and rare (requires another request wedged ≥24 min on the same session) — acceptable; a `res.destroyed` check after lock acquisition would close it if you ever care.

Nothing required changes on the branch — I found no defects to fix directly.

---
**中文摘要**:三个 bug 均已对照代码逐条核实(主 bug 的结构性缺陷、kernel 侧未加保护的 `messages.find`、提醒重复),本地重跑 typecheck/1526 项测试/build 全绿,CI(含真实 codex e2e)全绿,diff 干净无跑题改动——可以合并。遗留一项待你拍板:#806 §3 中"带内更新提示 / 可选自重启"两个升级方案是否做。

ranxianglei:
遗留一项待你拍板:https://github.com/ranxianglei/billion-context/issues/806 §3 中"带内更新提示 / 可选自重启"两个升级方案是否做。单独提issue

ranxianglei:

[bot] 🏷 已单独立项:**https://github.com/ranxianglei/billion-context/issues/811** — "feat(update): surface stale-install state to the human + opt-in self-restart (#806 §3)"。

查重过:现有 issue 里只有已关闭的 #328(最初的 stale warn)和 #17/#145(提醒去重,#808 已做),没有任何条目跟踪这两个升级方案,不算重复。

issue 里带了我的明确建议,供你拍板:
1. **先做 opt-in self-restart**(`--auto-restart-on-update`,default off)——它消灭的是真实故障模式(stale code 无限期服务);可行性你已在 #806 调试中验证过(单端口 re-bind 干净、客户端自动重连、sessions 磁盘持久化不丢)。安全门:zero-in-flight 判定 + re-exec 前 sanity check + 幂等防循环;且按 AGENTS.md §5,`src/update.ts` 改动必须先走一个 no-op release 验证升级链路。
2. **配最便宜的可见性**:Web UI badge + 现有一次性 warn log;**不做往 LLM stream 注入文本**(会污染本包要管理的 context)。完整的带内提示等 band(header / agent-plugin / web UI)选定后单独跟进。

未开始实现——等你在 #811 里选定范围再动手。PR #808 本身维持 review 结论:可以合并。

---
**中文摘要**:已把 #806 §3 两个遗留方案(带内提示 / 可选自重启)单独立为 issue #811,附查重结果与"先自重启(default-off)+ Web UI badge、不做 in-stream 注入"的建议,等你拍板范围;#808 可合并的结论不变。

PR #818, opened 2026-09-15
fix: preserve Responses tool images and retry interrupted summaries

### LINKED ISSUE #817: Responses tool images falsely overflow preflight; transient summary failures lock the session
## Problem

Responses `custom_tool_call_output.output` and `function_call_output.output` can contain `input_text` and `input_image` parts. The current codec stringifies the whole array into CoreMessage.text, while the host's image counting and summary placeholders only inspect `.content`. This turns image base64 into hundreds of thousands of text tokens and can reject a valid request before forwarding it.

A local reproduction with two 2048x1152 PNGs produced 558,772 text tokens and zero image tokens for a tool output, versus 1,108 text tokens and 4,250 image tokens for equivalent message content. The problem exists in the installed 0.1.113-based build and current master with acp-kernel 0.0.72.

## Failure recovery

Once preflight starts, HTTP 200 followed by a socket disconnect during response.text() throws TypeError: terminated outside fetchWithRetry's retry loop. One temporary failure therefore arms a five-minute dead-end cooldown. The cooldown key contains only model and window, so changed request content is also blocked. The original error cause is lost.

The image accounting defect and response-body failure are independently reproduced. This report does not claim that a particular image payload caused the historical connection failure.

## Reproduction

1. Return a mixed text/image content array from a Responses tool; use a synthetic PNG header plus base64 padding.
2. Convert it and estimate text/image tokens: the image becomes text and the image floor is zero.
3. Use a local summary endpoint that sends HTTP 200 and one SSE delta, then destroys its socket: there is no body-read retry.
4. Repeat an over-window request after a zero-progress failure, changing content without changing model/window: the old cooldown is reused.

## Proposed fix

- Keep tool image parts on the wire, project only text for counting, and include image parts in image estimates/placeholders.
- Preserve existing codec-derived IDs and structured outputs through both projection and fallback rebuilds.
- Share one bounded attempt budget across HTTP status and transport/body failures; preserve safe error-stage/code diagnostics and respect cancellation.
- Do not classify temporary upstream failures as content dead ends; scope deterministic cooldown to the actual prepared request body.

All reproduction fixtures are synthetic; no user screenshots, credentials, or conversation transcripts are included.

z0d0dOr:
Fixes #817

Responses tool outputs containing images were converted into base64 text, falsely pushing ordinary screenshot requests over the preflight window. Summary body-read failures could then leave the session in a five-minute cooldown, including after the request changed.

This change projects only text from image-bearing tool output arrays while preserving the original IDs and multimodal wire payload. Image counting and summary notes now also inspect tool outputs, and both Responses rebuild paths preserve image parts when text changes. Ordinary JSON tool outputs retain their previous behavior.

Summary requests now share one bounded retry budget across HTTP and transient transport failures, including body consumption. Cancellation and decoding errors are not retried blindly, diagnostics retain safe stage/code information, and transient failures do not arm a content-dead-end cooldown. Deterministic cooldowns include a digest of the prepared request instead of matching only model/window.

Validation:
- Typecheck and build passed.
- 28 focused regressions passed, including proxy/plugin forwarding, original ID stability, both rebuild paths, image-only notes, socket interruption, shared retry budgets, cancellation, gzip diagnostics, changed-content cooldown, and transient HTTP recovery.
- Real Codex CLI with a local fake upstream: 2/2 passed (overflow compression with retained sentinels; under-window control).
- Windows full suite: 1,631/1,637 initially passed. Certificate-directory and random restricted-port failures passed in isolated reruns. The remaining two platform tests (symlink expectation and CONNECT failure expectation) also fail on clean upstream master cfdfc52; they are not changed here.

No dependency/version changes. Previous compaction normalization (#770), Codex headroom (#799), and ordinary empty-summary cooldown behavior remain covered. The original image-accounting and transport failures were reproduced separately; this PR makes no unsupported claim about the cause of a historical upstream disconnect.

### Base branch refresh

Merged upstream master `3b46bdf` (release 0.1.115; acp-kernel 0.0.73) into this branch without conflicts. These version/dependency changes come from the base branch; the PR still changes only the original nine source/test files relative to master.

Fresh validation with acp-kernel 0.0.73: typecheck and build passed; 38 focused regressions passed; real Codex with the local deterministic upstream passed both overflow/compression and under-window control tests. The running local patched proxy was not reinstalled or restarted as part of this PR-only branch refresh.


ranxianglei:
继续

ranxianglei:

[bot] 🏷 Review of #818 (fixes #817) — checked out `z0d0dOr:2026-09-15_fix-multimodal-preflight-recovery @ 321fe52`, analyzed the diff line-by-line against the kernel wire codec, and ran full validation locally. **Verdict: approve, ready to merge.**

**Root causes confirmed against code (not just the report):**
1. Kernel `responsesToCore` stringifies array tool outputs (`wire/index.js`: `typeof output.output === "string" ? output.output : JSON.stringify(output.output)`) — so image base64 was charged as text tokens exactly as described in #817.
2. Old preflight called `response.text()` *outside* `fetchWithRetry`, whose loop only retries non-ok HTTP statuses (`src/fetch-util.ts:293-315`) — a 200-then-disconnect fell through untried and armed the model+window-keyed cooldown. Both claims check out.

**Image projection (new `src/responses-tool-output.ts`):**
- Verified kernel `coreToResponses` is stateless per message, so the per-message wrapper calls are equivalent to batch calls.
- Verified `patchResponsesInput` returns unchanged slots by object reference (so `byOriginal` identity lookup hits) and fresh objects for changed slots (covered by the `call_id` fallback); inserted messages also flow through the final map. No path escapes the rebuild.
- Ordinary string/JSON outputs remain byte-identical to kernel behavior (explicit regression test), codec-derived message ids are preserved, and image-only outputs get empty text instead of vanishing. All forwarding paths (main pipeline + forge) use the wrappers; grep confirms no remaining direct kernel importers.
- Token accounting (`image-tokens.ts`) and summary notes (`image-note.ts`) now see tool-output images consistently with the projected text.

**Preflight retry/cooldown:**
- One shared bounded budget across HTTP and transport failures including body consumption; cancellation correctly wins over transient codes (cause-chain walk depth-limited, diagnostics limited to name+code — the secret-leak test is convincing). Timer is cleared before backoff sleep (asserted via `_liveUpstreamTimersForTest`).
- Cooldown key now includes a sha256 digest of the prepared request body: changed content bypasses the dead-end (tested), identical retries still fail fast (#726 protection preserved, marker-format test updated accordingly), and transient failures (5xx/429/captcha-400/transport) no longer arm a content-dead-end cooldown (tested for both 503 and 400-captcha).
- Aborted path returns `respond:false` without arming anything; worst case for a timeout-induced body stall is a single 502, never a 5-minute block.

**Validation run here:** typecheck ✅ · build ✅ · full suite **1641 pass / 0 fail / 2 skip** (skips are the default-gated e2e). Real-codex e2e could not be executed in this environment (the default local upstream is down, and the purity phase needs a model-cooperative Responses upstream), so I'm relying on your reported Codex CLI 2/2 fake-upstream results plus the new host-level tests, which drive real HTTP through `startServer()` in both proxy and plugin modes.

**Minor notes (no changes required):**
- `src/preflight.ts:840-841` generic branch logs "summarization failed unexpectedly" without details. Currently unreachable for `requestSummaryBody` errors (everything is wrapped into `SummaryTransportError` first), so it's a pure safety net — fine as-is.
- Timeout-triggered body stalls classify as aborted (not retried). Conservative, and the no-cooldown fallback makes the blast radius one request — acceptable trade-off.

Diff is clean: 9 files, every hunk serves one of the two fixes, no dependency/version/generated-file churn.

中文摘要：修复了 Responses 工具输出中图片被 JSON 字符串化导致误判超窗、以及摘要请求传输中断后误触发 5 分钟冷却的问题（图片按像素计费且保留原始多模态载荷、重试预算跨 HTTP/传输故障共享、冷却键加入请求体摘要），根因已逐行核实、全量测试 1641 通过，可以合并。

PR #842, opened 2026-09-16
fix(agent): gate opencode native route behind nativeActive (#840)

ranxianglei:
## What
#840 audit fix (a): gate the opencode native entry's traffic-touching route/setup creation on `nativeActive`, mirroring pi's gated fetch intercept (`src/agent/pi-native.ts`).

Under a `bili opencode` launch the launcher owns the proxy; the native entry previously still created its `http.request` route unconditionally. The idle route never rewrote anything (no origin to probe), but on HTTPS (cert-MITM) launches it fired one misleading `proxy unavailable — model requests go direct (uncompressed)` line even though the launcher's proxy was compressing fine.

With this change the overlap-mode export degrades to exactly the launcher-mode shape (`src/agent/opencode.ts`): no route → no probing, no rewriting, no warning; header stamping falls back to `/bili/`-URL or `BILLION_CONTEXT_PROXY` env detection, which is correct in overlap mode (both plugins stamp identical values — the known harmless redundancy of #840 gap 3).

Why not fully inert the export (which would also fix gap 3): that changes behavior for manual-`BILLION_CONTEXT_PROXY` edge users (they currently keep native tools via env detection) and exceeds the scope of (a) as scoped in the issue. Gap 3 stays deferred for owner decision.

## Per-seam tests (owner constraint: per-seam e2e before merging)
- `tests/opencode-native-overlap.test.ts` — evaluates the REAL native entry under a launcher env (`BILLION_CONTEXT_PROXY` + `BILI_PROVIDER_REWRITES`): request reference untouched across hook fires, zero `proxy unavailable` / `bili-native` warnings, launcher-mode header stamping intact (env-detection and `/bili/`-wrapped URL cases). **Regression proof: with the source fix stashed, this file FAILS (the spurious warning fires); with the fix, green.**
- `tests/pi-native-overlap.test.ts` — real pi native entry under both launcher-env variants: global fetch NOT patched (intercept flag absent, fetch identity unchanged), no native host marker, plugin export unconditional. Codifies the hands-on pi-seam verification (passes pre-fix too — pi's seam was already correctly gated).

## Hands-on verification (real binaries, local upstream)
- **pi seam**: overlap run (native plugin installed + `bili pi`) — single proxy instance, compression active (prefix-affinity chain advancing, high cache-hit ratio), zero `bili-native:` lines in any log. Confirms the code-read inference: pi overlap is fully silent (fetch intercept never installed).
[...]

PR #824, opened 2026-09-15
feat(opencode): native plugin mode for OpenCode 2.x — self-spawning proxy (pi pattern) + #819 node-runtime fix

ranxianglei:
Implements the opencode line of #820 (native-mode umbrella), per the #809 design. Modeled on pi-native (#706) so both hosts share one bootstrap core.

**Stacked: base is `feat/pi-native-plugin`, not master.** This branch reuses #706's `native-intercept.ts` foundation (not yet on master) — merge after #706.

## What

OpenCode 2.x now runs billion-context as an ordinary plugin — no `bili` launcher, no daemon:

1. `bili plugin install opencode` writes a wrapper dir `<configDir>/plugins/billion-context/index.js` re-exporting `dist/agent/opencode-native.js` (directory form is mandatory for OpenCode 2.x plugin entries, #754), keeps writing the MCP shell when a live proxy origin exists (now optional — missing origin skips it with a note instead of failing), and sets `compaction.auto=false` merge-preserving, restoring from `.bili-bak` on remove (single compression owner).
2. On load the plugin bootstraps its own proxy via `ensureProxyRunning` (spawn-in-plugin, same as pi-native).
3. The `http.request` hook rewrites model-API URLs to `<proxy>/bili/<url>` by replacing the request reference (`Request.url` is readonly, #810). Liveness is probed against `GET /__bili/health` before every send; dead proxy → respawn (15s cooldown); failed respawn → one warning + direct sends, auto-recovering on later requests.
4. The V2 plugin factory is extracted into `src/agent/opencode-v2.ts` (`createOpencodeV2Setup`, parameterized by an optional per-request `route(e, state)` hook); `opencode.ts` keeps its V1+V2 export shape unchanged.
5. Shared side-effect-free bootstrap helpers moved to `src/agent/native-bootstrap.ts` (used by pi-native and opencode-native); importing it never spawns anything.

## #819 fix included

`ensureProxyRunning` spawned `process.execPath` unconditionally — under a non-Node host process (e.g. the Go-based opencode binary loading the plugin) that spawns the host binary instead of Node. New `resolveNodeRuntime()`: live node → `BILLION_CONTEXT_NODE` override → PATH search → actionable error; injectable via `LauncherDeps.nodeRuntime`.

## Coexistence with standalone extensions (#820 follow-up)

[...]

PR #850, opened 2026-09-16
fix(agent): restore /acp status command under OpenCode V2 (#809)

ranxianglei:
## What & why
Restores the `/acp` status command for OpenCode 2.x native mode.

Root cause: `/acp` was implemented only against the legacy `.server()` plugin API (`src/agent/opencode.ts`). The V2 migration makes opencode load plugins as `{id, setup}` and never call `.server()`, so `/acp` was orphaned — it never registered, and native mode looked "inactive".

This registers `/acp` in the shared `createOpencodeV2Setup` factory so BOTH the launcher entry and the native entry surface it.

## API contract (verified against @opencode/cli v2.0.3 source)
- Commands: `ctx.command.transform(editor.add({ name, description, execute }))`; `execute(input)` receives `input.sessionID`.
- Panel render: `ctx.session.synthetic({ sessionID, text })` inserts a synthetic message that does NOT trigger a model turn (V2 replacement for the legacy ignored no-reply prompt).

`/acp` fetches `<proxy>/__bili/plugin/status?conversationID=<sid>&fallback=latest` and renders the panel (idle/error notice when no proxy). Inert-safe on surfaces without `ctx.command`; TUI-only.

Scope check: the legacy `.server()` registered ONLY `command["acp"]` (`src/agent/opencode.ts:98/:104`) — no other local slash commands exist in `src/agent/`, so `/acp` is the complete set of orphaned commands.

## Files
- `src/agent/opencode-v2.ts`: add `command` seam to V2PluginContext + V2CommandEditor types; register `/acp` after tool registration.
- `tests/opencode-v2.test.ts`: fake ctx gains `command.transform` + `session.synthetic`; fake proxy serves `/status`; 2 new tests (panel render, no-proxy notice).

## Verification
- typecheck clean
- full suite: 1654 pass / 0 fail / 2 skip (win32-only)
- build OK; dist markers confirmed in both agent entries
- clean single-commit delta vs `feat/pi-native-plugin` (merge-base 1d4f38c, +133/-1, 2 files)

Refs #809. (#809 intentionally left open per owner — baseline reference.)

Note: this is the /acp delta only. If you'd rather bundle it with the #842 coexistence gate (#840) before it lands on the umbrella, say so and I'll close this and re-cut from the updated branch.


ranxianglei:

[bot] 🏷 Review of the `/acp` V2 restore (commit `11a82d7`) — verified, ready to land.

**What I did:** pulled the branch (`opencode-native-plugin` @ `11a82d7`, identical to `origin/2026-09-15_opencode-native-plugin`), cross-checked the diff against the server contract, and re-ran the full verification gauntlet locally. This is a review of completed work — I did not redo it.

**Diff integrity** — matches your description exactly: single commit on baseline `1d4f38c`, **+133/-1 across 2 files** (`src/agent/opencode-v2.ts` +71/-1, `tests/opencode-v2.test.ts` +63).

**Contract cross-check (all correct):**
- Handler sends `conversationId=…&fallback=latest` (`src/agent/opencode-v2.ts:231`); server reads `params.get("conversationId")` (`src/server.ts:733`). Names match. *(Minor: the issue body writes `conversationID` with a capital ID — cosmetic typo in the writeup; the code itself is correct.)*
- Response shape `{ok, panel, error}` matches `handlePluginStatus` (`src/plugin.ts:521-537`): `{ok:true, panel,…}` on success, `{ok:false, error}` on unknown/no-session.
- Faithful port of the legacy `/acp` (`src/agent/opencode.ts:108/:116`): same status URL, same panel/idle/error branching, same `fetchProxyVersion` fallback. V2 renders via `ctx.session.synthetic` (non-model-turn), TUI-only, inert-safe through optional chaining when `ctx.command` is absent.
- `console.error` in the agent file is fine — `src/agent/*` runs in-host (client-side), not in the proxy process, and every sibling agent module already uses `console.error` (the "no console.error" rule targets server-side modules).
- Scope claim holds: `/acp` is the only local slash command registered under `src/agent/`; nothing else was orphaned by the V2 migration.

**Verification gauntlet (re-ran, all green, matches your numbers):**
- `typecheck`: clean
- `npm test`: **1654 pass / 0 fail / 2 skip** (win32-only)
- `build`: OK
- dist markers: `Show ACP status` present in both `dist/agent/opencode.js` (×2 = V2 `setup()` + legacy `server()`) and `dist/agent/opencode-native.js` (×1 = V2 only) — as expected.

**Landing path (your actual question):**
The umbrella `feat/pi-native-plugin` has advanced past `1d4f38c` to `ca4b1ab` (it already absorbed the earlier part of this branch via PR #824). I verified:
- Neither file you touched was modified by any commit in `1d4f38c..ca4b1ab` (empty diff-stat).
- Cherry-picking `11a82d7` onto the current tip `ca4b1ab` applies **with zero conflicts**, still +133/-1.
- The #840 coexistence gate touches disjoint files (`src/agent/opencode-native.ts` + 2 overlap-test files) — **zero overlap** with this delta.

**Recommendation: land `/acp` standalone as its own small PR against the current `feat/pi-native-plugin` tip.** Rationale: it's a self-contained single commit, rebases cleanly with no conflicts, has zero file overlap with the #842/#840 gate, and is independently green — so there's no technical coupling forcing a bundle. Keeping it separate means the coexistence gate can't block it (and vice versa), and each lands with a smaller, easier review. Bundling only pays off if there's a shared-file conflict or ordering dependency, and there isn't one here. (Final call is yours — happy to re-cut if you'd rather bundle.)

I have not pushed or opened anything — confirm standalone vs. bundle and I'll cut the PR accordingly.

中文摘要：独立复核了 `/acp` V2 恢复提交（`11a82d7`）——diff 与描述一致（+133/-1、2 文件），契约核对无误，typecheck / 测试（1654 通过 / 0 失败 / 2 跳过）/ 构建 / dist 标记全部复现通过；该单提交可零冲突落到当前 umbrella 尖端、与 #840 共存门文件零重叠，建议作为独立小 PR 合入而非捆绑。

ranxianglei:
继续

Issue #868, opened 2026-09-16
Preflight gives up on a payload far over the window although every fold succeeds


joshwapohlmann:
A conversation well past the context window fails compaction with a message that blames the fold budget, even though every individual fold worked:

```
context 1385572 tokens reached preflight target 1048576 … preflight compressing
preflight compressed 8 range(s), ~329022 tokens saved (1385572 -> 1056550)
preflight fail-fast 502 preflight_compress_failed: the compress budget was exhausted after 8 rounds
```

Two things are wrong here. The loop stops while ranges remain: it had folded 8 ranges to within 8k tokens of the window and then declined to fold the ninth. And the reported number is misleading, because the failure message quotes the original context rather than the reduced one, so a conversation 8k short of the target reads as if folding had failed entirely. The session then enters the preflight dead-end cooldown, so the same client retries later and fails the same way.

Cause: the preflight folds one range per round, and both per-invocation budgets (`MAX_PREFLIGHT_ROUNDS`, `MAX_SUMMARY_CALLS_PER_PREFLIGHT`) were 8. That is fine for a payload a little over the window and structurally impossible for one well beyond it: a payload at 1.39M tokens against a 1,048,576 window needs more than 8 range folds no matter how well each one lands.

Expected: the preflight keeps folding until the payload is under the target or no foldable range is left, and when it does give up, the failure states the reduced context and the remaining range count rather than the original size.

Fixed by PR #869.

joshwapohlmann:
The fix is already filed as PR #869 (branch `2026-09-16_868-preflight-round-budget`). Please review there rather than implementing a parallel change.

ranxianglei:
> 🤖 Powered by [ework](https://github.com/ranxianglei/ework) · qwen3.8-27b

Reviewed #869 on `2026-09-16_868-preflight-round-budget` (head `ea17a12a`, base `d650db7`).

**Verified**
[...]

PR #574, opened 2026-09-06
fix: preflight tries all viable ranges before declaring exhaustion (#569)

ranxianglei:
## What

Preflight fail-fast returned `502 retryable=false` with "no range could be compressed" while later spans of the same history were compressible — minutes later the session owner's in-band compress shrank it 88% (#569).

## Root cause

`preflightCompress` (src/preflight.ts) picked exactly ONE range per round — the oldest viable startRef — and when that single range yielded no applied chunk (sub-minCompressRange chunk, empty content, unusable summary, or kernel apply rejection), it declared **global** exhaustion and broke out of the round loop. Remaining viable ranges were never tried, so one bad leading edge produced a false terminal verdict on a still-recoverable session.

In block-dense histories this is likely: the kernel's min-gate sums only non-summary message text per spec, while preflight's char precheck counts all view text including block summaries — so a leading range can pass preflight's check yet be rejected at apply time, or simply sit below `minCompressRange` after its blocks were folded.

## Fix (src/preflight.ts)

- Iterate ALL viable ranges (oldest-first, prefix-cache-stable) within a round until one yields an applied chunk; declare `exhausted` only after every range has been tried and nothing applied.
- A range whose summary was unusable / apply failed / refs unresolvable is added to a skip-set (refs are stable across folds) so it is never retried within the invocation — no wasted upstream calls, no infinite churn.
- New cost cap `MAX_SUMMARY_CALLS_PER_PREFLIGHT = 8` bounds summarization calls per preflight invocation (legacy worst case was also 8×1); hitting it reports a distinct budget-exhausted detail instead of the generic one.
- Exhaustion detail now names how many ranges were tried, making the 502 truthful and diagnosable.
- Upstream-error (429→503 retryable), abort, and credit math semantics unchanged. No wire-shape change: mode-agnostic core only (both plugin and proxy modes affected identically, per AGENTS.md §6).

## Tests (tests/preflight-multi-range.test.ts, new)

E2E through the real proxy with a mock Anthropic upstream:
1. **Regression**: 3-segment over-window conversation where the oldest segment's summary is unusable → now 200 with the next range folded (legacy code: 502, verified RED via stash).
2. **Truthful exhaustion**: every summary unusable → 502 only AFTER all ranges tried (`summaryCalls ≥ 2`; legacy made exactly 1), no forward, no blocks.
3. **Budget cap**: 12 segments all unusable → exactly `MAX_SUMMARY_CALLS_PER_PREFLIGHT` summarization calls, budget variant in the error.

[...]

PR #319, opened 2026-08-28
feat: anonymous tail-window reattach + fork lineage for prefix affinity (#316 PR-B)

ranxianglei:
Model: glm-5.3

**Issue:** #316 (PR-B — anonymous tail-window reattach + fork lineage). **Independent of PR-A** (codex turn-metadata, #317) — does not depend on it; both can merge in either order.

## Problem
`src/prefix-affinity.ts` (#311) matches by *full-depth containment*: the stored chain's depth must appear wholesale at the incoming head. When a client **drops its oldest messages** (truncation — common on e.g. dsh web), the match fails and the request is orphaned into a fresh `pfa-` session, **stranding the old session's compression blocks** (they should reattach to the original session).

## Fix
`resolve()` now falls back to a **tail-window comparison** when the full-depth match fails: the incoming's leading items are matched against each stored chain's trailing items using **position-independent per-item hashes** (`sha256` of each canonical message — the progressive `chainHashes` can't match across a truncation because they depend on the full prefix). A **unique** hit reattaches the original session and lets the kernel reconcile by message id; multi-hit / zero-hit falls through to a new session as before.

Kernel safety is **zero-change** and empirically anchored by a new test: a truncated replay against block-bearing state does **not** throw, deactivates the affected block gracefully (`active → false`), and revives it when the full history returns (`survivedCount++`).

New `pfa-` sessions also record a **best-effort lineage** (discarded candidates + reason `truncated`/`forked`) in `session.metadata.anonymousPrefixAffinity` — **UI/debug only, never used for matching**. Live-edge overwrite semantics ("incoming history is the truth") are unchanged: record only, don't arbitrate.

## Changes
- `src/prefix-affinity.ts`: new consts `TAIL_WINDOW=8`, `MIN_TAIL_MATCH=3`, `MAX_STORED_ITEMS=64`, `MIN_FORK_PREFIX=3`; helpers `perItemHashes` + `lcpLength`; `resolve()` gains the tail-window reattach + lineage pass; `note()` stores per-item hashes (4th arg); `AnonymousAffinity` gains `via` (`prefix`|`tail-window`|`new`) + `itemHashes` + optional `lineage`.
- `src/server.ts`: a tail-window reattach log line; `note()` 4th arg; metadata records `via` + `lineage`.
- `tests/prefix-affinity.test.ts`: `note()` call sites updated to the 4-arg signature.
[...]

PR #869, opened 2026-09-16
fix(preflight): fold deep enough for a payload far over the window (#868)

joshwapohlmann:
## Problem

The preflight folds one range per round, with both per-invocation budgets at 8. A payload far beyond the window therefore fails even when every fold succeeds, and the failure reads as a total one:

```
preflight compressed 8 range(s), ~329022 tokens saved (1385572 -> 1056550)
preflight fail-fast 502 preflight_compress_failed: the compress budget was exhausted after 8 rounds
```

That conversation was folded to within 8k tokens of a 1,048,576 window and then declared exhausted. The message also quotes the original context, so a reader sees 1.39M tokens and no folding, and the session takes the preflight dead-end cooldown on top.

## Fix

`MAX_PREFLIGHT_ROUNDS` and `MAX_SUMMARY_CALLS_PER_PREFLIGHT` in `src/preflight.ts` are raised to 16, and the latter is exported so the budget-cap tests derive it instead of pinning a literal. Nothing else changes: a round still folds one range, and a payload already over 16 ranges wide still stops and reports, but at a depth that matches a window-sized payload rather than a marginally oversized one.

## Tests

`tests/preflight-round-budget.test.ts` drives the fold loop over an openai-wire fixture of roughly 83k tokens against a 30k window. It refuses the request at 8 rounds (`preflight_compress_failed`) and forwards it at 16, using more than 8 summary calls on the way, which pins the behaviour rather than the constant. The #574 cap test in `tests/preflight-multi-range.test.ts` now takes its expected number from the exported constant instead of re-pinning 8, so the two cannot drift apart again.

## Verification

Full proxy suite on the branch: 1669 tests, 1661 passing, the same six environmental failures upstream shows. `npx tsc --noEmit -p tsconfig.build.json` clean. The live case that produced the report now folds further and forwards instead of failing; the numbers above are from that session, and the test fixture reproduces the same shape at a smaller scale so the behaviour is pinned without needing a 1.39M-token fixture in the suite.


joshwapohlmann:
I would like both, in this order, and you are welcome to add them here.

1. **The honest failure detail.** Make the exhaustion path and `failFast` report the reduced post-fold token count and the remaining compressible-range count instead of the original `tokenCount` (`src/server.ts:2862` reading the value captured at `src/server.ts:2831`). That is the half that changes what an operator reads at 502 time, and it is the half of the issue's "Expected" this PR does not yet meet. A commit on this branch is the right place while it is unmerged.

2. **The budget: document, do not redesign yet.** If the coverage bound can be stated beside the constants (how far over the window 16 folds reaches, and that beyond it the loop still exits with ranges left), that is enough. A budget scaled to the overshoot is a design change I would rather see reviewed on its own, and your point that 16 is a number tuned to this incident is right.

Neither becomes an issue: the filing rule for this project is pull requests only, because an issue in either repository sets the review agent implementing in parallel — which is what happened on the kernel's #319. If you would rather do 2 as a design change, say so and I will file it as its own PR instead.

Thanks for confirming the root cause and for deriving the test from the exported constant instead of pinning `8`.

ranxianglei:

[bot] 🏷 Both items are done and pushed to your branch `joshwapohlmann/billion-context@2026-09-16_868-preflight-round-budget` (`ea17a12..0a16f29`) as two commits. (Heads-up: my session hit a host-level disk-full mid-task earlier today — I posted a blocked status; everything below is now verified.)

**1. Honest failure detail — `e05894e fix(preflight): report the post-fold size and remaining ranges in the fail-fast 502`**
- `PreflightResult.rangesRemaining` (`src/preflight.ts`): compressible ranges still visible in the kernel's final view after the walk stopped, captured each round from the viable ranges.
- The fail-fast 502 (`src/server.ts`, the closure you pointed at) now reads, when folding happened:
  `context ~<post-fold> tokens (down from ~<original tokenCount> before preflight) exceeds the model window <limit> (model=<m>), with <N> compressible range(s) still visible and preflight compression should not have been needed…` — i.e. for the filed incident it would say `~1,056,550 tokens (down from ~1,385,572 before preflight) … with <N> compressible range(s) still visible … exhausted after 16 rounds`. Post-fold size is the loop's own `currentTokens` (same number the existing "tokens saved (A → B)" log prints). When no fold ran, the original phrasing is kept verbatim, so no existing message shape regresses; the exhaustion detail strings are untouched, which keeps the `/summarization budget/i` and `/no range could be compressed/i` pins green.
- `MAX_PREFLIGHT_ROUNDS` is now exported next to `MAX_SUMMARY_CALLS_PER_PREFLIGHT`, so tests derive budgets from constants instead of pinning literals.
- New e2e case in `tests/preflight-round-budget.test.ts`: the fixture scaled past the bound (400 turns, ~340k tokens vs a 30k window) → asserts 502 + "exhausted after 16 rounds" + "down from … before preflight" + "… compressible range(s) still visible" + exactly one summary call per round + nothing forwarded. One matching assertion added to the #574 budget-cap case in `tests/preflight-multi-range.test.ts` (zero-fold path).

**2. Coverage bound documented, not redesigned — `0a16f29 docs(preflight): state the fold-depth coverage bound beside the budgets`**
Comment beside the constants in `src/preflight.ts`: one range per round × ≤ `CHUNK_FRACTION`×window removed per fold ⇒ 16 rounds cover at most ~9.6×window of overshoot (payload up to ~10.6×window, best case); real coverage lower (sub-budget ranges save less, #726 halving can spend several calls on one range); beyond the bound the loop exits cleanly with the post-fold size + remaining-range count reported. It also states plainly that 16 is tuned to the #868 incident class and that an overshoot-scaled budget is left as a separate design question — I did not touch the mechanism. No new issues or PRs filed, per your rule.

**Verification:** `npm run typecheck` clean; full suite 1651 tests / 0 failures / 2 skipped (the gate-only e2e) — the 6 environmental failures from your run did not reproduce here; `npm run build` clean. PR: https://github.com/ranxianglei/billion-context/pull/869

中文摘要：按你的顺序完成了两项——502 报错现在报告折叠后 token 数（含折叠前原值对比）与剩余可压缩区间数，覆盖边界文档写在常量旁而未改动预算机制；两个提交已推送到你的分支，typecheck、全量测试（1651 通过 0 失败）和构建均已验证。

ranxianglei:

[bot] 🏷 Correction to my previous comment: I garbled the quoted message format. The exact shape of the new 502 body (for the filed incident's numbers) is:

`context ~1,056,550 tokens (down from ~1,385,572 before preflight) exceeds the model window 1,048,576 (model=<m>), with <N> compressible range(s) still visible and preflight compression could not bring it under: the compress budget was exhausted after 16 rounds. The over-window payload was NOT forwarded.`

Everything else in the report stands unchanged.

ranxianglei:
解决冲突

ranxianglei:
继续

ranxianglei:

[bot] 🏷 Conflict resolved and pushed. The branch `joshwapohlmann/billion-context@2026-09-16_868-preflight-round-budget` now has master (up to `55da3f0`, release v0.1.117) merged in as one commit `eb6c475`, and GitHub reports the PR **mergeable / clean**.

**What conflicted.** A single hunk in `src/preflight.ts` (the end of `preflightCompress`). This branch inserts its new `result.rangesRemaining = rangesRemaining;` line directly after `if (result.compressedRanges > 0) deps.session.stats.lastInputTokens = currentTokens;` — and master's #857 (`cd0cd88`) rewrote exactly that line into the image-floor-aware baseline block. Everything else auto-merged, including `src/server.ts` (the fail-fast message change lands cleanly on top of master's changes).

**Resolution.** Kept master's #857 block verbatim and appended this branch's `rangesRemaining` assignment after it. Rationale: the old one-liner was base context that #857 deliberately superseded (never persist the image floor into the usage baseline; raise-only with `lastInputTokensSource = "estimate"`), not a change made by this PR — so the union of intent is "master's baseline semantics + this branch's new field". No behaviour from either side was dropped.

**Verification on the merged head (`eb6c475`):**
- `npm run typecheck` — clean
- `npm test` — 1672 tests, 1670 passing, 0 failing, 2 skipped (the gate-only e2e)
- `npm run build` — success
- GitHub API: `mergeable: true`, `mergeable_state: clean`

No other commits were touched; the five branch commits keep their hashes, so everything in my earlier report still applies. Ready for you to merge when you are.

中文摘要：解决了分支与 master（至 v0.1.117）在 src/preflight.ts 末尾的唯一冲突——保留 master #857 的基线逻辑、追加本分支新增的 rangesRemaining 赋值，合并提交 eb6c475 已推送；typecheck/全量测试(1670 通过)/构建全部通过，PR 现显示可干净合并。

Issue #870, opened 2026-09-16
Plugin passthrough leaves the client an empty turn when the model echoes a render tag


joshwapohlmann:
The plugin heads every user and assistant text block with a ref annotation (`renderTags=text-only`), and the model sometimes echoes that annotation line as its own visible content. Where the echo is the entire visible content of a turn, the tag filter deletes it, the emptied chunk is suppressed, and the client receives a turn with no text and no tool call, which the harness aborts:

```
✘ upstream stream ended before a completion event; this turn may be incomplete
```

Releasing the echo instead is not acceptable: #463 keeps a turn that is a pure tag echo swallowed, and #674 plus the degenerate-turn warning exist precisely so that imitated render tags never reach the client. So the turn has to be made usable without leaking the annotation.

Expected: when a turn ends cleanly (`stop` / `end_turn` / `stop_sequence`) with zero visible text and no tool call, the pipe should obtain a usable continuation rather than hand the client an empty turn. Constraints that follow from the existing contracts: at most one retry, the retry must respect the wire's framing (one `message_start` / `message_stop` on anthropic, retry block indices offset, the openai terminal being the `finish_reason` chunk rather than `[DONE]` so the first attempt's terminal does not arrive twice), and if the retry is also empty the original empty turn must still pass through with the existing `[degenerate-turn]` warning rather than silently vanishing.

Fixed by PR #871.

joshwapohlmann:
The fix is already filed as PR #871 (branch `2026-09-16_870-plugin-empty-turn-retry`). Please review there rather than implementing a parallel change.

ranxianglei:
> 🤖 Powered by [ework](https://github.com/ranxianglei/ework) · qwen3.8-27b

Reviewed #871 as requested (rebased head `8d0a089`, onto current master) — no parallel change implemented. **Verdict: ready for merge** (human-side, per repo rules).

- Traced the pipe against every constraint in this issue: one-shot retry gate placed after event processing (coalesced content+finish_reason chunks count their text first), clean-terminal whitelist (`stop`/`end_turn`/`stop_sequence` only), OpenAI terminal = finish_reason chunk so the first attempt's `[DONE]` never doubles, Anthropic single `message_start`/`message_stop` with retry blocks index-offset after the client's own, and retry-also-empty falls back to the original empty turn with the existing `[degenerate-turn]` warning. All hold; truncation (#721) survives takeover too.
[...]

PR #867, opened 2026-09-16
feat(proxy): serve the Google/Gemini wire (#866)

joshwapohlmann:
## Solution

Native support for the Google/Gemini wire in the proxy, so a Gemini conversation is tracked, folded and re-emitted instead of passing through verbatim.

Detection and identity: `googlePathKind(urlPath)` and `googleModelFromPath(urlPath)` classify `:generateContent`, `:streamGenerateContent` and `:countTokens` and resolve the model, `WireProtocol` gains `google`, and `requestModel` prefers the body's model with the path as the fallback. Anonymous Gemini traffic is linked into a session by prefix affinity over `contents` (`conversationSignalGoogle`), since the wire carries no client session header.

Folding: `prepareGoogle` / `prepareGoogleCountTokens` run the kernel path with `strategy` and tool injection (`injectGoogleTool`), `appendGoogleNudge` folds a nudge into a trailing non-model content rather than adding one, `pickAdapter` gained a `google` entry, and `src/loop/adapter-google.ts` plus `src/stream-google.ts` (`rewriteGoogleJsonResponse`) serve the loop and non-streaming paths. `googleThoughtSignature` is written back on replayed reasoning and carried through `ToolCallEmit`.

Accounting: `usageTotals` / `promptInputTotal` / `usageOutputTotal` map input = `promptTokenCount` minus `cachedContentTokenCount` and output = `candidatesTokenCount` plus `thoughtsTokenCount`; the window table gained the Gemini 3 line; `outputBudgetField` gained `generationConfig.maxOutputTokens` with matching read/write helpers; overflow detection understands the Gemini token phrasing.

Auxiliary paths: tag stripping, image estimation (`inlineData` / `fileData`), strip-images, in-band truncation frames, the fake-completion hint, the preflight summary payload and the plugin's usage sniffing all learned the wire.

## Two defects fixed on the way, both Gemini-only in practice

- The preflight summary was posted to the client's own URL. That endpoint is `:streamGenerateContent`, which answers SSE whatever the request asks for, while the summary parser only engaged for a streaming client, so the reply read as empty: `carried 41 SSE event(s) but no summary text`, eight attempts, then `preflight fail-fast 502` and a 300s dead-end cooldown. The body's shape now decides the parser (`SSE_DATA_LINE_RE` plus a JSON-vs-SSE branch).
[...]

Issue #866, opened 2026-09-16
Google Gemini is untracked: no conversation id, so nothing is folded or compressed


joshwapohlmann:
bili terminates TLS for `generativelanguage.googleapis.com` and forwards the POSTs, but a Gemini conversation is invisible to everything downstream of detection. Every turn is logged as an anonymous forward while the other wires log a session-scoped `processTurn`:

```
[01a0a0cb-…] processTurn: 158 msgs …        (deepseek)
[unknown] forward POST → https://generativelanguage.googleapis.com/v1beta/models/gemini-3.8-flash:streamGenerateContent?alt=sse
```

Consequences observed on a live session: no conversation id is minted, prefix affinity never links the turns into one session, the nudge never computes growth, ACP compression never fires, and usage is not accounted. The client's own compaction is cancelled by the plugin by design (`session_before_compact` returns `{cancel:true}` for reason `threshold` or `overflow`), so with the proxy also unable to compact, nothing compacted and the context reached 339k tokens, holding large tool results (135 KB DOM dumps) verbatim.

Cause: protocol detection matched only `/chat/completions`, `/v1/messages` and `/responses`. Gemini's `/v1beta/models/<model>:streamGenerateContent` matched none of them, so `protocol` came back `null`, the kernel path was skipped, and the request was forwarded byte for byte. Client identity cannot be taken from a header here either: the client stamps a session header only for the anthropic and openai wires, so a Gemini session has to be recognised by prefix affinity over `contents`.

What has to be supported for this wire to behave like the others:

- Detection and model resolution from the path (`:generateContent`, `:streamGenerateContent`, `:countTokens`), and a window entry for the Gemini 3 line.
- A conversation identity for anonymous Gemini traffic, by prefix affinity over `contents`.
- Folding and re-emission on that wire, keeping `systemInstruction`, the alternating roles, `thoughtSignature` round-trips and named `functionResponse` pairing valid, plus tool declarations in the flat `{name, description, parameters}` shape.
- Usage accounting from `usageMetadata` (input = `promptTokenCount` minus `cachedContentTokenCount`, output = `candidatesTokenCount` plus `thoughtsTokenCount`), and the output budget mapped to `generationConfig.maxOutputTokens`.
- The auxiliary paths: tag stripping, image estimation, in-band truncation, the fake-completion hint, `countTokens`, and the preflight summary call.

[...]

Issue #732, opened 2026-09-12
[BUG] In some cases, with Codex's auto model + DeepSeek, it will suddenly terminate automatically.


BsoBird:
log
```
2026-09-12T14:43:08.075Z [info] forward POST → https://api.deepseek.com/responses
2026-09-12T14:43:22.209Z [info] [01a09604-f526-7822-bd15-37389368b47d] [acp-usage] round 1 input=86426 cached=84864 (cache hit 98%)
2026-09-12T14:43:22.210Z [info] [01a09604-f526-7822-bd15-37389368b47d] [acp-proxy: compress requested 1 range(s): m00109–m00241]
2026-09-12T14:43:22.210Z [info] [01a09604-f526-7822-bd15-37389368b47d] [acp-proxy: ctx has 150 message(s), state has ? ref(s) mapped]
2026-09-12T14:43:22.210Z [info] [01a09604-f526-7822-bd15-37389368b47d] [acp-proxy: first msg ids: acp_summary_b1(4061c), h_04717dfa80527a27(874c), h_c409b04ed442fbde(57c), acp_summary_b2(3581c), acp_summary_b3(11663c), h_111d38449f61bea7(4469c), h_d15b859a5b434346(290c), h_7e894daa91504165(154c), h_acd8d70bbfb03b00(5254c), h_8ea31d5c56c4658a(154c)]
2026-09-12T14:43:22.215Z [info] [01a09604-f526-7822-bd15-37389368b47d] [acp-compress-obs] shrink 35% (~29987/86426 tok) foldPoint=m00109 blocks=1 anchor≈2479 tok (2 active blocks, sys excluded) postCtx≈56439 → next-request cache ceiling ≥4%
2026-09-12T14:43:22.215Z [info] [01a09604-f526-7822-bd15-37389368b47d] [acp-proxy: [Compressed m00109–m00241 → 1 block(s), ~29987 tokens saved.]]
2026-09-12T14:43:22.218Z [info] [01a09604-f526-7822-bd15-37389368b47d] [acp-loop] round 1: re-request view refreshed to post-compress fold
2026-09-12T14:43:22.218Z [info] [01a09604-f526-7822-bd15-37389368b47d] [acp-loop] round 1: proxy tool executed; re-requesting so the model sees the result
[...]

PR #871, opened 2026-09-16
fix(plugin): retry a turn whose only visible output was a stripped tag echo (#870)

joshwapohlmann:
## Problem

The plugin path strips render tags from every forwarded chunk. When the model echoes a ref annotation as the entire visible content of a turn, the tag is removed, the emptied chunk is suppressed, and the client receives a turn with neither text nor a tool call, which the harness aborts. Releasing the echo is not an option: #463 deliberately swallows a pure tag echo, and #674 plus the degenerate-turn warning exist so that imitated tags never reach the client.

The adapter path (proxy mode with the compress tool injected) already answers this with a one-shot re-fetch; the plugin's direct pipe had no equivalent, so the defect only bit plugin-mode sessions.

## Fix

On a turn that ends cleanly (`stop`, `end_turn`, `stop_sequence`) with zero visible text, zero tool calls and no error finish reason, the pipe now re-issues the same request once with a continuation nudge and relays the retry in the original framing.

- `src/degenerate-retry.ts` (new): `DEGENERATE_RETRY_NUDGE`, `injectContinuationNudge(protocol, body)` and `makeContinuationRefetch(opts)`, moved out of `src/loop/core.ts` so both paths share one nudge and one body builder.
- `src/wire-body.ts` (new): `appendTrailingUserText(protocol, body, text)`, extracted from the fake-completion hint, which now delegates to it.
- `src/plugin.ts`: `pipePluginChatWithStrip` takes a `refetch` callback, tracks `degenerateRetried` / `inRetry` / `retryIndexOffset` / `blocksForwarded`, and gates on a clean terminal reason set.

Two details the tests forced, both about not corrupting the client's view of the turn: the openai terminal is the `choices[].finish_reason` chunk rather than `[DONE]` (using `[DONE]` let the first attempt's terminal reach the client twice), and the gate runs after the event is processed, because a coalesced chunk can carry both content and the finish reason and would otherwise be judged empty and dropped. An anthropic retry lands at `"index":1` with a single `message_start` / `message_stop`, and an unissuable retry presents the original terminal with valid JSON.

If the retry is also empty, the original empty turn is passed through and the existing `[degenerate-turn]` warning fires, so the contract that keeps the annotation away from the client is unchanged.

## Tests

`tests/plugin-empty-turn-retry.test.ts` (8, against the pipe with a fake response and a fake refetch): retries once on an echo-only turn and the client sees the retry's prose with exactly one `[DONE]` and one `finish_reason`; no retry when visible text, a tool call or `content_filter` ends the turn; an empty retry passes the empty turn through and warns; an unissuable retry presents the original terminal as valid JSON; anthropic keeps one `message_start` / `message_stop` with the retry at index 1; and the nudge-body shapes per wire.

## Verification

Full proxy suite on the branch: 1669 tests, 1661 passing, the same six environmental failures upstream shows. `npm run build` clean (dist 2.89 MB), `npx tsc --noEmit -p tsconfig.build.json` clean. The branch is deliberately not part of the local installation, which carries the wire and the fold budget; the reproduction it answers was a live plugin-mode session whose turn ended with no text and no tool call.


joshwapohlmann:
Filed note on the Gemini half, deliberately not in this branch.

The commit that teaches the gate the Gemini terminal spelling (`candidates[].finishReason: "STOP"`) and gives the trailing-nudge writer a `contents[].parts[]` branch cannot typecheck on its own: with master's protocol unions (`"anthropic" | "responses"` in `src/plugin.ts`, `"anthropic" | "openai"` in `src/wire-body.ts`) those `=== "google"` comparisons are `TS2367` errors, and the union widening is what the Google wire PR (#867) does. So this branch is the retry fix for the wires master supports today, and the Gemini coverage belongs here as a small follow-up once #867 has landed.

Branch head is now `8d0a089`, rebased onto `d650db7`: `npx tsc --noEmit -p tsconfig.build.json` clean, `tests/plugin-empty-turn-retry.test.ts` 8/8. The diff is the three commits the PR describes; nothing else moved.

The other half of the same stall is a separate problem and a separate PR: a render tag wrapped around the model's whole turn now leaves an empty turn instead of orphan markup (#872), which is the shape this retry exists to re-ask.

joshwapohlmann:
Live verification in the reporter's own session, on the locally installed build of this change, after the proxy was restarted onto it at 18:04:48Z.

Counted per hour in that machine's `bili.log` for 2026-09-16, one line per event:

| hour | strips | retries | `[degenerate-turn]` warnings |
| --- | --- | --- | --- |
| 15:00Z | 156 | 0 | 110 |
| 16:00Z | 468 | 0 | 169 |
| 17:00Z | 1498 | 0 | 404 |
| 18:00Z | 98 | 2 | **0** |

`[degenerate-turn]` is the residual-stall signature: a turn reached the client with no visible text and no tool call.

- The retry fired twice in the 18:00Z hour, both in the reporter's own session, at 18:05:28.144Z and 18:09:34.174Z. No warning followed either, so both turns came back usable.
- Both echo shapes occur in that traffic and both recovered: 18:05:28.016Z stripped an unbalanced wrapped head (`tokens="1" text="text` with no close), 18:09:37.309Z stripped a balanced one (`tokens="1" text="text">`), which is what a plain tag echo leaves behind.
- One turn watched end to end: head stripped at 18:05:28.016Z, retry at 18:05:28.144Z, the retry's own request upstream at 18:05:32.477Z (`processTurn 516 msgs`, then the forward POST).
- A second live session was asked to stand in for the model and re-trigger the original stall deliberately; it recovered.

Before this went upstream, both shapes were re-proved against the installed bundle with a mock upstream: two upstream requests, the second carrying the continuation nudge, and a readable tool call reaching the client.

joshwapohlmann:
Adding the third pipe. The Responses passthrough had no gate at all, so an echo-only codex subagent turn still reached the host empty: 8 of one day's degenerate turns in a single session, 56 across all sessions.

One design note, because this pipe's terminal is not the chat wires' terminal. A Responses turn ends on a FAMILY — `response.output_text.done`, `response.content_part.done`, `response.output_item.done` — decided by `response.completed`, and every one of those events carries ids the client already holds. So the retry holds the family until the completion decides the turn: on a degenerate completion the held events *and* that completion are dropped, the agent's own body is re-issued with the same continuation nudge, and the retry is reframed onto the ids the client already holds — its `response.created` and its added events are dropped, and `item_id`, `output_index` and the completed output ids are rewritten to the first attempt's. A healthy turn releases the held events in place, so holding never reorders a turn.

Wired at all three Responses passthrough sites: plugin mode, the native-compact tee, and proxy mode without injection.

Tests: five in `tests/plugin-empty-turn-retry.test.ts` — the retry with the ids rewritten (one `response.created`, one `response.completed`, no id from the retry reaching the client), the healthy release order (`output_text.done` still precedes `response.completed`), a turn whose text arrives only in the done event (no retry, because the client would have received it), a function call, and a failed terminal. 13/13 in that file, `tsc --noEmit` clean. Commit `817a8cf`.

ranxianglei:

Reviewed the rebased head (`8d0a089`, onto `d650db7`) against the #870 contract. Verdict: **logic is correct, no regressions vs master, ready for human merge.** Details below.

## Verification (local, on rebased head)

- `tsc --noEmit` clean; build succeeds (dist bundle ~2.9 MB).
- Full suite A/B: PR head **1672 tests / 1669 pass / 1 fail / 2 skipped** vs master `d650db7` **1664 / 1661 / 1 / 2**. Delta is exactly the 8 new tests, all passing. The single failure (`#857 D: upward self-heal trusts usage baselines only`) fails **byte-identically on clean master** — pre-existing, unrelated to this change (also explains the red `test` jobs in CI on both master and this head; all non-test jobs are green here). One early run showed a second transient failure that did not reproduce in any later run (PR head ×2, master ×1).
- All 8 tests in `tests/plugin-empty-turn-retry.test.ts` pass in isolation.

## Constraint-by-constraint check (#870 OP)

- **Gate placement** — the terminal gate runs *after* the event is processed (`src/plugin.ts`, loop tail: process → count text → gate → write-or-discard), so a coalesced chunk carrying content *and* the finish reason counts its own text before the turn can be called empty. Correct.
- **At most one retry** — `degenerateRetried` is set *before* `await refetch()`, so even a failed/`null` refetch consumes the one shot and passes the original terminal through. Correct.
- **Clean-terminal whitelist** — `CLEAN_TURN_REASONS = {stop, end_turn, stop_sequence}`; `max_tokens`/refusal/safety/content_filter/tool_use never re-prompt. Correct.
- **OpenAI framing** — gating on the `finish_reason` chunk (not `[DONE]`) means the first attempt's terminal is dropped *before* its `[DONE]` is ever read (buf reset + reader swap inside `retryEmptyTurn`), so the client sees exactly one finish_reason and one `[DONE]`. Correct.
- **Anthropic framing** — `retryFraming` consumes only the retry's `message_start`; block indices are offset by `blocksForwarded` via top-level-`index` rewriting on `content_block_(start|delta|stop)` data lines. Single start/stop, retry content lands after the client's own blocks. Correct. `offsetRetryIndices` degrades safely on multi-line data (unparseable line → untouched) and is a no-op outside retry / for OpenAI chunks (no top-level `index`).
- **Retry-also-empty** — the retry's own terminal hits the gate again, `degenerateRetried` short-circuits it, the empty turn passes through, and the existing `[degenerate-turn]` warning fires post-loop (`maybeWarnDegenerate`, wire `plugin-passthrough-*`). Correct.
- **Truncation (#721)** — takeover resets `sawTerminal`, so a cut retry stream still raises the in-band truncation signal. Correct.
- **Nudge injection** — `appendTrailingUserText` is a verbatim extraction of the old `injectFakeCompletionHint` body (behavior-preserving split into `src/wire-body.ts`); merges into the last user turn, appends a fresh one only when the body ends assistant-side, `null` on unusable bodies. Unit-pinned. Correct.
- **Wiring** — both `server.ts` call sites pass the agent's own forwarded `body` (same variable `resolveFakeCompletion` uses), `buildForwardHeaders(headers)`, upstream URL, dispatcher, and the client abort signal; `null` from `makeContinuationRefetch` (aborted / unbuildable body / non-ok / exception) always falls back to passthrough. Responses wire deliberately untouched — matches scope.

## Non-blocking observations

1. **Thinking-only turns also trigger the retry** (visible-text counter ignores `reasoning_content`, which is forwarded verbatim). Reading the #732/#821 precedent, this seems intentional — those hosts aborted on thinking-only turns too — but flagging it explicitly since the OP describes the tag-echo shape. Each such turn costs one extra full request, bounded to one per turn.
2. When the retry takes over, a filter tail held at turn-end is swallowed with the dropped terminal output (previously `flushTails` would have delivered it at EOF). Only affects turns whose entire visible content is tag-shaped — i.e. exactly what #463 wants gone — so this reads as more correct, not a regression.
3. On takeover the first attempt's reader is abandoned without `cancel()`/`releaseLock()` (the `finally` releases only the current reader). The upstream body is already ~drained (terminal seen), worst case one briefly-held socket; cosmetic.
4. Optional coverage for a follow-up: retry-stream truncation (#721 signal after takeover) and the coalesced content+finish_reason chunk both trace correctly in code but have no dedicated test.
5. Gemini exclusion (waiting on #867's protocol-union widening) and the adjacent #872 (tag wrapping the whole turn — same empty-turn shape, covered by this retry) are sensible scoping calls; noted.

Diff hygiene: 8 files, every line on-purpose, no lockfile churn, `version` untouched, CHANGELOG entry present. Merge stays human-side per repo rules.

中文摘要：修复了插件透传模式下模型整轮只吐出被剥离的 render tag、导致客户端拿到空回合而中断的问题（干净终止且零可见文本/零工具调用时，用 agent 自己的 body 追加一次续写 nudge 重发，严格限一次重试并保住两种 wire 的帧结构）；已逐条核对 #870 的全部约束并在 rebase 后的分支上验证 typecheck/测试/构建（唯一失败项在 master 上同样失败，属既有问题），可以合并。

joshwapohlmann:
Three things: the head moved, your five observations answered, and two of them acted on.

## The head moved since your review

You reviewed `8d0a089`. The branch is now `817a8cf` plus one follow-up commit, and the new work is the third pipe: **the Responses passthrough had no gate at all**, so an echo-only codex subagent turn still reached the host empty — 8 of one day's degenerate turns in a single session, 56 across all sessions. Your "Responses wire deliberately untouched — matches scope" was accurate for the head you saw, not for the branch now.

Its terminal is a family rather than one event, so the retry holds `response.output_text.done` / `response.content_part.done` / `response.output_item.done` until `response.completed` decides the turn: on a degenerate completion the held events *and* that completion are dropped, the agent's own body is re-issued with the same nudge, and the retry is reframed onto the ids the client already holds (its `response.created` and its added events dropped, `item_id`, `output_index` and the completed output ids rewritten to the first attempt's). A healthy turn releases the held events in place, so holding never reorders a turn. Wired at all three Responses passthrough sites. Five tests pin it, 13/13 in `tests/plugin-empty-turn-retry.test.ts`.

## Your five observations

1. **Thinking-only turns also re-trigger** — intended, and I would rather keep it. The hosts aborted on thinking-only turns for the same reason (nothing visible to them), which is what #732/#821 fixed, and the cost is bounded: one extra request, once per turn, and only after a *clean* terminal with zero visible text.
2. **Held tail swallowed with the dropped terminal** — agreed, more correct per #463. A tail still held at turn end is by construction tag-shaped, which is exactly what that contract removes; your reading is the one I intended.
3. **Abandoned reader** — fixed. Both pipes now close the first attempt's reader explicitly before swapping in the retry's, instead of relying on the `finally` (which by then only owns the new one).
4. **The two coverage gaps** — added, both, as you suggested:
   - a chunk that coalesces content with the finish reason counts its own text — asserted twice, once with prose (no retry, the prose delivered) and once with an echo as that coalesced content (exactly one retry, tag never leaks). The second is the one that actually pins the ordering your review traced: process, then count, then gate.
   - a retry stream that is cut still raises the #721 truncation signal, plus the assertion that the client is still terminated — that is the `sawTerminal` reset you checked in code.
5. **Scope calls** — agreed on both. The Gemini wire stays out until #867's protocol-union bump exists, and the adjacent #872 turn shape is covered by this retry rather than by a second mechanism.

## What I would look at on the re-review

The third pipe is the part you have not seen, so that is where a second pass earns its keep: the id rewriting, the drop-on-takeover of the held family, and the framing parity with your points about `response.created`. Everything else in the PR is unchanged apart from the reader fix and the three tests above.

Nothing else is pending on our side; merge stays human-side per the repo rules. Thanks for the constraint-by-constraint trace — the coalesced-chunk ordering is exactly the kind of thing that looks correct in code and fails in production.

joshwapohlmann:
One more commit on the branch, and it is a bug this PR's own third pipe introduced — found by proving the pipe against the *installed* bundle rather than the build.

A Responses turn whose only output is an echoed render tag now reaches the retry correctly, but the retry's `response.output_item.done` was passed to the client still carrying the retry's own `item.id`. Everything else had already been reframed onto the first attempt's ids — `response.created`, the deltas' `item_id`, the completed output ids — so the client held one item and then watched its identity change mid-turn, which is the shape this pipe exists to prevent.

`rewriteRetryIds` now rewrites the nested `item.id` too, and the regression pin emits that frame from the retry. Before the change the assertion on it fails, and the reason is worth naming: the fixture had never produced a `response.output_item.done` at all, so the assertion that the retry's ids are rewritten was passing over a frame it never saw. The hole was in the pin, not in the assertion.

How it surfaced is the part I would keep: the chat pipe's prover cannot see this path, so I wrote the Responses twin (`prove-installed-responses-turn.mjs`), which drives one plugin-mode turn through the installed CLI against a mock Responses upstream and reads the client's own bytes. On the pre-fix bundle it printed `retry ids present: YES (1 frame(s))` with the offending event; on the rebuilt bundle, `no`, with the completed output carrying the item id the client already held. Unit tests passed on both, which is the point: the assembled ids look right in the pipe's own state and wrong at the wire.

joshwapohlmann:
## Gap found in production: a turn that ends in *visible* orphan markup does not retry

The gate counts `visibleTextChars` from what the client actually receives post-strip (`src/plugin.ts:1041`, `:1049`, `:1100`), so a turn whose only text the filter swallows retries correctly. A turn whose orphan markup the filter *declines* to swallow — the unterminated render-tag case, released at the swallow cap — leaves `visibleTextChars > 0`, and the gate returns false at `src/plugin.ts:885`.

Two production occurrences on 2026-09-16, both on the OpenAI wire with DeepSeek:

| when | session | evidence |
| --- | --- | --- |
| 17:29:09Z | architect `01a0a0cb-4a90-74db-8e19-814e80cd4e20` | recorded as the fixture at `tests/tag-echo.test.ts:612-628` |
| 19:36-19:41Z | `01a0ab88-a02e-7459-b278-56d7dbfe9746` | the client received `m00693</cp>` plus a `｜｜DSML｜｜`-wrapped parameter block, found no tool call in it, and stalled until interrupted |

The second occurrence is the first one again: the model opens a render tag and writes the whole turn inside it with the attributes still open, so no close is ever written and no tag regex can match the span.

Proposed condition, in the shape of the existing one: treat the turn as empty when everything the client received was markup the filter declined to swallow. `src/plugin.ts` already carries the predicates to decide that (`mayStartRenderTag`, `mayStartMarkerLine`, the pending-span state at `:1091`), so the gate can ask whether the surviving text holds any non-markup content instead of whether it is non-empty.

## Second condition: surface the failure instead of ending empty

Both occurrences shared a worse property than the missing retry: the turn **stalled rather than failing**, so nothing reached the client until the user interrupted by hand. An empty turn is indistinguishable from a model that simply produced nothing, and the session looks idle while it is dead.

After the retry has been spent, a turn that ends in orphan markup with no tool call should raise an in-band error. The channel exists: `src/stream-error.ts` already emits a signal when a stream ends without a terminal event, and the client renders it.

Not implemented in this branch yet. The wire bytes of the second occurrence are gone — the proxy log records events rather than bodies, and the payload store keeps only the proxy's own transformed view — but the client-visible residue survives in that session's transcript, so a fixture has to be reconstructed from the client's view rather than recorded off the stream. The `｜｜DSML｜｜` marker is the model's own envelope: it appears nowhere in `src/` or in the shipped bundle.

joshwapohlmann:
## Third occurrence, 2026-09-16T20:46:05Z: per call, not per turn

Same session as the second occurrence (`01a0ab88-a02e-7459-b278-56d7dbfe9746`), this time on a `hub` call: the parameter block opened normally and then closed with a stray `</parameter>` inside it, the intent parameter carried `｜｜DSML｜｜`, and the call's real parameters were absent, so nothing executed. The model re-issued the same call correctly on the next attempt and it ran.

Two properties this adds:

- **per call, not per turn** — the call immediately before it, in the same minute and of the same kind, arrived intact;
- **not tool-specific** — the first occurrence was a `read`, this one a `hub`.

In all three cases the client showed a stalled session rather than an error, which is the second condition above.

Where the bytes are not, for anyone chasing a fourth: `bili.log` (and its rotated `.old`) records events rather than bodies; the payload store keeps the proxy's post-transform view of a conversation, and a turn the client discards never enters a later request; the harness's `http-400-requests/` dumps full bodies but only for 400s, and this is a parse failure in the client rather than a 400.

joshwapohlmann:
Implemented in a008628.

The gate now counts the text a dropped tag leaves behind as markup rather than as visible output, at both of the chat wire's rewrite sites, so a turn whose only visible content is a surviving tag interior is re-issued. Measured behaviour: that interior has already reached the client by the time the terminal arrives, so the retry cannot erase it — what the fix guarantees is that the turn carries content the host can act on, which is what the stall was missing.

Fixture `unclosedEchoTurn()` in `tests/plugin-empty-turn-retry.test.ts`, reconstructed from the client-visible residue of the second occurrence because the wire bytes are kept nowhere. It fails before the change (no re-issue) and passes after. `tests/plugin-empty-turn-retry.test.ts` 17/17, `tests/tag-echo.test.ts` 44/44 including the recorded architect-session fixture, `tests/plugin-passthrough-tag-strip-chat.test.ts` 20/20 — the #463/#673 pins are untouched.

Still open, unchanged: the in-band error when the retry is also degenerate, and the same gate on the Responses pipe, whose `heldVisibleChars` has the twin of this hole.

joshwapohlmann:
## Evidence for the second condition, from the session that hit it five times

Reported by `01a0ab88-a02e-7459-b278-56d7dbfe9746`, which saw the shape five times on 2026-09-16 between 20:44Z and 20:50Z, all on `hub` calls:

- the client surfaced **nothing**: no parse error, no warning, no error line. The residue appeared as ordinary assistant text, the real parameters were absent, and the turn ended normally;
- the first occurrence stalled until the user interrupted, the later ones did not — the same corruption produces two different client-visible outcomes, and neither is reported as a failure;
- the trigger is the marker being **emitted**, not the tool being called: writing it verbatim into a message reproduced the fault immediately, losing that message's recipient parameter and truncating the body at that position.

That last point locates the cause away from this repository. The marker is DeepSeek's own envelope vocabulary — fullwidth vertical bars, `U+FF5C`, its special-token delimiter — and the proxy neither writes nor handles it: zero occurrences in `src/` and in the shipped bundle. A serving stack mishandling its own envelope explains the corruption. What this repo owes the user is that the resulting turn is not silently empty, which is the second condition above.

joshwapohlmann:
## `fc8a350` ("reach the degenerate-turn retry on the Google wire too") is blocked on the wire branch, not pending

That commit sits in `integration` but never reached this branch, which is why refreshing `integration` conflicts: it edits `src/plugin.ts` and `tests/plugin-empty-turn-retry.test.ts`, the same files as the fix above.

Replaying it here does not compile, and the errors name the reason: `src/plugin.ts(1147)`, `src/wire-body.ts(37)` — `This comparison appears to be unintentional because the types '"anthropic" | "responses"' and '"google"' have no overlap`. Its google branch needs `"google"` in the `WireProtocol` union, which the wire branch (#866) defines and this branch does not.

So it belongs here only after the wire lands, in the same sequencing as the kernel dependency in #867. Before replaying it, check whether the wire branch already wires the retry for the google path — that commit was written against `integration`, where both are present, so the two may be partly the same change.

joshwapohlmann:
Checked the sequencing question above, and it is not redundant: the wire branch has no `src/wire-body.ts` at all, and its `src/plugin.ts` has no `retryEmptyTurn`. It carries the google wire's suppression path (`googleFrameHasNonText`) but never wired the retry to it. So `fc8a350` is a genuine addition that needs the wire under it, in that order.

joshwapohlmann:
## The Responses pipe has the same hole; the measurements for whoever takes it

Tested rather than assumed: an unclosed render tag on the Responses wire does **not** retry. The probe is the chat one's shape on that wire (`responsesUnclosedTurn()`), and the re-issue never fires.

Four attempts to mirror the chat fix failed, and the measurements say why that pipe's accounting is not symmetric:

- `heldVisibleChars` counts the done-family, which **repeats the turn's text**, so it is not a second measure of what the client saw; the deltas are.
- For an unclosed tag the deltas' filtered text is empty — the span is still pending — so `visibleTextChars` is 0 and the text arrives through `flushTail`, which neither counter includes.
- The held path strips with `stripResponsesText`, not the shared filter, so a drop there sets no flag.
- Dropping the `heldVisibleChars > 0` clause breaks a pinned test, so that clause is load-bearing for the echo case.

Net: a correct fix must measure the attempt's text **once**, across the delta, the flush tail and the held family. Left out rather than half-done. The chat wire, which every observed failure was on, has both halves working.

ranxianglei:
ci失败

ranxianglei:
继续

ranxianglei:

**CI failure — root-caused, fixed, re-pushed. All jobs green on `86abe3a`.**

Reproduced all three red test jobs locally with an A/B against base `d650db7`. Two were genuine regressions introduced by this PR's new degenerate-turn retry; one was a stale-base artifact:

**1. `native-compact-terminal` — "trigger stays final" (PR-introduced).** The `resetAfterSuccess` passthrough serves the native-compaction *trigger* request through `pipePluginResponsesWithStrip` with the new continuation refetch. The trigger's empty-output completed response hit the retry gate, and the re-issued body appended the nudge **after** `compaction_trigger` — breaking the wire shape (#283: the trigger must stay the last input item). In production that risks double-compaction / malformed follow-up requests. Fix: `makeContinuationRefetch` now refuses to re-issue a body whose final input item is a compaction trigger (lazy-parsed at retry time, so healthy turns pay nothing). This mirrors the guard the proxy-mode nudge path already carries (`isCompactionTrigger`, `src/server.ts`).

**2. `responses-tool-images` — plugin-mode TypeError (PR-introduced).** The new gate also retried turns where the model produced *nothing at all* — no text, no reasoning, no tool call — i.e. the upstream's own empty completion. That double-bills every empty completion and appends a stray user turn to the forwarded request (the `endsWith` crash came from filtering that appended item). The compress loop has kept a stricter boundary since #732/#821: only invisible-but-real output retries, never a genuinely bare turn. Fix: both plugin pipes now require evidence of invisible-but-real output before retrying — chat pipe: thought, a stripped tag/marker echo, or released markup; Responses pipe: reasoning, or a stripped echo seen in deltas / done-family events.

**3. `#857 D: upward self-heal trusts usage baselines only` — pre-existing, not from this PR.** Fails byte-identically on the clean base; already fixed on master by `323d8c1f` (post-#852 DeepSeek 1M window table). Resolved by the rebase below.

**What changed:** rebased onto current master `55da3f0`, plus one commit `86abe3a` — 40 lines in `src/degenerate-retry.ts` + `src/plugin.ts`, nothing else. Pre-flight: typecheck ✓, full suite **1688/1688** ✓, build ✓. CI on `86abe3a`: all 8 jobs green, including the four test-matrix jobs that were red. The PR's own 18 retry-contract tests pass unmodified.

Ready for merge whenever.

修了什么问题：PR 新增的空轮重试有两个回归——(1) 会对 native-compaction 触发请求重发、把 nudge 追加到 `compaction_trigger` 之后破坏 wire 形状；(2) 对模型完全无输出的空 completion 也重试（双倍计费）。已在 `86abe3a` 修复（trigger 结尾的 body 一律不重发；重试门前要求存在"不可见但真实"的输出证据），并 rebase 到最新 master（第三个失败是旧基线问题，master 已修）。本地全量 1688 测试全绿、CI 8 个 job 全绿，可以合并。

PR #875, opened 2026-09-17
fix(agent): /acp must not wake the model under OpenCode V2

ranxianglei:
Stacked on #850 (base = its head).

**Fixes #876.**

Every `/acp` invocation passed `{sessionID, text}` to `ctx.session.synthetic()` and left OpenCode V2's defaults in place: delivery `steer` **and** `execution.wake(sessionID)` (core `Session.synthetic`, packages/core/src/session/session.ts: resume!==false and !revert -> wake). Net effect: each /acp started a full model turn with no user input - spurious empty turns/loops, extra upstream forwards right after typing /acp.

Pass `resume: false`. Verified live against an opencode v2.0.4 TUI through bili: same render path, zero model requests after /acp (proxy forward count frozen), whereas before the fix it grew on every invocation.

Preflight: typecheck clean; opencode-v2 suite green (assertions extended: the synthetic call must carry resume:false); full npm test green.

Suggested merge order: this first, then the stacked panel-visibility PR.


ranxianglei:

**Review of #875 — verified, LGTM to merge (after #850).**

**Diff** (1 commit `d6d60f8`, 2 files, +8/−4):
- `src/agent/opencode-v2.ts:90` — widens the `session.synthetic` type to `{ sessionID; text; description?; resume? }`.
- `src/agent/opencode-v2.ts:261` — `/acp` handler now calls `ctx.session?.synthetic?.({ sessionID: sid, text, resume: false })` (was `{ sessionID: sid, text }`).
- `tests/opencode-v2.test.ts` — fake `synthetic` widened to match + two new assertions `assert.equal(fake.syntheticCalls[0].resume, false)`.

Clean and minimal, no unrelated churn. No small issues to fix directly.

**Why it's correct.** `/acp` is documented to render a *non-model-turn* message (`src/agent/opencode-v2.ts:211-213`). OpenCode V2's `Session.synthetic` defaults to `delivery:"steer"` + `execution.wake()` unless `resume===false`, so every `/acp` kicked off a spurious empty model turn (the extra forwards / token spend in #879). `resume:false` suppresses the wake while still delivering/rendering the panel. This is the **only** `synthetic(` call site in the repo, so the fix is complete.

**Independent preflight on head `d6d60f8`** (checked out locally, `npm ci`):
- `npm run typecheck` → clean
- `tests/opencode-v2.test.ts` → **16/16 pass**
- full `npm test` → **1658 pass / 0 fail / 2 skipped** (exit 0)

Matches your reported preflight exactly.

**Stacking & CI.** Parent of `d6d60f8` = `f857ff1` (#850 head) → cleanly stacked on #850. GitHub: one check-run green ("real codex vs deterministic fake upstream (zero tokens)"); overall commit status still `pending`, which is expected for a stacked PR whose base isn't merged yet — my local run above covers the same ground.

**Merge order:** #850 first → this → #880, as noted in the body. Nothing for me to change.

中文摘要:修复了 OpenCode V2 下每次 `/acp` 都会因默认 `wake()` 触发一次无输入空模型轮次的问题(改为传 `resume:false`);diff 干净最小、本地预检全绿(typecheck + opencode-v2 16/16 + 全量测试通过),叠在 #850 之上,按建议顺序(#850→本 PR→#880)可合并。

Issue #881, opened 2026-09-17
[bug] opencode v2: first /acp before any model request says 'proxy returned no status panel' instead of the idle notice


ranxianglei:
Repro on the merged #850+#875/#878 line (v2 native plugin, real v2.0.4 TUI):

1. Fresh proxy + TUI, send NO model request yet.
2. Run `/acp`.
3. Got: `bili: proxy returned no status panel`
   Expected: `billion-context@<version> — proxy connected, no ACP session yet. Send a model request, then run /acp again.`

Second /acp after a model request works fine (panel renders).

## Root cause
With zero sessions, `handlePluginStatus` (src/plugin.ts) answers the status query with HTTP **404** `{ok:false}`. The V2 renderer uses `shared.fetchStatus()` (src/agent/shared.ts), whose soft-fail contract turns every non-2xx into `undefined` - indistinguishable from 'proxy unreachable'. In src/agent/opencode-v2.ts both branches require a truthy `status` (`status && …panel`, `status && status.ok === false`), so the idle case fell through to the generic 'no status panel' text. (V1's inline fetch reads the JSON body regardless of status code, which is why only V2 misbehaves.)

## Fix (client-side)
Treat `status === undefined` as its own case and probe the manifest first: if the manifest answers, render the friendly `billion-context@<v> — proxy connected, no ACP session yet…` line; if nothing answers, report the proxy as unreachable. Endpoint wire semantics untouched.

Fixed in branch 2026-09-17_opencode-v2-acp-idle-first (stacked on #850 head).

ranxianglei:
> 🤖 Powered by [ework](https://github.com/ranxianglei/ework) · qwen3.8-27b

Reviewed and independently verified the fix on **PR #882** (branch `2026-09-17_opencode-v2-acp-idle-first`, head `6d79228`). It's correctly stacked on #850's head (`2026-09-15_opencode-native-plugin`) and the PR body carries `Fixes #881` — so no duplicate PR was opened.

**Root cause — confirmed.** With zero resident sessions, `handlePluginStatus` answers the status query with HTTP `404 {ok:false}` (`src/plugin.ts:466-468`). `shared.fetchStatus()` soft-fails every non-2xx into `undefined` (`src/agent/shared.ts`), so pre-fix both guards in the v2 renderer required a truthy `status` and the idle case fell through into the final `else` → `"bili: proxy returned no status panel"`. Your repro steps map onto exactly this path.

[...]

PR #882, opened 2026-09-17
fix(agent): first /acp under opencode v2 shows the idle notice, not 'no status panel'

ranxianglei:
Stacked on #850 head (which already carries the #875 no-wake + #878 visible-panel fixes).

**Fixes #881.**

Before the conversation's first model request the proxy knows no ACP session and answers the status query with HTTP 404 {ok:false}. shared.fetchStatus() soft-fails every non-2xx into undefined - indistinguishable from 'proxy unreachable' - so the v2 renderer fell through both the panel branch and the ok===false branch into 'bili: proxy returned no status panel'. V1's inline fetch reads the body regardless of status code, which is why only V2 misbehaved; second-and-later /acp calls were always fine.

Fix (client-side, src/agent/opencode-v2.ts): handle the undefined result explicitly - probe /__bili/plugin/manifest first, and when it answers render the friendly 'billion-context@<v> — proxy connected, no ACP session yet. Send a model request, then run /acp again.' line; when nothing answers, report the proxy as unreachable. Endpoint wire semantics untouched (the shared 404 contract is relied on by dsh/pi consumers and e2e fakes).

Tests: fake proxy now serves a 404 {ok:false} status for an idle conversation id plus a live manifest; new test asserts the versioned idle notice on description, inert body, resume:false. Suite 18/18, typecheck clean, full npm test green.

Verified shape matches the expected user-visible string exactly.


ranxianglei:

[bot] 🏷 Reviewed PR #882 (single commit `6d79228`, stacked on #850 head `2026-09-15_opencode-native-plugin`).

**Diff cleanliness** — clean. Two files, +41 lines, every line on-topic: `src/agent/opencode-v2.ts` (+8, one new branch) and `tests/opencode-v2.test.ts` (+33). No unrelated files, no generated/lock-file churn, no whitespace reflow.

**Correctness** — verified against the actual wire contract, not just the description:
- Server side (`src/plugin.ts` `handlePluginStatus`) returns **HTTP 404 `{ok:false, error:"unknown plugin conversation"}`** for an unknown/idle conversation. Confirmed the premise.
- `src/agent/shared.ts` `fetchStatus` soft-fails *any* non-2xx to `undefined`, so that 404 becomes `undefined`. Before this change it fell through both the `panel` and `ok===false` branches into the final `else` → `"bili: proxy returned no status panel"`. The new `else if (!status)` branch catches exactly that case and probes `/__bili/plugin/manifest` via `fetchProxyVersion`: manifest live → the versioned idle line; nothing answers → "cannot reach". Branch ordering (`panel` → `ok===false` → `!status` → fallback) has no fall-through gaps, and the second-and-later paths are untouched. Endpoint wire semantics unchanged, as claimed. ✓

**Edge case I traced (no action needed):** a genuinely *unreachable* proxy makes `fetchStatus` **throw**, which the existing outer `catch` turns into `"bili: /acp failed (...)"` — it does *not* return `undefined`. So the new branch's "cannot reach the bili proxy" sub-message is defensive-only; it fires only if the proxy answers the status endpoint but not the manifest endpoint. That's fine and consistent with the sibling `ok===false` branch already doing the same manifest probe.

**Tests** — the new test mirrors the real contract precisely (fake proxy 404s `ses_acp_idle` + serves a live manifest) and asserts the three invariants: versioned idle text on `description`, inert body (`"not an instruction"`), `resume:false`.

**Verification:** typecheck clean · `tests/opencode-v2.test.ts` 18/18 · full `npm test` green (1660 pass / 0 fail, 2 skipped = the gated e2e). CI's `real codex vs deterministic fake upstream` check passed; no failing checks observed.

**⚠️ Merge-order dependency:** this PR bases off **#850's head**, and **#850 is still open**. So #882 cannot land on `master` until #850 (and its base branch) merges first. Please keep the stack in mind when merging.

**Verdict:** correct, minimal, well-tested, CI green — ready to merge once #850 lands. Per project rules I don't merge PRs myself; please merge it yourself (after #850): https://github.com/ranxianglei/billion-context/pull/882

---
一句话总结:修复了 opencode v2 在首次模型请求前执行 `/acp` 时误报 "proxy returned no status panel" 的问题——代理对未知会话返回的 `404 {ok:false}` 被 `shared.fetchStatus` 软失败成 `undefined`,新增 `!status` 分支先探测 manifest 再渲染带版本的空闲提示(无响应才报不可达);改动干净、单测/全量测试/CI 全绿,待 #850 合入后即可合并。

ranxianglei:
这个我已经合并了，但是明显有问题。

```
`billion-context@${version} — proxy connected, no ACP session yet. Send a model request
```
这个东西明细是被复制粘贴很多次，在不同的插件实现上复制粘贴。而不是一个整体
你提交一个issue专门说明这个问题

PR #885, opened 2026-09-17
fix(agent): opencode v2 preset-env attach — close the pseudo-attach 404 hole (three-mode delivery verified)

ranxianglei:
Stacks on #850 (base = its source branch). Delivers and verifies the three opencode v2 integration modes end-to-end, and fixes the bug behind the "tools 404 / ACP never activates" reports.

## The bug
Presetting `BILLION_CONTEXT_PROXY` and running bare opencode disarmed the native bootstrap (no spawn, no attach) while tools still found the proxy through the env: model traffic silently went **direct**, the proxy never saw the session, and every tool forward 404'd (`unknown plugin conversation`). The model then told users to "refresh/restart" — the exact symptom reported during v2 testing.

## The fix
- `planNativeOpencode`: a preset `BILLION_CONTEXT_PROXY` is now an **attach target** (probe + rewrite + stamp — same semantics as `BILLION_CONTEXT_ATTACH`; explicit `ATTACH` wins when both set). Kill switches (`BILLION_CONTEXT_PLUGIN=0`, `BILI_NATIVE_OPENCODE=0`) and `/bili/` launches (`BILI_PROVIDER_REWRITES`) still stand down; a garbage preset falls back to spawn. Double-rewrite is impossible: `isModelApiUrl` already skips `/bili/`-shaped URLs.
- `opencode-v2` tool forward: the never-registered 404 now returns guided text ("send one normal message first; ACP activates automatically") instead of the raw error the model misread.

## Verification (real upstream, in-container, all four postures)
| posture | session bound | conversation | acp_status | /acp panel |
|---|---|---|---|---|
| 1 proxy (baseURL `/bili/`) | ✅ | ✅ | ✅ injected | n/a |
| 2 launcher (`bili opencode`) | ✅ | ✅ | ✅ local→proxy | ✅ full |
| 3 native spawn (bare opencode + plugin) | ✅ | ✅ | ✅ local→proxy | ✅ full |
| 4 preset-env attach (the fixed hole) | ✅ | ✅ | ✅ local→proxy | — |

Bun daemon `http.request` Request-replacement (the #809 core risk) confirmed live in postures 3 and 4: traffic registered on the plugin-managed proxy with `[plugin] [acp-usage]` header evidence and 99% prefix-cache hits.

Pre-flight: `npm run typecheck` ✅, full `node --import tsx --test` suite ✅ (b50 run), focused re-run `opencode-native`/`opencode-v2` 19/19 ✅, `npm run build` ✅.

Fixes the pseudo-attach half of #809's verification list; the residual daemon-lifetime caveat (a stale `serve --service` daemon predating a bili launch keeps its old env) remains a usage note, not code.


ranxianglei:
Fixes #886 (issue filed for the pseudo-attach 404 this PR closes).

PR #891, opened 2026-09-17
fix: recover from mid-stream upstream truncations (#413 relaxation + continuation retry)

### LINKED ISSUE #862: [Question] zocde调用子智能体的时候，子智能体有概率出错，导致子智能体停止工作
具体的报错：
[acp-proxy: upstream stream truncated (no completion event; round 1, 0 text chars + 90789 reasoning chars received)]
只修改了下面的设置
  "modelContextLimit": 300000

ranxianglei:
## What

Recovery for the two mid-stream truncation shapes seen with high-reasoning upstreams behind relays (#862, field-validated over 6.5h / 30+ rounds):

1. **#413 gate relaxation** — a stream that dies after forwarding only *invisible* output (reasoning/meta prefix) now gets the blind re-fetch instead of erroring. Previously this shape (`0 text + N reasoning`) was a hard error — 7 occurrences across 5 sessions in the reporter's logs.
2. **Continuation retry (new)** — when the stream dies *after visible text* has started (the long final-report shape), one re-fetch with an ephemeral nudge quoting the forwarded tail (≤800 chars) and asking the model to resume seamlessly without repeating. The client keeps its partial text; the retry's output appends to it. `roundBody` is updated so any later blind re-fetch also carries the continuation context.

Two independent one-shot budgets (`truncationRetried` / `continuationRetried`): max 2 extra fetches per request, each logged, so both cut orderings self-heal. Rounds with tool-call fragments keep the plain truncation error. The nudge never enters `coreMessages`/session state (same ephemerality as #732's).

## Why wire-scoped (review refinement of the contributed patch)

The patch contributed in #862 relaxed the #413 gate to `!forwardedVisible` unconditionally. That is safe on the **openai wire** (stateless chunks — the reporter's channel), but on **stateful wires** a re-fetched response would re-emit per-response identity framing that already reached the client:

- anthropic: `message_start` plus open content blocks would be duplicated;
- responses: a second `response.created` — breaking #440's single-created invariant (its T2 test fails against the unconditional relaxation).

So the relaxation lands as `(!forwardedAny || (protocol === "openai" && !forwardedVisible))`: the original zero-side-effect case stays universal, the new invisible-prefix case is openai-only, and the continuation retry is openai-only for the same reason (visible text implies the identity framing already went out). Existing `#413-T3/T4` (anthropic) and `#440-T2` (responses) tests now pass unchanged and guard this scoping. Behavior on the reporter's channel is identical to their deployed patch.

## Tests

- Flip `tests/loop-truncation-reasoning.test.ts` (asserted the old strict gate).
- New `tests/loop-truncation-continuation.test.ts`: both cut orderings (visible→invisible→clean, invisible→visible→clean), budget exhaustion → surfaced error, tool-call rounds excluded, failed-retry fallback, `textProtocol` exclusion.
- Pre-flight on master 0.1.117: typecheck clean, 1675/1677 pass (2 skipped), build ok.

## Follow-ups (reported in-thread, intentionally not in this PR)

- Stateful-wire truncation recovery needs adapter-level handling of the dead attempt's framing (close open blocks / synthesize a terminal before appending the retry's identity frame) — separate work.
- weak-overflow low-utilization blind spot (`input/window < 0.9` never counts, so conservative windows never learn) — real samples collected by the reporter; human to file the issue.
- Visibility-marker suppression → config item + `MARKER_INTEGRITY_NOTE` wording — separate PRs, green-lit in-thread.

Based on the field diagnosis and initial patch contributed in #862.

Closes #862


ranxianglei:

[bot] 🏷 **PR #891 verified — ready for human merge.**

Per the [PR] protocol I did not redo the work: pulled the branch, reviewed the full diff against current master, and checked CI.

**Branch state:** `2026-09-17_truncation-continuation-retry`, single commit `9a69b0e` directly on current master `55da3f0` (v0.1.117) — no base drift, mergeable/clean, 3 files +293/−24.

**CI:** green 8/8 on head `9a69b0e` — test matrix (ubuntu/windows × 22/24), version-guard, build-artifact, and the real-codex-vs-deterministic-fake-upstream e2e job (the responses-wire regression gate).

**Diff review (`src/loop/core.ts`):**
- The #413 gate relaxation lands wire-scoped exactly as described: `(!forwardedAny || (protocol === "openai" && !forwardedVisible))`. For non-openai wires the predicate reduces to the original `!forwardedAny` — zero behavior change there. Confirmed the guards it must not break exist and pass unchanged: `#413 T3/T4` (anthropic open-block shapes, `tests/loop-truncation.test.ts:172,186`) and `#440 T2` (responses single-created invariant, `tests/loop-responses-orphan.test.ts:109`).
- Continuation retry: openai-only, one-shot `continuationRetried` independent of `truncationRetried`. Traced all cut orderings through the per-attempt flags (visible→invisible→clean, invisible→visible→clean, both-budgets-exhausted) — consistent with tests C2/C3/C4. The nudge is genuinely ephemeral (spread into `adapter.buildRequest([...coreMessages, nudge], …)`; `coreMessages`/session state untouched), tail capped at 800 chars with a `…` marker, `roundBody` is updated so any later blind re-fetch carries the continuation context, the clear-timer swap mirrors the existing #413 retry path, tool-call rounds stay excluded via `calls.length === 0`, and a failing re-fetch falls back to the plain truncation error (C7).
- The flipped test in `tests/loop-truncation-reasoning.test.ts` is correct: a reasoning-only *incomplete* stream is the truncated shape (blind re-fetch), distinct from #732's completed-with-thinking degenerate shape (nudge retry) — the two gates stay mutually exclusive on `sawDone`/`truncatedDone`.

**One nit found and fixed directly on the branch:** three comment lines in the new block carried a stray extra leading space (`src/loop/core.ts:436-438`). Pushed as `10f2a83` (comment-only; local typecheck clean, 18/18 across the four truncation-related suites). CI will re-run on the new head.

No blocking issues. The follow-ups listed in the body (stateful-wire framing recovery, weak-overflow blind spot, visibility-marker suppression) remain intentionally out of scope.

中文摘要:已核对分支并逐行 review——#413 门放宽与续写重试均正确限定在 openai wire、双独立一次性预算、nudge 不污染会话状态,CI 8/8 全绿且既有 anthropic/responses 守卫测试原样通过;review 中发现注释块三行多一个前导空格,已在 PR 分支直接补 commit 修复(10f2a83),无阻塞问题,可以合并。

ranxianglei:

[bot] 🏷 Follow-up: CI on the updated head `10f2a83` is green 8/8 (test matrix ubuntu/windows × 22/24, version-guard, build-artifact, codex e2e). Nothing outstanding — ready for human merge whenever.

PR #895, opened 2026-09-17
test(launcher): isolate the host client, proxy and CA env vars

joshwapohlmann:
A harness that launches bili exports its own client binary, its proxy URL and its CA into every child process. Inherited by `tests/launcher.test.ts` they break it: two tests hang and two fail.

`src/launcher.ts:2190` reads `BILI_CLIENT_BIN` and returns it as the command, ahead of the `client: "pi"` resolution the tests configure. Their injected `spawnImpl` fires the child's `exit` only when the command is their fake client, so with the variable set the launcher awaits an exit that never arrives. Nothing keeps the event loop alive: the test is parked on a promise nothing will settle, which is why the file hangs instead of failing. `trae` then reads back the inherited `NODE_EXTRA_CA_CERTS` and `kimi` the inherited `BILLION_CONTEXT_PROXY`, both of which they assert absent.

Reproduce on master with one variable at a time, without a host harness:

```
$ BILI_CLIENT_BIN=/usr/bin/true node --import tsx --test --test-timeout=15000 tests/launcher.test.ts
ℹ tests 167   ℹ pass 165   ℹ cancelled 2
# the two pi tests never return; without --test-timeout the file never exits

$ NODE_EXTRA_CA_CERTS=/tmp/dummy-ca.pem BILLION_CONTEXT_PROXY=http://127.0.0.1:1 \
    node --import tsx --test --test-name-pattern="cert-MITM envs" tests/launcher.test.ts
✖ runLaunch trae: cert-MITM envs (SSL_CERT_FILE combined bundle), no budget/transport, default MITM whitelist (#655)
    AssertionError: trae uses SSL_CERT_FILE, not NODE_EXTRA_CA_CERTS
✖ runLaunch kimi: cert-MITM envs (combined CA on SSL_CERT_FILE + NODE_EXTRA_CA_CERTS), discovered host whitelist, model windows (#757)
    AssertionError: kimi has no agent-side plugin consumer
```

With those variables unset the same file is 167/167 in about a second. The checkout is the only input that matters, since the suite imports `src/launcher.ts` from it; an installed package, the node version and the kernel version change nothing.

The file already isolates `XDG_STATE_HOME` for the #707 starting marker; this clears the launch variables in the same block and restores them in its `after`. Test-only, 25 insertions, `npm run typecheck` clean.

_AI agent generated_


ranxianglei:

[bot] 🏷 On it — reviewing PR #895 (test-only env isolation for `tests/launcher.test.ts`). Pulling the branch now, will verify the repro claims (`BILI_CLIENT_BIN` hang, inherited CA/proxy vars) against master and the PR head, then run typecheck + the suite under both clean and hostile environments.

ranxianglei:

[bot] 🏷 **Review complete — verified, ready to merge.**

**Repro confirmed on master** (`55da3f0`, node v22.23.2):
- Clean env: `tests/launcher.test.ts` → 167/167 in ~1.5s, as described.
- `BILI_CLIENT_BIN=/usr/bin/true`: file breaks — 23 pass / 144 cancelled with `--test-timeout=15000` (the first hang cascades cancellations; without the flag the file never exits). Root cause verified at `src/launcher.ts:2190` — `resolveClientCommand` reads `env.BILI_CLIENT_BIN` before any `client:` resolution, so the injected `spawnImpl` never sees its fake client path, never fires `exit`, and `runLaunch` awaits forever with nothing keeping the loop alive.
- `NODE_EXTRA_CA_CERTS` + `BILLION_CONTEXT_PROXY` inherited: cert-MITM pattern → 1 pass / 2 fail, exactly the two assertions cited (`tests/launcher.test.ts:3716` trae `NODE_EXTRA_CA_CERTS` undefined, `tests/launcher.test.ts:3921` kimi `BILLION_CONTEXT_PROXY` undefined) — the launched client env is built from `{ ...process.env, … }` spreads, so ambient values leak into what the assertions read back.

**PR head** (`6ace3ab`, single commit directly on current master, `mergeable: true`):
- Clean env: 167/167
- Only `BILI_CLIENT_BIN=/usr/bin/true`: 167/167, zero cancelled
- CA+proxy hostile: cert-MITM pattern 3/3
- All six vars set at once: 167/167
- `npm run typecheck`: clean
- Full suite `npm test`: 1668 pass / 0 fail / 2 skipped (the default-gated e2e skips)
- CI: all green (ubuntu/windows × 22/24 tests, version-guard, build-artifact, codex e2e checks)

**Diff review:** 25 insertions in `tests/launcher.test.ts` only — no `src/` changes, no version/lockfile churn. It follows the existing `XDG_STATE_HOME` isolation pattern in the same block (module-scope save → clear → restore in `after()`), with correct undefined-vs-value restore semantics. Fixing this at the test layer is the right call: the launcher's env inheritance and `BILI_CLIENT_BIN` precedence are intentional production behavior, so the defect is ambient-env dependence in the tests, not the launcher. The blanket clear also covers the many tests that don't do their own per-var save/restore (e.g. the trae test saves `BILI_CLIENT_BIN` locally, but the pi tests rely purely on PATH resolution).

No findings to report separately. Please merge: https://github.com/ranxianglei/billion-context/pull/895

中文摘要：已在 master 复现三个失效路径（BILI_CLIENT_BIN 导致挂起、继承的 CA/proxy 变量导致 trae/kimi 断言失败），PR 头部在干净与全部六种污染环境下均 167/167、typecheck 干净、全量测试与 CI 全绿，纯测试改动且遵循既有 XDG_STATE_HOME 隔离模式，可以合并。

Issue #908, opened 2026-09-17
feat: opencode v2 local compression kernel — mode parity with `bili pi` (compress executes inside opencode, no proxy)


ranxianglei:
## Background

PR #824 delivered opencode v2 support; #850/#885 added the proxy / launcher / native-spawn modes plus the /acp command. But `bili pi` users get a fundamentally better shape: the agent-side plugin owns compression locally — no proxy hop, no `BILLION_CONTEXT_PROXY`, wire traffic goes straight to the real upstream. opencode v2 had no equivalent (its "native" mode still self-spawns a proxy).

User request (paraphrased from session): "bili pi is what we want — bili opencode is the remote mode. pi implements modes 1 (proxy) + 2 (launcher); make opencode match, with pure-local mode 3." Reference for the pure-local shape: https://github.com/ranxianglei/opencode-acp

## What this adds

`src/agent/opencode-local.ts` — an opencode v2 plugin that runs the acp-kernel compression pipeline IN-PROCESS:
- session registry keyed by conversation id, `session.hook` rebuilds the wire view through the kernel (folded messages, `acp` tags on text parts)
- `chat.params.tools` injects the kernel's compress/decompress/acp_status tool set; `chat.parameters` enforces the compress system prompt
- tool execution goes through the kernel (`executeTool`), so `acp_status` renders the same panel as `/acp`
- No proxy, no MITM, no env vars: the user's model config is untouched; context compression happens between opencode and the model, invisible on the wire except tags

## Verification done

- unit tests (`tests/opencode-local.test.ts`, 11 tests) + full suite 1673 pass
- Docker-isolated E2E with real opencode v2 + local sglang upstream: 4 postures (proxy / launcher / native-spawn / pure local) all green
- compression stress test in-container: model read ~200K tokens of real source, self-triggered compress 7 times (105K folded), refs strictly monotonic, no id reuse, two successful recall-from-summary answers, no tag leakage

Fixes ride on the same branch and are recorded in #906 (abort-guard, system-array push).

ranxianglei:
PR: https://github.com/ranxianglei/billion-context/pull/909 (stacked on #850, same as #885)

ranxianglei:
> 🤖 Powered by [ework](https://github.com/ranxianglei/ework) · qwen3.8-27b

[...]

Issue #906, opened 2026-09-17
textProtocol extracted-trigger tool results fall into the marker visibility branch (role:system) instead of tool-call/result pairing


ranxianglei:
Found while stress-testing the opencode v2 local kernel (the model read src/ and reported candidates; line numbers verified by hand).

## Problem

`src/loop/core.ts` round-2 rebuild:

```ts
const allCalls = [...calls, ...textProtocol extracted triggers];   // :451
const functionCallIds = new Set(calls.map(c => c.callId));        // :452 — structured calls ONLY
```

The `functionCallIds` gate (:546-553, deciding whether a tool result is consumed as a real `tool-call`/`tool-result` pair vs. emitted as a `role:"system"` visibility marker) is keyed on the **structured** set only, while `allCalls` already merges text-protocol extracted triggers. So a trigger extracted from plain text is adjudicated as a marker, contradicting the block's own self-description at :606-613 ("feeds the result back as a normal tool output"). Depending on the adapter's handling of system messages mid-toolflow, such a result can be silently dropped from the rebuilt wire message.

## Impact

In practice the anthropic adapter re-voices `role:"system"` results as `developer`/`user` (established during the #885 work), so this is **not** observed data loss — but it is an asymmetry between the two call sources that makes the text-protocol path depend on adapter-side re-voicing rather than the intended pairing. OpenAI-format adapters that drop bare system messages mid-conversation would lose the result.

## Suggested fix

Key the visibility gate on the merged source: `new Set(allCalls.map(c => c.callId))` — or explicitly decide extracted triggers are marker-carried and document that in the :606-613 comment.

## Also fixed on branch 2026-09-17_opencode-v2-local-kernel (commit 624b436) — sibling findings from the same stress test, recorded here per repo policy

1. **Client abort counted as weak overflow** (`src/loop/core.ts:624`): a user ESC breaks the parse loop with `!sawDone && streamError === undefined` — the exact shape the #498 detector counts. 3 aborts in 15min learn a shrunken window (`weak-overflow.ts` MIN_EVENTS=3, #570 false-positive family). Fixed by adding `&& !signal?.aborted`, mirroring the existing retry guards. Regression test: `tests/loop-truncation.test.ts` #887 T1 (aborted cut must not count; same cut without abort must).
[...]

PR #909, opened 2026-09-17
feat(agent): opencode v2 local compression kernel — no proxy, mode parity with bili pi

ranxianglei:
Fixes #908. Stacked on #850 (same convention as #885: base is the #850 source branch so the stack lands together).

## What

`src/agent/opencode-local.ts` — an opencode v2 plugin that runs the acp-kernel compression pipeline IN-PROCESS, giving opencode the same pure-local mode `bili pi` users get: no proxy hop, no `BILLION_CONTEXT_PROXY`, no MITM, wire traffic straight to the real upstream.

- session registry keyed by conversation id; `session.hook` rebuilds the wire view through the kernel (folded messages, `acp` tags on text parts)
- `chat.params.tools` injects the kernel compress/decompress/acp_status tool set; `chat.parameters` enforces the compress system prompt (with a local tag-format note — the shared prompt teaches proxy-style tags)
- tool execution goes through the kernel, so `acp_status` renders the same panel as `/acp`, panel-first, with its own host identity (`billion-context-opencode-local@version`)

## Fixes riding on this branch (recorded in #906)

- **v2 tool-result cores keyed by call id** — v2's context projection gives tool-result messages no id; `${msg.id}:${i}` collapsed every result onto one ref (`undefined:0`), orphaning each tool call and looping the model (the `acp_status` ×7 / "Tool result missing" bug). Keyed by `tr_` call id now.
- **acp_status panel-first** with own hostLabel through RewriteCtx/AcpStatusCtx + VERSION resolution for dist/agent/*.js depth
- **real echo strip** — stripEchoes gate was a literal `'\\x3cacp'` that never matched real tag bytes; model-fabricated tags now stripped on ingest
- **weak-overflow abort guard** (`src/loop/core.ts`) — a user ESC broke the parse loop with `!sawDone && streamError === undefined`, the exact shape the #498 detector counts; 3 aborts in 15min would learn a shrunken window. Now `&& !signal?.aborted`, mirroring the existing retry guards. Regression: `tests/loop-truncation.test.ts` #887 T1 (aborted cut must not count; same cut without abort must).
- **opencode-local system array**: spread-assignment instead of in-place push — a reused `e.system` or double-fired hook can no longer append duplicate compress prompts.

## Pre-flight

- `npm run typecheck` ✓
- `npm test` — 1673 pass / 0 fail (incl. 11 new `tests/opencode-local.test.ts` tests)
- `npm run build` ✓ (tsup entry includes `src/agent/opencode-local.ts`)

## E2E (Docker-isolated, real opencode v2 + local sglang)

Four postures green: proxy / launcher / native-spawn / pure-local. Compression stress test in-container: model read ~200K tokens of real source, self-triggered compress 7× (105K folded), refs strictly monotonic, no id reuse, recall-from-summary verified, no tag leakage.

(Note: an early local commit accidentally tracked `iso/build/opencode.exe` (207MB rig build artifact); history was rewritten before the first successful push — the branch on origin has never carried it.)


ranxianglei:

[bot] 🏷 Reviewing #909 now. Plan: pull the head branch, diff against the #850 base, run typecheck + full test suite + build on the rebased head, and verify each of the five fixes riding on this branch (tool-result core keying, acp_status panel-first, echo strip, weak-overflow abort guard, system-array spread). Will report findings here with a mergeability verdict.

ranxianglei:

[bot] 🏷 Review complete. I pulled the head, diffed against the #850 base (`a119940`), ran the full pre-flight, audited every riding fix, and fixed two problems **directly on the branch** (pushed, head is now `e6e7abc`).

## Fixed on this branch (2 commits)

**1. `50cfaaf` — restore 3 scripts accidentally deleted.** Commit `e59776b`'s message describes the tr_-keying fix, but its diff also deletes `scripts/update-codex-models-snapshot.mjs`, `scripts/update-registry-snapshot.mjs`, and `tools/analyze-dumps.mjs` (−267 lines). All three exist on master *and* on the #850 base, and `package.json` still wires `npm run registry:snapshot` / `codex-models:snapshot` to the first two — as merged, those npm scripts would be broken. Looks like collateral of the history rewrite that dropped the tracked rig binary. Restored verbatim from the base; content byte-identical.

**2. `e6e7abc` — hex-escape raw `<acp>` tags.** AGENTS.md §3 mandates `\x3c`/`\x3e` escapes for any acp XML in source (so agent editing tools can't strip them) and every other file complies, but `src/compress-tool.ts` (LOCAL_TAG_FORMAT_NOTE string const + doc comment) and `src/agent/opencode-local.ts` (stripEchoes comment) carried literal angle-bracket tags. Converted all 6 occurrences. Runtime string values are byte-identical — verified with typecheck + the three affected test files (57/57 pass).

## Riding fixes — verified

- **tr_ call-id keying** (`opencodeToCore`): tool-result messages get no id in v2's projection, so the old `${msg.id}:${i}` key collapsed every result onto one ref. Keying by `tr_${callId}` is correct; matches the #906 repro.
- **abort guard** (`src/loop/core.ts:624`): `&& !signal?.aborted` added to the weak-overflow note condition; `signal` is the existing param used by the retry guards at :364/:403, so scope is right. The regression test (#887 T1 in `tests/loop-truncation.test.ts`) is genuinely discriminating: same cut stream, 95% usage fallback, aborted run asserts zero "weak overflow signal" hits while the no-abort control asserts ≥1 ("1/3") — i.e. it would fail without the guard.
- **real echo strip**: gate is now the actual tag bytes before `stripAcpTags`; the old literal never matched. Correct.
- **panel-first acp_status + own host identity**: `hostLabel` flows through RewriteCtx/AcpStatusCtx, falls back to the proxy label, and the version walk-up handles `dist/agent/*.js` depth. One subtlety checked: the status handler runs `processTurn({renderTags:"none"})` a second time per status call and discards the returned state — safe, because the kernel's contract is explicit state-in/state-out (pure pipeline over `{messages, state}`), both calls receive identical inputs, and ref assignment is idempotent over an already-assigned map (no number re-issue).
- **system-array spread**: replaces in-place push, fixing unbounded growth when the host reuses `e.system` across requests. Residual: a hook firing twice on the *same* event object could still append once per fire (bounded). I considered a marker-prop guard but rejected it — unknown props risk leaking onto the wire for strict backends.

## Non-blocking notes (no action taken)

- `memory` Map (sid → last CoreMessage[]) in opencode-local never evicts; a long-lived opencode process with many concurrent sessions grows without bound. Proxy-side analog is orphan-gc; consider later.
- Echo stripping covers assistant *text* parts only, not reasoning parts.
- `coreToOpencode` groups by `id.lastIndexOf(":")` — assumes every produced id contains `:` (true today; would silently mis-group if the kernel ever emitted an un-namespaced id).

## Pre-flight & E2E

- `npm run typecheck` ✓ · `npm test`: 1673 pass / 0 fail / 2 skipped (matches your claim) · `npm run build` ✓ (emits `dist/agent/opencode-local.js`)
- Full codex E2E (`ACP_TEST_E2E=1`, 4-phase core): **could not complete in this environment** — the default local sglang upstream refused connections (proxy log shows `ECONNREFUSED` on the forward hop; direct probe of the upstream also fails from here). Not a code signal: the proxy-side log shows the new pipeline executing correctly up to the forward (processTurn renderTags=text-only, nudge gating, session lock). Your Docker-isolated four-posture run + stress test remain the E2E evidence; the manual-dispatch CI workflow is the place to re-confirm with a live upstream secret.

Diff is otherwise clean — every line traces to the feature or the five recorded fixes. Everything verified above passes with my two commits included. Ready for you to merge (merges stay human-side).

中文摘要:修复了分支上两个问题——e59776b 误删的 3 个 snapshot/devtool 脚本(package.json 仍引用,会坏掉 npm scripts,已按 base 原样恢复)和两处源码里违反 hex-escape 规范的裸 acp 标签(运行时字节不变,已验证);五个随行修复逐条核对无误,全量测试通过,本环境缺本地上游跑不了完整 E2E(代理侧日志显示管线执行正常),可以合并。

ranxianglei:
**Closing — superseded by architectural decision (align on the `opencode-native` self-spawn-proxy pattern, i.e. exactly what `bili pi` does).**

Why: this PR's in-process kernel has no wire seam, so it cannot do stream-time echo stripping (`loop/tag-echo-filter.ts`, #206/#717). Real-session forensics (`opencode.db`) showed the model forging `<acp tokens="1" type="text">m00007</acp>` tags that leak into the TUI and the persisted history — `bili pi` never leaks because its self-spawned proxy strips echoes on the stream before the client ever sees them.

The pi-parity architecture already exists on master: `src/agent/opencode-native.ts` (#885) — self-spawned proxy via the same `ensureProxyRunning` bootstrap, per-request URL rewrite through the v2 `session.hook("http.request")` seam (v2's equivalent of pi's fetch patch), `x-bili-plugin` header (plugin mode), `forwardTool` for tools — one-to-one isomorphic with `pi-native.ts`/`pi.ts`.

Salvaged from this PR into #910: the mode-independent loop fix (client abort miscounted as weak overflow, `src/loop/core.ts` + `#887 T1` regression). The `opencode-local.ts` host and its tests are abandoned with this branch.

Issue #911, opened 2026-09-17
opencode v2 native: acp_status tool returns legacy flat report instead of the status panel


ranxianglei:
## Repro
Native mode (self-spawned proxy, `dist/agent/opencode-native.js`): in the opencode v2 TUI, let the model call `acp_status` (or just ask it for context status). The tool output is the legacy flat report:

```
ACTIVE SURFACE: pack=default | host=billion-context 0.1.115
CONTEXT BREAKDOWN
  0 tool (0%) | 47 text (100%) | 0 summaries (0%)
...
Tip: buildStatusReport(...)
```

while the `/acp` slash command renders the rich panel (host label, usage, blocks, nudge) — tool and command disagree.

## Root cause
The v2 plugin forwards `acp_status` via `forwardTool` → the proxy's `/__bili/plugin/tool` executes the kernel's `acp_status`, whose text output is the legacy flat `buildStatusReport`. The rich panel (`buildStatusPanel`) is only served by the read-only status endpoint `/__bili/plugin/status` (`status.panel`), which the `/acp` command consumes.

Panel-first rendering was originally shipped for the in-process local kernel (#909), which is now closed/superseded by the native-proxy alignment — the fix must be re-landed on the v2 mainline.

## Impact
Model-visible status is the legacy format in ALL opencode v2 modes that forward tools to the proxy (native self-spawn, launcher, attach). Confusing next to the `/acp` panel; loses the richer per-section data (host identity, sent vs pruned, nudge reasoning).

## Suggested fix
In `src/agent/opencode-v2.ts` tool execute: for `acp_status` (read-only), read `fetchStatus(base, sessionID)` first and return `status.panel` when non-empty; fall back to `forwardTool` (older proxy / unknown conversation). No proxy-side change; bili pi untouched.

ranxianglei:
> 🤖 Powered by [ework](https://github.com/ranxianglei/ework) · qwen3.8-27b

[bot] 🏷 Verified — this fix already exists and is ready to merge: **https://github.com/ranxianglei/billion-context/pull/912** (head `96368b2`, base `2026-09-15_opencode-native-plugin`, clean / no conflicts).

[...]

PR #912, opened 2026-09-17
fix(agent): opencode v2 acp_status renders the status panel (panel-first)

ranxianglei:
Fixes #911

## What
In native/launcher/attach modes the v2 plugin forwarded `acp_status` to the proxy's tool endpoint, which returns the **legacy flat report** — while the `/acp` slash command renders the rich panel. Tool and command disagreed.

Now `acp_status` (read-only) reads the panel from the proxy's status endpoint (`status.panel`, same source as `/acp`) and returns it; falls back to `forwardTool` when no panel is served (older proxy / unknown conversation). No proxy-side change; bili pi untouched.

Re-lands the panel-first UX from the closed #909 on the v2 mainline (align on the native-proxy architecture; the in-process kernel is superseded).

## Test
- Updated `tests/opencode-v2.test.ts`: panel served → tool returns panel; status 404 → falls back to forwarded tool result.
- Full suite on this line: 1662 tests, 0 fail.

## Pre-flight
- `npm run typecheck` ✅ · `npm run build` ✅

Verified live in the oc2 rig container: native self-spawn proxy + model request routed + `status.panel` served for the live session.


ranxianglei:

[bot] 🏷 Picked up for review. This is a [PR] mirror of #912 (head `2026-09-18_opencode-v2-panel-first`, base `2026-09-15_opencode-native-plugin`). I will NOT re-do the work — I'm verifying: (1) duplicate screening vs open issues/PRs incl. the superseded #909, (2) diff cleanliness vs its base, (3) code review of the panel-first `acp_status` path, (4) local pre-flight (typecheck / full test suite / build) on the head. Will report findings here shortly.

ranxianglei:

[bot] 🏷 Review complete — **verified, no changes needed, ready for human merge.**

**What I verified (independently re-ran everything at head `96368b2`):**

- **Duplicate screening:** #909 (the original panel-first work) is closed unmerged on this same base branch — this is the documented re-land, not parallel work. No other open PR/issue covers panel-first `acp_status` on the v2 line. Predecessors #875/#882/#885 are already in the base.
- **Hot-file contention:** concurrent open PR #842 (same base) touches only `src/agent/opencode-native.ts` + two new test files — zero file overlap with this PR's `src/agent/opencode-v2.ts` / `tests/opencode-v2.test.ts`; semantically orthogonal (route gating vs tool execution). Either merge order is safe.
- **Diff cleanliness:** single commit, 2 files +20/-1, every line on-purpose, no whitespace churn, no lock/generated-file noise, `version` untouched.
- **Code review:** the new block (`src/agent/opencode-v2.ts:212-216`) reads `status.panel` via the same `fetchStatus` the `/acp` command handler uses (`:255`) — tool and command now agree by construction, single source of truth. Fallback semantics are exact: non-empty panel → return it; anything else (404 unknown conversation, proxy down, empty panel, older proxy) → falls through to `forwardTool`, which is byte-identical to pre-PR behavior including the caught-error guided text. Sits inside the existing try/catch; no new failure mode. §6 both-modes check: agent-side rendering only, no proxy wire path touched, bili pi untouched.
- **Tests:** fixture covers both branches — panel served → tool returns `ACP-PANEL-OK` without forwarding; status 404 → falls back and the assertion confirms the forward actually happened (`toolCalls.at(-1).tool === "acp_status"`). Deterministic (port via `listen(0)`).
- **CI:** the check-run on head ("real codex vs deterministic fake upstream") = success.
- **Local pre-flight at head:** `npm run typecheck` ✅ · full suite **1662 tests / 0 fail** (1660 pass + 2 skipped, matches your claim) · `npm run build` ✅ · acp-kernel pinned exact `0.0.73` in package.json matching installed node_modules.
- **Base/topology:** correctly based on `2026-09-15_opencode-native-plugin` per the stack convention — this v2 mainline has no common ancestor with master by design, so no rebase to master applies here.

One non-blocking observation for the record: an `ok:true` response with an *empty-string* panel also falls back (covered by the `length > 0` guard) but has no dedicated test case — same code path as the 404 fallback, so I didn't hold the review for it.

**中文摘要：** 让 opencode v2 的 `acp_status` 工具与 `/acp` 命令同源渲染状态面板（优先读 `status.panel`，无面板时回退到原有转发路径），消除了工具输出与命令面板不一致的问题；diff 干净、双分支测试齐全、本地全量预检（typecheck/1662 测试/build）全部通过，可以合并。

Issue #953, opened 2026-09-18
[dsh bundle] boot crash window: stale pnpm metadata installs 0.1.117 copy while global 0.1.120 declares the bundle


ranxianglei:
**Boot crash window after 0.1.120 release: stale pnpm metadata installs a 0.1.117 profile copy while dsh bundles detection hits the global 0.1.120.**

## Repro (real, hit on two machines)

1. Machine has a **global** billion-context >= 0.1.120 (npm i -g)
2. pnpm registry-metadata cache is still fresh from before the release
3. `dsh plugin --profile <name> add billion-context` → pnpm resolves `^0.1.117` from cache (output shows `resolved 1, reused 1, downloaded 0`) → profile copy = 0.1.117 (no `exports["./dsh"]`)
4. dsh `exportsPatch` detects the bundle via `resolveBundleDir` which is **installation-anchor-first** → hits the GLOBAL 0.1.120 (has the `dsh.bundle` declaration) → profile joins `dsh.profile.bundles`
5. `dsh web` → cordis loader imports `billion-context/dsh` **from the profile dir** → 0.1.117 has no such export → `ERR_MODULE_NOT_FOUND` → **plugin tree fails to load, dsh will not boot**

## Impact

- Pure dsh users (no global bili): detection hits the profile copy (no declaration) → plain-dependency warning, NO crash — self-heals on cache expiry / re-add
- Machines with global >=0.1.120 + stale cache + add inside the window: **hard boot crash**, stable across remove/add (same cached resolution)
- Window closes as pnpm caches expire; new installs resolve 0.1.120+ which carries the export. Only machines that added during the window keep a 0.1.117 copy until they update.

## Fix / workaround

- Immediate: `dsh plugin --profile <name> add billion-context@0.1.120` (pin) or `@latest` once caches refresh — verified, boots fine
- Nothing we can ship in-package: the mismatch is dsh-side (detection resolves installation-first, loading resolves profile-first). A future bundled-yml change cannot rescue old copies that lack the subpath export.
- Doc note already in PR #950 READMEs ("Requires a published release that carries dsh.bundle.patch.yml"); could add "pin @latest if dsh fails to boot after add". Closing as windowed/obsolete once caches age out — keep as the canonical record.

Found while verifying the 0.1.120 release (#949 / PR #950).

ranxianglei:
> 🤖 Powered by [ework](https://github.com/ranxianglei/ework) · qwen3.8-27b

[...]

PR #950, opened 2026-09-18
feat: dsh-side install channel — zero bili commands for dsh users (#949)

ranxianglei:
Fixes #949

## What

`dsh plugin --profile <name> add billion-context` installs the npm package into the profile via pnpm; dsh reads the package's `dsh.bundle.patch` manifest declaration and mounts it as a bundle layer automatically — same plugin, same behavior as `bili plugin install dsh`, but the user never runs a bili command.

- `exports["./dsh"]` -> `./dist/agent/dsh-native.js` (new public subpath, like `./server`)
- manifest `"dsh": {"bundle": {"patch": "./dsh.bundle.patch.yml"}}` + the yml in `files` (insert `bili-native` + `compaction-basic auto:false`)
- **Mutual exclusion (hard requirement):** cordis rejects duplicate loader entry ids across layers, so a bundle-installed profile must not also receive the managed block / `--patch` overlay. `dshBundleInstalled()` detects profiles whose manifest lists billion-context in `dsh.profile.bundles`; install/remove skip them with pointers to the dsh-side commands, `plugin status` reports them, and `dshNativeInstalled()` (the `bili dsh` launcher overlay gate) recognizes them
- README / README.zh-CN: new lane documented (tables + dsh sections)

## Verified end-to-end (dsh 0.1.5-rc.2, local tarball via `dsh plugin add`)

1. pnpm installs into the profile; dsh reconciles `dsh.profile.bundles` += billion-context
2. `--dump-config` shows the `billion-context` layer: `bili-native` insert + compaction patched `auto:false`
3. live boot: plugin tree loads, bili-native **self-spawns the proxy** (bili.log `acp-proxy listening`), parent-pid watchdog flushes on teardown
4. `bili plugin install dsh` + `plugin status` correctly skip/report the bundle profile

## Known interaction (documented, not ours to fix)

dsh resolves bundle dirs from the dsh-install anchor **first**; on machines with a **global** billion-context <=0.1.119 (no declaration yet) that copy shadows the profile one and dsh treats it as a plain dependency. Self-heals once the global install is >= this release; pure dsh users (no global bili — the target audience) resolve the profile copy directly.

## Preflight

- `npm run typecheck` OK
- `npm test` 1881 pass / 0 fail / 2 skip (dsh suite 10->14: bundle detection + skip/report roundtrip)
- `npm run build` OK; `npm pack --dry-run` shows `dsh.bundle.patch.yml` in the tarball

Note: needs release >=0.1.120 to be usable from the npm registry.

PR #979, opened 2026-09-19
feat: claude native posture — managed settings block + SessionStart hook

ranxianglei:
Fixes #964

## What

Claude Code has no in-process extension point (hooks and MCP servers are child processes), so the native posture is a documented **hybrid**, per the design agreed in the issue:

1. **Installer** (`bili plugin install claude`) writes a managed block into `~/.claude/settings.json` (`CLAUDE_CONFIG_DIR` honored): `env.ANTHROPIC_BASE_URL` pinned to a **stable** loopback port `/bili/`-wrapping the upstream (`BILI_CLAUDE_UPSTREAM` / `claude.anthropicBaseUrl` discovery), `env.DISABLE_AUTO_COMPACT=1`, and a `SessionStart` hook entry — plus the existing user-scope MCP shell, now pinned to the same stable port. Port resolution: `BILI_CLAUDE_NATIVE_PORT` > config `claude.nativePort` > default 48787. Pure JSON merge/strip: foreign keys never touched, `.bili-bak` pre-install snapshot, `bili plugin remove claude` restores exactly. Malformed settings.json → refused, file untouched. `BILI_NATIVE_CLAUDE=0` → install refused.
2. **Hook** (`dist/claude-native-bootstrap.js`, new bin+tsup entry): fires before claude's first model request; attaches to a healthy proxy on the stable port or spawns one detached whose parent-pid watchdog tracks **claude's own pid** (the hook's ppid — the hook exits immediately), so the proxy lives and dies with the session. Never fails claude: every error logs to stderr and exits 0. Gates: `BILLION_CONTEXT_PROXY`/`BILI_PROVIDER_REWRITES` set (someone else owns routing) → exit 0; `BILI_NATIVE_CLAUDE=0`/`BILLION_CONTEXT_PLUGIN=0` → spawn a **passthrough** proxy on the same port so the now-static URL still answers and claude stays functional (compression off).
3. **Launcher coexistence**: `bili claude` on a machine with the native block installed overrides the static URL with its own ephemeral proxy (`env` + `--settings` belt-and-braces) and the hook stays dormant for that session.
4. **#839 hardening**: `pluginStatusAll`'s claude row never throws — a broken `.claude.json` degrades to `not installed` instead of `error:`.

## Fix along the way

The hook passes `scriptPath` (sibling `dist/index.js`) to `ensureProxyRunning` — its default (`process.argv[1]`) would re-invoke the hook script itself as the proxy (infinite self-spawn; caught by the e2e test).

## Tests

[...]

PR #980, opened 2026-09-19
feat: unify the dsh install lane on dsh's own plugin channel (#966)

### LINKED ISSUE #966: [dsh] unify install lanes: `bili plugin install dsh` should go through the bundle channel + auto-update rewrite
## Goal (planned, later — do not start before the runtime-info wave lands)

Today dsh has TWO install lanes that can fight each other (#953):

1. **Managed block** (what `bili plugin install dsh` writes today): a `# bili begin/end` block in every `~/.dsh/profiles/*/cordis.patch.yml` pointing `file://` at the global bili install. Follows global bili updates for free, but requires the `bili` command and competes with lane 2 for the same patch layer.
2. **Bundle channel** (PR #950, dsh-native): `dsh plugin --profile <name> add billion-context` — pnpm-installs the package into the profile, dsh's own plugin inventory owns it. No bili needed, but the profile copy does NOT follow global bili updates (version-drift window: a newer global bili with an older profile bundle can crash — seen with 0.1.116/0.1.117).

## Design direction

- Retire the managed-block lane: `bili plugin install dsh` should EXECUTE `dsh plugin --profile <name> add billion-context` for each selected profile instead of writing patch blocks (single lane = the inter-management problem disappears).
- Close the update gap: after `src/update.ts` completes a global self-update, scan `~/.dsh/profiles/*/package.json` for a `billion-context` dependency and re-run the bundle install (or `pnpm update`) so profile copies stay in lockstep with the global version. This also closes the #953 crash window permanently.
- Local/dev installs must keep working: `dsh plugin add <path-to-tgz>` is supported (verified in #950 E2E) — the installer should pass the same source it was installed from (registry version or local tgz).

## Acceptance

- `bili plugin install dsh` produces the same end state as `dsh plugin --profile X add billion-context`; no managed blocks written
- `bili plugin remove dsh` uninstalls through the same channel
- After a global bili self-update, profile bundle copies are refreshed (verified with an old-global/new-global container drill)
- README (en/zh) dsh install sections updated to describe the unified lane

## References

- #941 / PR #942 (dsh three-mode, bundle channel + managed block coexistence)
- #950 (bundle channel E2E: `declares no dsh.bundle` warning, resolveBundleDir installation-first semantics)
- #953 (version-drift crash between global and profile copies)

ranxianglei:
Closes #966

## What changed

**One lane.** `bili plugin install dsh` no longer writes managed blocks into every `~/.dsh/profiles/*/cordis.patch.yml` — it now executes `dsh plugin --profile <name> add billion-context` for each existing profile, i.e. the exact command of the bundle channel (#950). Same end state from either entry point; `bili plugin remove dsh` uninstalls through the same channel (`remove`).

- The spec follows how bili itself is installed (#925, same rule as pi/opencode): npm-form installs pass the registry name, checkout/dev builds pass their absolute path (pnpm turns it into a `link:` dep, so local work stays live).
- **Legacy migration:** pre-#966 managed blocks (`# bili begin`/`# bili end`) are stripped on both install and remove — user entries and comments survive, an emptied file gets its placeholder `[]` back. `bili plugin status` reports legacy blocks with a migrate hint instead of a bare "installed".
- `dshNativeInstalled()` (consumed by the `bili dsh` launcher to skip its `--patch` overlay) still returns true for legacy-block profiles, so users mid-migration don't get a duplicate `id: bili-native` insert that would hard-fail dsh boot.

**Update lockstep (#953).** After a global self-update succeeds, `checkForUpdate` calls the new `refreshDshProfileBundles()`: it scans `~/.dsh/profiles/*/package.json`, and for every registry-pinned `billion-context` dependency runs `dsh plugin --profile <name> add billion-context@<new-version>` so the loaded plugin and proxy never drift apart again. Profiles pinned to a local source (`link:`/`file:`/…) are left alone. Best-effort by contract: per-profile failures are logged and never fail the update itself.

## Structure

- New `src/dsh-channel.ts` — dependency-light (fs/path/child_process + client-config): profile discovery, legacy-block migration helpers, manifest-state queries, spawn planning, and the refresh. Shared by `plugin-install.ts` and `update.ts` without pulling jsonc-parser into the updater's graph.
- Spawn plumbing mirrors the launcher's #679 rules: win32 bare names / `.cmd` shims ride `comspec /d /s /c` with `windowsVerbatimArguments` (spawnSync — execFileSync options don't accept the flag); resolved `.exe` and posix spawn direct. Sync for the CLI path, async for the update path (a blocking pnpm in the updater would freeze the proxy event loop).
- Tests run under injected runners (`_setDshRunnersForTest`), so they're deterministic across the CI OS matrix without needing pnpm or a real `~/.dsh` — including the Windows comspec-wrap shape, failure propagation (`pnpm not found` stderr surfaces verbatim), legacy migration roundtrips, and refresh isolation (one profile failing doesn't stop the rest).
- README en/zh updated: client table, Option-1 block, and the dsh section (two bullets → unified-lane bullet + auto-update lockstep bullet).

## Pre-flight

- `npm run typecheck` — clean
- `npm test` — 1924 pass / 0 fail (2 skipped = e2e gate)
- `npm run build` — ok

Note: touches `src/update.ts` (load-bearing infra per §7.4) — human merge.


ranxianglei:

[bot] 🏷 Reviewing #980 now — branch `2026-09-19_dsh-unified-install-lane` (bc83ae3) fetched locally. Diff scope looks tight (7 files, no lock/version churn); master has only the v0.1.121 release bump since the base, so rebase should be trivial. Verifying one suspected corner-case regression in the legacy-block migration (comment preservation in emptied patch files), then running the full pre-flight on the rebased head. Full verdict to follow.

ranxianglei:

[bot] 🏷 **Review complete — one small regression found and fixed directly on the branch; otherwise ready for human merge.**

## What I verified

**Scope & diff hygiene.** 7 files, no lock/version churn, every line on-purpose. README en/zh stay in sync (client table, Option-1 block, dsh section rewritten as unified-lane + lockstep bullets, no stale managed-block references left). Base check: master moved only the v0.1.121 release bump since the PR branched — no semantic contention with other open PRs on the touched files (`plugin-install.ts`, `update.ts`).

**Design (root-cause vs symptom).** The unification targets the real defect: two lanes writing different end states for the same plugin, plus the #953 drift window where a stale profile copy crashes against a newer global kernel. Driving dsh's own channel per profile makes the bili form a pure convenience wrapper over the canonical command — same end state either way, and the #953 refresh closes the drift at its source (profile manifests), not by patching symptoms. Sound.

**Mechanics spot-checks.**
- `update.ts`: `installViaTarball` is the single install path in `checkForUpdate`; the `refreshDshProfileBundles(latest, loggerLog)` hook sits in its sole success branch, so every successful self-update triggers the refresh. Best-effort contract honored: per-profile `try/catch` warns, never rejects; missing profiles root returns silently; local-source pins (`link:`/`file:`/…) are skipped by `isRegistryDepSpec`.
- Spawn plumbing correctly mirrors the launcher's #679 rules: win32 bare names/`.cmd` shims ride `comspec /d /s /c` + `windowsVerbatimArguments` (sync path uses `spawnSync` because `execFileSync` options reject the flag; async path uses promisified `execFile`, which accepts it); resolved `.exe` and posix spawn direct.
- `dshNativeInstalled()` still true for legacy-block profiles → mid-migration users don't get a duplicate `id: bili-native` insert (cordis hard-fail avoided). Marker constants are stable across the move.
- Tests are deterministic under injected runners (`_setDshRunnersForTest`, always reset in `finally`), covering the comspec-wrap shape, verbatim stderr propagation (`pnpm not found`), ENOENT → "dsh CLI not found", migration roundtrips, and refresh isolation (one failing profile doesn't stop the rest).

**Pre-flight (run by me, on a head containing current master):** typecheck clean · `npm test` 1925 pass / 0 fail / 2 skipped (e2e gate) · build ok. To keep your feat commit intact I landed the fix below as a plain fast-forward on top rather than rewriting the branch base.

## Fixed on the branch (commit `d0971dd`)

**Comment loss in legacy-block migration (corner case).** The relocated `restoreDshPatchPlaceholder` regressed against the pre-#966 code it replaced: when a `cordis.patch.yml` contained *only* user comments plus the managed block, stripping emptied it of "meaningful" lines and the function wrote back the standard header — silently deleting the user's comments. Your PR contract says "user entries and comments survive"; the old `plugin-install.ts` implementation preserved comments explicitly, the refactor dropped that. Fixed by restoring the comment-preserving placeholder logic, plus a regression test (`stripLegacyManagedBlock: preserves user comments when nothing meaningful remains`). Both dsh test files: 26/26 green.

## Minor findings (report only, not changed here)

1. `.gitignore` covers `tmp/` but not `.tmp/`. This sandbox points `$TMPDIR` at the workspace `.tmp/`, so any `git add -A` in an agent session sweeps node compile-cache into commits (it bit me once this session; caught and cleaned before final push). Worth a one-line ignore entry or daemon-side staging discipline — your call.
2. `bili plugin install dsh` with an existing-but-empty profiles dir reports "installed billion-context into 0 dsh profile(s)" — accurate but slightly odd phrasing vs the old "no profile directories found" note. Cosmetic.

Per §7.4 this stays a **human merge**: it touches `src/update.ts` (load-bearing updater infra) and changes install/remove wire behavior for dsh profiles. Everything above the bar is verified green on the current head.

中文摘要:修复了旧版受管块迁移时会丢失用户注释的回归(清空后的 patch 文件被写回标准头而非保留原注释),已在 PR 分支直接补了修复和回归测试并验证全绿(typecheck/1925 测试/build 均通过);统一泳道与更新锁步的设计本身成立,因涉及 update.ts 关键基础设施,需人工合并。

ranxianglei:
Acceptance pass — two fixes pushed (d620f33, rebased onto master):

1. **Rebase**: the branch predated #979's Windows `.cmd` fixes, so `test (windows-latest, 22/24)` were red on the old `execFileSync(claude, …)` EINVAL. Rebased cleanly onto master (57fbb90); conflicts were just the child_process import (took the superset) and two README dsh/claude table rows.
2. **Hermetic test fix**: `dshNativeInstalled: true iff any profile…` leaked into the developer's real `~/.dsh` on its final "no profiles root" assertion — green on clean CI runners, guaranteed red on any machine that actually has a dsh install (the bare `dshNativeInstalled()` ran after `withEnv` restored `DSH_HOME`). Wrapped it in `withEnv({DSH_HOME: <sandbox>/absent})`. Local: full suite 1948 pass / 0 fail / 2 skipped (e2e gate), typecheck clean, build clean.

CI re-running on the new head.

ranxianglei:

[bot] 🏷 Both fixes from your acceptance pass check out, but the re-run finished **red again** on both Windows legs — and I've verified the fix you just pushed (`7129a07`) resolves it. Details:

**Your two fixes — verified.** Rebased head = `35b9199` (feat) + `9226080` (my comment-preservation fix from the first review round, intact) + `d620f33` (hermetic `DSH_HOME` wrap). Pre-flight on `d620f33`: typecheck clean, **1948 pass / 0 fail / 2 skipped** (matches your local numbers exactly), build ok. The leak fix is correct — the bare `dshNativeInstalled()` call did run after `withEnv` restored `DSH_HOME`.

**Why Windows was still red (run 35426028840, jobs 105852005840 / 105852005888):** 5 failures, all dsh tests from this PR — `dsh install drives…`, `dsh remove uninstalls…`, `planDshSpawn: posix passthrough`, and both `refreshDshProfileBundles` cases. Root cause: the `_setDshRunnersForTest` seam sits **behind** `planDshSpawn`, so on win32 the injected runner receives comspec-wrapped plans while four assertions compared against the logical argv. From the node-22 log verbatim:

```
+ '/d /s /c "dsh plugin --profile headless add D:\\a\\billion-context\\billion-context"'
- 'plugin --profile headless add D:\\a\\billion-context\\billion-context'
```

The `posix passthrough` case had a second, independent defect: it never passed an explicit `platform`, so on win32 it planned with the host platform and wrapped too. (The recorders' `args.indexOf("--profile")` extraction also silently misfired when wrapped.)

**Race disclosure:** I diagnosed the same root cause in parallel and prepared an overlapping fix — my push was rejected non-fast-forward because `7129a07` landed first. So I audited yours instead of forcing a merge of approaches: it covers all five failures, and it's the better design — assertions stay readable logical strings, the unpacking lives in the recorder (`planTokens`, plus a dedicated unit test pinning both directions), whereas my version normalized expectations through the planner. I dropped my duplicate local commit (never pushed; nothing to clean up remotely).

**Verification at `7129a07` (current tip):**
- typecheck clean · full suite **1949 pass / 0 fail / 2 skipped** (e2e gate; +1 = the new `planTokens` test) · build ok
- This sandbox can't execute win32, so I replayed all five red assertion flows through real `planDshSpawn(…, "win32")` output — including the exact argv from the CI logs — into your `planTokens` unpacker: every case round-trips to the precise logical argv, profile/action extraction correct. Also confirmed the basename-based `cmd.exe` detection holds for full-path `COMSPEC` under win32 path semantics (`path.win32.basename("C:\WINDOWS\system32\cmd.exe") → "cmd.exe"`).
- One documented limitation, no action needed: the unpacker splits on whitespace, valid only for space-free tokens — exactly what its comment states ("no spaced test tokens"), and every fixture complies.

**Verdict:** branch tip `7129a07` is ready subject to the two windows-latest legs going green on the next run. Still **human merge** per §7.4 — `src/update.ts` remains load-bearing territory, and the change alters wire/install behavior across both compression modes.

中文摘要：Windows CI 双腿在 7129a07 之前仍是红的（5 个 dsh 测试失败：注入的测试 runner 位于 planDshSpawn 之后，win32 下记录的是 comspec 包装参数，断言写的却是逻辑参数；另有 posix passthrough 测试未传显式 platform）；已审计并用显式 win32 计划重放了全部 5 个失败断言的精确形态，确认 7129a07 的平台中性修复完整正确、本地全绿（typecheck / 1949 pass / build），待 CI 重跑转绿后即可人工合并。

Issue #987, opened 2026-09-19
context window problem continues


walsycommunity:
<img width="468" height="130" alt="Image" src="https://github.com/user-attachments/assets/22cccafb-8599-4065-992d-42293ee04a04" />
<img width="448" height="128" alt="Image" src="https://github.com/user-attachments/assets/840b32a7-9a0a-4990-b5ea-2274933e00b0" />
<img width="478" height="119" alt="Image" src="https://github.com/user-attachments/assets/76f4683b-8d12-48ac-8602-b52c1c3acd17" />
<img width="579" height="27" alt="Image" src="https://github.com/user-attachments/assets/1452805e-f9fe-40a8-a07d-6a880f78b3d0" />
<img width="471" height="94" alt="Image" src="https://github.com/user-attachments/assets/e1abab74-ea0c-4cca-8018-b49bb055ded1" />

like Context (session accounting, host footer scale): 103% (34k / 33k)
Context (session accounting, host footer scale): 0% (0 / 250k)
Context (session accounting, host footer scale): 14% (35k / 250k)
Context (session accounting, host footer scale): 113% (41k / 36k) The context window size keeps changing—starting with the value I initially selected (I tried various figures like 50,000, 250,000, and 1,000,000)—but then it automatically sets the limit based on the context size generated at that moment.So, while the context window I initially set (250k) is active, something happens where it essentially decides—on its own—that your limit is, say, 18k if you happen to be at 18k usage at that moment.  

like:⚠ Retrying in 4s · attempt 7 · context ~18231 tokens exceeds the model window 16898 (model=subagent), with 1 compressible range(s) still visible and preflight compression could not bring it under: no range could be compressed across 1 viable range (each was below minCompressRange, had an unusable summary, or failed to apply) Last unusable summary: the upstream returned a non-SSE body with no summary text (first 200 bytes: {"id":"chatcmpl-1789806312367","object":"chat.completion","created":1789806312,"model":"muse-spark-1.3-contributor","choices":[{"index":0,"message":{"role":"assistant","content":""},"finish_reason":"s).. The over-window payload was NOT forwarded. Preflight will not call the upstream again for the next 5m for this identical request. Change the request or wait for the cooldown before retrying. ım using 9router+opencode2 . 

[...]

Issue #969, opened 2026-09-18
model window problem(I'm using opencode2 and 9router)


walsycommunity:
Error: context ~19344 tokens exceeds the model window 18320 (model=main), with 1 compressible range(s) still visible and preflight compression could not bring it under: no range could be compressed across 1 viable range (each was below minCompressRange, had an unusable summary, or failed to apply) Last unusable summary: the upstream returned a non-SSE body with no summary text (first 200 bytes: {"id":"chatcmpl-1789752825981","object":"chat.completion","created":1789752825,"model":"muse-spark-1.3-contributor","choices":[{"index":0,"message":{"role":"assistant","content":""},"finish_reason":"s).. The over-window payload was NOT forwarded. Preflight will not call the upstream again for the next 5m for this identical request. Change the request or wait for the cooldown before retrying.

Error: context ~18985 tokens exceeds the model window 16161 (model=main), with 1 compressible range(s) still visible and preflight compression could not bring it under: no range could be compressed across 1 viable range (each was below minCompressRange, had an unusable summary, or failed to apply) Last unusable summary: the upstream returned a non-SSE body with no summary text (first 200 bytes: {"id":"chatcmpl-1789752921605","object":"chat.completion","created":1789752921,"model":"muse-spark-1.3-contributor","choices":[{"index":0,"message":{"role":"assistant","content":""},"finish_reason":"s).. The over-window payload was NOT forwarded. Preflight will not call the upstream again for the next 5m for this identical request. Change the request or wait for the cooldown before retrying.

ranxianglei:
> 🤖 Powered by [ework](https://github.com/ranxianglei/ework) · qwen3.8-27b

Traced this end-to-end against current master (v0.1.120). Short version: **this may not be a real over-window situation — the window bili resolved for `model=main` (18320) is far below what this model family actually advertises, so preflight fires and then trips on an empty summary.** Two layers, in order:

## What I verified (against `src`, not dist)
- The message is assembled in `src/server.ts:3020-3035` (`failFast`) from a `preflightCompress` result; the `Last unusable summary: … non-SSE body …` clause comes from `src/preflight.ts:547` (`diagnoseEmptySummary`). Chain: context ≥ resolved window → preflight walks the 1 viable range → the summarization call returns a JSON body with empty content → unusable → fail-fast 502 + 5-min dead-end cooldown (`src/preflight.ts:1006`, `src/server.ts:3161-3167`).
[...]

PR #988, opened 2026-09-19
fix: size preflight summary output to window headroom + name empty JSON completions (#987)

ranxianglei:
## What

Two preflight fixes for #987 ("context window problem continues", 9router + opencode v2):

1. **Summary calls now fit inside the model's window.** Preflight always requested `max_tokens=32768` regardless of the model's window. Upstreams that enforce input+output ≤ window (muse-spark behind 9router) answer such calls with an **empty completion** (`content:""`) instead of an error, so every summary attempt is unusable → "no range could be compressed" → client retry loop. Now clamped to `window − input estimate` (floor 64, 256-token scaffolding margin); unchanged when headroom is ample or unknown.
2. **Precise diagnostics for that failure mode.** `diagnoseEmptySummary` no longer reports "non-SSE body" for plain-JSON empty completions; it names the shape with `finish_reason`/`stop_reason`/`status` and the model id the upstream **answered as** — which exposes alias mismatches (request said `subagent`, upstream answered `muse-spark-1.3-contributor`).

## Relation to PR #992 (owner's removal work)

#992 removes the runtime window-learning mechanism (the "window keeps changing" symptom) — that half is superseded by it and no longer in this PR. This fix is **orthogonal**: zero file overlap with #992, independently mergeable in either order, and still needed after it — once the window is purely declarative, a correctly-declared small window (e.g. `models.subagent.context`) would otherwise still dead-end preflight via the 32k `max_tokens` request above. The clamp reads the effective declared window (`deps.config.modelContextLimit`).

## Tests

- `tests/preflight-window-clamp.test.ts` (new): clamp engages at an 8k window — invariant `input + max_tokens ≤ window` on every summary call — and stays at 32768 with ample headroom (control).
- Value-based `max_tokens === 32768` summary-call detection in five test files (#568 hold, wire-exit-gap cells, #781 images, #663 max-output-tokens, #853 cap) switched to shape-based detection / relaxed bounds so the clamp doesn't break them.
- New unit assertions for the diagnostics: openai/anthropic/responses empty-completion shapes + shape guard for unrelated JSON.

## Pre-flight (head `df840a6` on master `ea2b33f`)

typecheck ✅ · full suite 2000 pass / 0 fail / 2 skip (e2e gate) ✅ · build ✅

Addresses #987.


ranxianglei:

[bot] 🏷 Received. This is a [PR] wake-up — I will NOT redo the work. Plan: locate the PR, pull its branch, review the diff against current master, re-run typecheck / full test suite / build on the rebased head, check CI status, then report conclusions here.

ranxianglei:

Reviewed PR #988 (`2026-09-19_preflight-window-clamp`, head `0e3ff5a`) — verified against current master `519634b`: single commit, 0 behind / 1 ahead (fully rebased), mergeable. I did not modify anything; this is a review + re-verification report.

**Code review — both changes hold up:**

1. **Window clamp** (`src/preflight.ts:287-293`): the window source is correct — `deps.config.modelContextLimit` is the per-request `reqConfig` (server.ts:1689), i.e. *after* learned-limit application (#969/#976), codex alignment, and output-headroom reservation, so the clamp sees the same effective window the rest of the pipeline uses. Behavior boundaries verified: unknown/≤0 window → unchanged; ample headroom → byte-identical 32768; headroom ≤ 0 → unchanged (clamping can't save a call whose input alone doesn't fit) and the failure now gets the precise diagnosis from change 2 instead of looping on "non-SSE body". Single call site (`preflight.ts:694`) — absorb/acp-status/side-request paths issue no summary LLM calls of their own, so nothing else needs clamping. Mode-neutral: only the proxy's internal summarization payload (`max_tokens`) changes; client-facing wire shape is untouched in both plugin and proxy modes (§6).
2. **Diagnostics** (`src/preflight.ts:563-591`): branch ordering inside `diagnoseEmptySummary` is safe — error-signal extraction and the SSE scan run first, so no existing diagnosis is shadowed; the shape guard returns null for unrecognized JSON and keeps the old message exactly. The `answered as model=` field is the right clue for alias mismatches (`subagent` → `muse-spark-1.3-contributor`).

**Tests:** new `tests/preflight-window-clamp.test.ts` asserts the invariant per-call (`max_tokens >= 64` AND `input + max_tokens <= window`) plus proves the clamp engages at 8k, with a 100k control asserting 32768 retention — deterministic (ephemeral ports, estimator-consistent bounds). Diagnostics tests cover all three wire shapes + the shape guard. The fixture updates (#568 hold, wire-exit-gap cells, #781, #663, #853) swap brittle literal `max_tokens === 32768` detection for shape-based detection / relaxed bounds — justified, since legitimate clamping breaks that literal; #853 opts out via a large window so it isolates the registry-ceiling clamp.

**Verification (all on rebased head `0e3ff5a`):**
- Local: typecheck ✅ · full suite **1997 pass / 0 fail / 2 skip** (e2e gate) ✅ · build ✅
- GitHub CI on `0e3ff5a`: all green — CI matrix (ubuntu/windows × node 22/24), build-artifact, version-guard, e2e-fake-codex (zero tokens), codex overlay.
- Real-upstream codex e2e is manual-dispatch-only (needs a live upstream); the zero-token fake-codex e2e that passed covers the pipeline mechanics. Not blocking.

**Non-blocking observations:**
- When 0 < headroom < 64 the floor still requests 64 tokens, which can exceed the residual headroom — deliberate (a degenerate 1–60-token summary is useless), strictly better than before, and such a call now fails with the precise empty-completion diagnosis. Fine as-is.
- The input estimate uses heuristic `defaultCountTokens` (chars/4, CJK per-char); dense code/JSON could in theory undercount by enough that input+output still slightly exceeds the window after clamping. Best-effort by nature, mitigated by the 256-token margin, and covered by the new diagnosis if it happens. No change needed.

**Conclusion:** mergeable — clean single-purpose diff (8 files, +244/-25), regression tests for the exact reported failure mode, green everywhere. Per repo rules the merge itself stays with you: https://github.com/ranxianglei/billion-context/pull/988

中文摘要：核查了预检压缩在小型窗口下请求 32k 输出导致上游返回空补全、压缩死锁的修复（按窗口余量收紧 max_tokens + 精确的空补全诊断，含 finish_reason 与上游实际应答模型名）——代码审查、本地三件套与 GitHub CI 全绿，测试覆盖了报告的失败模式，可以合并。

PR #1009, opened 2026-09-19
fix: detect silent client history rewrites; archive+prune at the boundary, normalize reversed ref ranges (#1001)

### LINKED ISSUE #1001: 模型切换后压缩块跨模型失配，新模型收到 61.3 万 token 全量上下文
模型切换后压缩块跨模型失配，新模型收到 61.3 万 token 全量上下文（preflight 兜底耗时 234s）；另发现 preflight 反向区间与 0.1.117 行格式 compress 回归

# 环境

- bili: 0.1.117（Windows，`node dist/index.js start --no-auto-update --debug`）
- client: `opencode/999.0.0 ai-sdk/provider-utils/4.0.23 runtime/bun/1.3.14`，chat/completions 协议
- 同一 `x-session-id`（ses_f47d4a85…）贯穿全程，模型中途从 Kun（有效窗 226,000 = 258,000−32,000）切到 glm-5.3-flashx（有效窗 352,000 = 384,000−32,000）

# 复现步骤

1. 模型 A（Kun）在同一会话内多轮成功 compress，累计节省 ~23.5 万 token（48,229 + 37,214 + 38,292 + 56,929 + 55,372），转发视图稳定在 4-5 万；
2. 中途切到模型 B（glm-5.3-flashx）；
3. B 的第一个请求：

```
2026-09-19T10:31:30.517Z processTurn: 567 msgs, renderTags=text-only, 522 text tagged
2026-09-19T10:31:30.519Z nudge idle: usage=19% (67997/352000) ... pendingT1=542575/60000 model=glm-5.3-flashx
2026-09-19T10:31:30.642Z context 613345 tokens reached preflight target 352000 (model window 352000, model=glm-5.3-flashx); preflight compressing before forward
```

# 问题 1（主）：A 时代的压缩块对 B 完全失配

- 原始上下文已达 **613,345 tokens**，B 视角 pendingT1=**542,575**（几乎全裸）——A 时代的所有 block 没有作用到 B 的请求上；
- 佐证行：

```
2026-09-19T10:36:50.864Z [acp-proxy: ctx has 381 message(s), state has ? ref(s) mapped]
2026-09-19T10:36:50.857Z [acp-usage] round 1 input=296559 cached=1280 (cache hit 0%) fold=new
```

（`state has ? ref(s) mapped` 的 "?" 与 `fold=new` 都指向 fold 状态没有跨模型延续）

- 后果：preflight 兜底压缩 **6 个 range 耗时 233,869ms（~4 分钟）**，期间只能靠 keep-alive 吊住客户端（#568 的提前 SSE + keep-alive 生效了，但等待太久）：

```
2026-09-19T10:32:00.661Z preflight still running after 30000ms grace — committed early SSE headers + keep-alive to hold the client (#568)
2026-09-19T10:35:24.512Z preflight compressed 6 range(s), ~296065 tokens saved (613345 → 307040) in 233869ms; rebuilding payload
```

- 压缩完仍有 307,040（87%），继续 OVER-LIMIT 注入，直到手动驱动一次大压缩（m02403–m02761，省 ~266,069）才回落。

**期望**：同一 x-session-id 内切换模型，A 攒的 blocks 对 B 继续生效（首请求应 ~5-8 万，而不是 61 万全量 + 4 分钟卡顿）。

# 问题 2（搭车发现）：preflight 生成了 endId < startId 的反向区间并被接受

```
2026-09-19T10:33:22.139Z [acp-proxy: [Compressed m02243–m00334 → 1 block(s), ~88924 tokens saved.]]
2026-09-19T10:34:55.667Z [acp-proxy: [Compressed m02310–m00594 → 1 block(s), ~108476 tokens saved.]]
2026-09-19T10:35:24.402Z [acp-proxy: [Compressed m02362–m00834 → 1 block(s), ~99764 tokens saved.]]
```

三笔都来自本次 preflight 的 6 个 range。要么是区间校验没拦 wrap-around（m00334 < m02243），要么是 re-request 后 m 编号被重排导致区间语义错乱——无论哪种，区间合法性都应在前置校验拦下。

# 问题 3（搭车发现）：0.1.117 行格式 compress 被打回（回归）

工具描述写着 "Line form (preferred)"，但行格式调用稳定失败，同一内容换 object form 立即成功。今天出现 5 组以上的成对日志：

```
2026-09-19T09:10:14.103Z compress({"content": ["m01588–m01712 走bili UI move + ...", "## TASK ..."]})
2026-09-19T09:10:14.104Z   → compress result: [Compression FAILED: no valid ranges parsed (kind=no-valid-ranges, dropped=2)
2026-09-19T09:11:00.851Z compress({"content": [{"startId": "m01588", "endId": "m01722", ...}]})
2026-09-19T09:11:00.851Z   → compress result: [Compressed m01588–m01722 → 1 block(s), ~48229 tokens saved.]
```

（09:19/09:27/09:48 还有 3 组同样的"行格式失败 → 对象格式成功"）

# 排除项

- 上下文窗口学习机制（weak overflow / upstream context overflow / retracted stale）：当天日志零事件，本例与它无关（但 `resolveLearnedLimit` 的会话级全局回退在 0.1.117 仍在，已知你计划移除，顺带确认）。
- 新加的 maxContextLimit=45% 压力线在本事件中正常工作（OVER-LIMIT 注入 @87%/86%/87%），不是它引起的。

[bili-issue-logs.md](https://github.com/user-attachments/files/32414926/bili-issue-logs.md)

ranxianglei:
## What changed

Root cause of #1001: opencode **silently native-compacted its own history mid-session** after the model switch (567 msgs / 613,345 tok vs ~2200 raws in the A-era; giant `h_*` blobs are client compaction summaries; ref counter continued past the A-era high-water mark m02196 → same session state, not a reset). bili has no announcement channel for this on the chat/completions path, so the A-era blocks were orphaned — the kernel renders orphaned active blocks at the head of the rebuilt context (`prune`: earliest surviving effective id ?? 0) alongside the full new content → 613k to model B. The 234s preflight then had to regenerate summaries for content the client had already discarded; that cost is inherent, but the *handling* of the rewrite was wrong.

Three fixes, one test file:

1. **Unannounced-rewrite detection** (`src/session.ts`): `detectUnannouncedHistoryRewrite(session, knownRefsBefore, liveRawIds)` — timing-invariant ratio test: capture the known-ref snapshot BEFORE `processTurn` (post-turn every incoming id has a fresh ref and the ratio reads 1.0), then flag when ≥20 refs were known pre-turn, the session carries compression history, and <50% of incoming messages carry pre-turn refs. Append-only turns stay near 1.0; a rewrite collapses it. Wired into both `prepareAnthropic` and `prepareOpenai` (`src/server.ts`): detection marks the existing `compactionBoundary`, so the rewrite takes the SAME auditable path as an announced `/compact` — `applyCompactionArchive` archives the orphaned blocks (reason "content replaced by client native compaction summary") and prunes `byRaw`/`byRef` to live ids (stops the additive dead-ref leak and the theoretical ref re-issue window).
2. **Reversed range normalization** (`src/stream.ts` + `src/preflight.ts`): after a rewrite, ref numbers are no longer monotonic with position, so position-derived spans come back numerically reversed (the issue's m02243–m00334 et al.). The kernel resolves boundaries BY POSITION and swaps silently — now `normalizeRangeOrder()` normalizes up front in both preflight (with an explicit warn log) and the `applyRanges` funnel that guards every compress caller, so specs/logs/stored ranges stay honest.
3. **Observability bug**: `stream.ts` logged `state has ? ref(s) mapped` because it read `.size` on a plain object (`messageRefs.byRef` is `{}`-shaped, not a Map) — always printed "?". Now counts keys. (This line in the issue logs was NOT evidence of lost fold state; refs had continued correctly.)

## Issue 3 verdict (no code change)
The line-form "regression" is not one: the kernel parser is byte-identical between 0.0.73 and 0.0.74 (bili 0.1.116→0.1.117 only bumped the pin), and the failing calls split header+body into TWO array elements — each element is parsed as its own range, so the first fails "line entry: missing summary after the refs header line" and the second "no mNNNNN/bN refs in header" → `kind=no-valid-ranges dropped=2`, exactly matching the logs. Correct line form is ONE string per range: `m01588–m01712 topic\n\n<summary>`. A salvage-merge (treat `[headerOnly, body]` as one range) would be an acp-kernel change — cross-repo, manual per AGENTS.md scope note.

## Side question from the issue
`resolveLearnedLimit` did exist session-wide in 0.1.117 (`src/weak-overflow.ts:94`); it is already REMOVED on master by #987 (commit 721c4a5, weak-overflow.ts deleted) — nothing left to do here.

## Tests
New `tests/fix-1001-history-rewrite.test.ts` (5 tests):
- unit: detector positive (8/24 known → detected) + negatives (append-only, fresh session, no blocks);
- unit: `normalizeRangeOrder` (swaps reversed mNNNNN pairs, leaves bN/mixed untouched);
- e2e openai-wire through the real proxy: turn 1 grows past the window → preflight creates a real block → the client rewrites history (last 8 kept + 16 fresh) → asserts boundary marked+consumed, block archived with reason, `byRaw` pruned to exactly the live ids, no duplicate ref numbers, rebuilt payload carries the summary and fits, and an append-only turn 3 continues cleanly.

## Pre-flight (rebased head, origin/master 23df1ef)
- `npm run typecheck` ✓
- `npm test` ✓ 1970 pass / 0 fail / 2 skipped (gated e2e)
- `npm run build` ✓

Closes #1001


ranxianglei:

[bot] 🏷 **PR #1009 verified — code matches description, local pre-flight reproduces, CI fully green.**

This is my own PR ([PR]-mirror rule: verify, don't redo). Pulled `wt-1001-ework-daemon` @ `599f19f`, reviewed the full diff against current master (`23df1ef`), ran the pre-flight locally, and polled CI to completion.

**Diff hygiene** — single linear commit over current master, 5 files (+336/−5), every line on-topic: no unrelated files, no whitespace churn, no lock-file drift. Fresh base, no stale-base risk.

**Code review findings:**
- *Detector (`src/session.ts`)* — timing invariant holds: `knownRefsBefore` is snapshotted before `processTurn` in both `prepareAnthropic` and `prepareOpenai`, so the ratio can't read 1.0 post-turn. Gates (`≥20 known refs`, `blocks.length > 0`, `ratio < 0.5`) correctly exclude fresh sessions and first replays. False-positive edge analyzed: an append-only turn adding more new messages than the entire prior history could trip the `<0.5` ratio, but then nothing left the context → `applyCompactionArchive` archives zero blocks and the prune keeps every (live) ref, so worst case is one warn log + metadata bookkeeping. Harmless.
- *Archive+prune wiring* — `markCompactionBoundary` sets `pending: true`; the existing `applyCompactionArchive` consumes it, archives only active-before/inactive-now blocks with the `content replaced by client native compaction summary` reason, prunes `byRaw/byRef` to live ids, and leaves `nextIndex` untouched (confirmed in `session.ts:454-455` comment + code) → no ref-slot re-allocation onto retained tails. Consistent with the kernel id-never-reused contract.
- *Normalization (`src/stream.ts`)* — verified the kernel's `CompressRange` is a plain mutable `{startRef, endRef}` object, so the in-place swap in `normalizeRangeOrder` is safe. Regex is m-ref-only, so `bN`/mixed endpoints are untouched (correct — no cross-namespace ordering). Applied at both sites: the `applyRanges` funnel (guards every compress caller) and the preflight position-span loop with its own warn log.
- *Observability fix* — `.size` on the plain-object `messageRefs.byRef` was indeed always `undefined` → `"?"`; `Object.keys(...).length` is the right fix.
- *Both compression modes* — detection lives in the shared `prepare*` paths that run for every request regardless of `pluginMode`; the opencode-native-plugin case (the #1001 repro) goes through `prepareOpenai` + the preflight span loop, so both modes are covered.

**Local pre-flight (rebased head, reproduced):** typecheck ✓ · `npm test` **1970 pass / 0 fail / 2 skipped** (exactly matches the PR's claim) · build ✓. All 5 regression tests present as described in `tests/fix-1001-history-rewrite.test.ts` (detector positive/negatives, `normalizeRangeOrder` unit, openai-wire e2e through the real proxy), 34 assertions.

**CI:** all 8 check runs on `599f19f` completed **successful** — test matrix ubuntu/windows × Node 22/24, codex CODEX_HOME overlay, real codex vs deterministic fake upstream, build-artifact, version-guard.

No new problems found while verifying; the issue-3 verdict (not a regression; salvage merge would be an acp-kernel change) and the `resolveLearnedLimit` side answer stand as written in the PR body.

**结论：修复了客户端静默改写历史导致的孤儿 block 泄漏与非单调 ref 区间问题（边界处归档+剪枝、区间归一化），本地复跑与 CI 全绿，diff 干净，可以合并。** Merge stays yours: https://github.com/ranxianglei/billion-context/pull/1009

ranxianglei:
**Review + handoff note (2026-09-20)**

Reviewed the three fixes on behalf of #1001 — logic verified sound:

1. **Detection timing invariant** — `knownRefsBefore` is captured pre-`processTurn` in both `prepareAnthropic` (server.ts:1978) and `prepareOpenai` (server.ts:2140); post-turn the ratio would always read 1.0. Correct.
2. **Boundary wiring** — `markCompactionBoundary` → `applyCompactionArchive` consumes it with `activeBefore` captured at server.ts:1995/2147 (pre-turn), and `nextIndex` is left alone, so the id-never-reused contract (AGENTS.md §2) is respected. Correct.
3. **`normalizeRangeOrder`** inside the `applyRanges` funnel covers every compress caller (stream.ts:30, stream-openai.ts:25, stream-responses.ts:39, loop/core.ts:161, compress-loop-responses.ts:92, preflight.ts:970) + the preflight span emit. The `byRef` `.size` observability fix confirmed (plain object, always printed `?`).

Housekeeping done on this branch: rebased onto current master (was behind #1005/#1007/#1011 — server.ts hot-file overlap resolved, no semantic conflict). Full pre-flight green: typecheck ✅, 1972 pass / 0 fail, build ✅.

**Gap: the issue's 问题 3 (line-form compress rejected — `kind=no-valid-ranges, dropped=2`) is NOT covered by this PR.** Root cause is kernel-side and now fixed in acp-kernel #330: models emit the documented line form split across sibling array elements (`["m01588–m01712 topic", "## TASK …"]`); the header-only half fails "missing summary" and the summary half fails "no refs in header", so the whole call dies. #330 adds `coalesceLineEntries()` (refless string following a header-only string is adopted as its summary body). Once #330 merges and ships as 0.0.79, this branch will bump the pin so #1001 is fully closed end-to-end.

Scope note: `prepareResponses` is not wired for silent-rewrite detection — codex announces its compactions through its own channel (`reconcileNativeCompactionBoundary`, server.ts:2290), so this matches the issue's chat/completions scope; flag here if a Responses client is ever seen rewriting silently.

ranxianglei:
**Follow-up commit ab06840 — closes the remaining 问题3 (line-form compress regression):**

- **acp-kernel pin 0.0.78 → 0.0.79** (kernel #330): the documented line form was being split across sibling array elements (`["m01588–m01712 topic", "## TASK …"]`), each half invalid → whole call died `kind=no-valid-ranges dropped=2` while the object form of the same content succeeded. Kernel now coalesces a bare refs-header string with the summary-only string that follows it.
- **`src/stream.ts` hint**: the failure message now describes BOTH accepted content shapes (object `{startId,endId,summary}` and line-form string, incl. the split-element form) — previously object-only, which is exactly why the model flailed between formats.
- **New regression guard** in `tests/fix-1001-history-rewrite.test.ts`: split-element line form must produce a block through the bili `applyRanges` funnel (guards the pinned kernel actually carries the coalesce).

Pre-flight on the pushed head: typecheck ✅ · full suite 1975 tests, 0 fail (2 skipped, pre-existing) · build ✅. With this, all three problems in #1001 are covered: P1 unannounced-rewrite detection + boundary archive/prune, P2 reversed-range normalization, P3 line-form coalesce. `Fixes #1001`.

Issue #359, opened 2026-08-30
pi plugin /acp panel via ctx.ui.notify() — transient toast in web hosts (same class as billion-context-pi#255)


ranxianglei:
## Repro

1. Run pi through the proxy (baseURL → `/bili/...`), pi plugin installed (`bili plugin install`).
2. Run `/acp` in a web host (agegr/pi-web).
3. Panel appears as a transient toast — not in the session transcript.

## Root cause

`src/agent/pi.ts:260` — the `/acp` handler ends with:

```ts
const panel = typeof status.panel === "string" ? status.panel : undefined;
notify(panel ?? renderAcpStatus(status), "info");  // → ctx.ui.notify()
```

Same root cause as [billion-context-pi#255](https://github.com/ranxianglei/billion-context-pi/issues/255) (fixed there in PR #256): persistent transcript content routed through the transient notification channel. pi-web renders `notify()` as a toast by design and renders `custom_message` session entries persistently — so the fix direction is the same: `pi.sendMessage({ customType, content, display: true })`.

## ⚠️ Why the fix is NOT a straight port of billion-context-pi#256

The model-interference concern is stronger in proxy mode:

- **Extension mode** (billion-context-pi): ACP's context transform runs *inside pi* and rebuilds the sent view from session entries — it filters `acp-status` panels out before they reach the LLM (src/messages.ts, `ACP_STATUS_CUSTOM_TYPE`). Model never sees the panel.
- **Proxy mode**: pi natively sends the full conversation to the proxy. An injected `custom_message` becomes an ordinary **user message** in the request (pi's `convertToLlm` projects all custom messages as user messages). The proxy's acp-kernel pipeline then treats it as real conversation content — it sits in the recent zone, so it stays and is forwarded to the model. The proxy cannot tell a UI-only panel from a real user message.

A plugin-only port (just switching to `pi.sendMessage`) would leak every `/acp` panel into the model's context — exactly the interference the extension-mode fix explicitly prevents.

## Suggested fix (two parts)

1. **Plugin** (`src/agent/pi.ts`): emit `pi.sendMessage({ customType: "bili-acp-status", content: panel, display: true })`, fall back to `ctx.ui.notify()` when `sendMessage` is unavailable (older pi).
[...]

PR #1034, opened 2026-09-20
fix(#1029): self-contained compress failures, incoming marker strip, refs span hint

### LINKED ISSUE #1029: ACP 状态标记被拼接进用户消息文本块；跨会话代际 stale refs 压缩整批失败
# ACP 状态标记被拼接进用户消息文本块 + 跨会话代际的 stale refs 压缩失败

**版本**：billion-context 0.1.125（DSH Desktop v0.6.5-beta.1 接入，openai 协议，DSH 全量历史重发 + prefix-affinity 会话）
**影响**：ACP 的 ❌ 状态标记混入用户自己的消息文本，在客户端阅读流里看起来像用户/模型输出的一部分；会话代际更替后 compress 必然整批失败且提示不易自愈。

## 缺陷 1：状态标记拼接进用户消息文本

`[ACP] Compression FAILED …` / `[ACP] stream error …` 等标记被拼接进用户消息的同一个 text part。实测一轮用户消息渲染为：

```
继续|刚刚怎么出问题了？|❌ [ACP] Compression FAILED: no valid ranges parsed (kind=no-valid-ranges, dropped=2)
```

用户输入与代理标记在一个 text part 里，导致：①阅读流里模型与用户都无法区分哪些是用户原文；②模型曾据此推理"自己的压缩是否成功"。

建议：状态标记走独立 text part / 独立事件，或至少保证与用户文本分段。

## 缺陷 2：`Rejected entries:` 列表恒为空

`no valid ranges parsed (kind=no-valid-ranges, dropped=2). Rejected entries:` 后不跟随任何条目，模型无法自纠正（同一会话内观测到 identical 的 malformed 调用连续重试 5 次、每次 dropped=2）。

## 缺陷 3（本次实测）：会话代际更替后 stale refs 必然整批失败

DSH 侧重启/分叉后（prefix-affinity fork，新会话实例 refs 从 m00001 重排），模型上下文里残留上一代的 nudge/compress 引用。实测一轮：

```
compress FAILED m00001–m00518, m00520–m00576 → 0 blocks
session highest ref=m00263, unknown ranges in request=2/2, session history=0 compression(s), 0 block(s)
```

失败消息已正确指向 `acp_status`，但：①模型常直接重试旧 refs 而不先 acp_status；②伴随 `❌ [ACP] stream error: fetch failed` 时上下文里更容易积累多条失败标记。建议考虑：compress 拒绝时在 in-band 标记中内联一小段"当前可用 refs 摘要"（或在 nudge 注入时携带代际标识），减少一轮 acp_status 往返。

## 环境

DSH 客户端每步重发全量历史（宿主压缩已让位给代理折叠）；今晚宿主多次重启导致多代会话实例更替，属高频触发场景。

ranxianglei:
## Background

Issue #1029 reports three ACP status defects observed on DSH Desktop v0.6.5-beta.1 (proxy mode, OpenAI protocol, full-history resend every step, prefix-affinity sessions):

1. ❌ status markers mixed into user-visible text (observed concatenated into the user's own message text part);
2. `Rejected entries:` list always empty → the model cannot self-correct (5 identical malformed retries observed);
3. after a session-generation change (DSH restart/fork), stale refs from the previous generation fail every compress batch.

## Root causes (verified against master @ 3adc9ac, acp-kernel 0.0.80)

**Defect 2 — dangling header (bili bug).** The no-valid-ranges failure from `applyRanges` put each reason on its own line (` Rejected entries:\n- entry 0: …`), so line 1 ended with a bare header. The streamed visibility marker (`buildVisibilityMarker`) keeps **only line 1** of the result — so the ❌ marker stored by the client always ended with `Rejected entries:` and zero bullets, even though the tool result itself carried the reasons. Proxy-mode tool calls are ephemeral (never enter client history), so on the next turn the model's only memory of the failure was that truncated marker → blind identical retries. Additionally, the #847 repeat-failure guard only tracked *execution* failures; malformed-argument parse failures had no streak escalation at all.

**Defect 3 — stale refs across generations.** Kernel 0.0.80 already handles this correctly (gateMessage explains the generation scenario + diagnostics + points at `acp_status`). What was missing is bili-side convenience: the rejection carried no hint about which refs are currently valid, forcing a dedicated `acp_status` round-trip that models often skip.

**Defect 1 — marker accumulation (bili enabler; the literal concatenation is DSH-side).** In proxy mode, failed-compress markers stream to the client as plain assistant text (#717 design). DSH stores them in its own history and resends the full history every step, so stored markers re-entered the model context and reading flow on every turn, accumulating under retry storms. No incoming-side stripping existed anywhere (outgoing tag/marker stripping did; incoming history was forwarded verbatim). Note: the literal "marker inside the user's text part" concatenation happens on DSH's storage/render side (its source is not available here) — details requested in the issue thread; the bili-side enablers are fixed below.

## Changes

- `src/stream.ts`
  - the no-valid-ranges failure is now **self-contained on line 1**: up to 3 reasons (≤160 chars each) inlined as `Rejected entries: X | Y.` — both the tool result and the streamed marker carry the full actionable text;
  - the repeat-failure guard now also covers the **parse path** (keyed by kind+invalidItems+reasons), advising to fix argument shape instead of re-issuing;
  - zero-block rejections (including all-unknown-refs after a generation change) inline a compact snapshot: `[Current context: N visible message(s), refs mNNNNN–mMMMMM, K active block(s). Refs restart at m00001 after a session-generation change — request only refs inside this span, or call acp_status for exact compressible ranges.]`
- `src/acp-panel.ts` — new `stripAcpStatusMarkers()`: strips ACP status marker lines (emoji-prefixed `[ACP] …`, line-initial; plain `[ACP]` text untouched) from incoming user/assistant message content in string and text-part form. Mirrors the #359 panel-strip pattern: messages are never deleted (tool_calls pairing safety); emptied content degrades to `" "`.
- `src/server.ts` — applies the strip in all three prepare paths (Anthropic/OpenAI/Responses) right after the existing panel-strip, with an info log when lines are removed.
- `src/loop/core.ts` — suppresses duplicate identical status markers within one request (retry storm → one marker instead of N).
- Tests: new `tests/issue1029-acp-marker-stale-refs.test.ts` (5 regression tests reproducing each defect); updated `tests/compress-reject-reasons.test.ts` assertions to the inline format.

Both compression modes considered (§6): plugin mode (`handlePluginTool`) returns raw results and is unaffected by marker changes; it benefits from the self-contained failure text + refs snapshot. Incoming stripping applies to all requests and is a no-op where no markers exist. No config schema / persistence / cross-repo changes.

## Pre-flight (rebased head 9ed1e2b on master 3adc9ac)

- `npm run typecheck` ✅
- `npm test` ✅ 2029 pass / 0 fail / 2 skipped (e2e-gated)
- `npm run build` ✅
- E2E: zero-token preflight ✅ (codex binary present); **full run not possible in this environment** — the default local upstream (127.0.0.1:8199) is unreachable from the daemon sandbox (connection refused). The manual-dispatch CI e2e workflow covers the real-client path.

Closes #1029


ranxianglei:

[bot] 🏷 Reviewing this PR now — pulling the branch (`2026-09-20_issue1029_acp-marker-stale-refs`, head `9ed1e2b`), checking CI, diff cleanliness, and the overlap with open PR #1008 ("name the live compressible span in failed compress receipts"). Will report findings + verdict shortly.

ranxianglei:

[bot] 🏷 **Review of PR #1034 — verified, rebased onto current master, ready to merge** (merge itself stays with you: https://github.com/ranxianglei/billion-context/pull/1034)

**Verdict: all three fixes are correct and well-scoped.** No issues found in the code; two housekeeping items handled directly on the branch (below).

## Verification (on the rebased head, not the original base)

The PR was based on `3adc9ac`, but master moved to `af42145` (#867 Google-wire merge, which rewrites the same region of `src/loop/core.ts`). Per review discipline I rebased onto live master and re-ran everything:

- Rebase conflict in `src/loop/core.ts` (~L590–604): resolved as the union of intents — kept master's Google-wire `signature: call.signature` on the `proxyResults.push(...)` **and** this PR's `seenMarkers` dedup block. No semantic loss either way.
- On rebased head: `npm run typecheck` clean · `npm test` **2051 pass / 0 fail / 2 skipped** (e2e-gated) · `npm run build` green. New regression file alone: 5/5 pass.
- Pushed as 2 commits: `30f887b` (rebased fix) + `4de8233` (changelog entry — see below). PR now shows `mergeable: true` against `af42145`; CI on the new head is running (`e2e-fake-codex` and `PR Build Artifact` already green, main CI + Windows overlay in flight).

## Root causes confirmed

1. **Dangling "Rejected entries:" header** — real. `buildVisibilityMarker` (`src/compress-loop.ts`) keeps only line 1 of non-`acp_status` tool results, and the old no-valid-ranges failure put each reason on its own line → the client-stored ❌ marker carried zero reasons. Inlining ≤3 reasons (≤160 chars each) onto line 1 is the right fix, and I checked the kernel side: every `invalidReasons` string in acp-kernel is a static literal (no embedded user content), so the single-line invariant is safe — no newline can sneak into the inline list.
2. **Parse failures bypassing the #847 streak guard** — real. The guard was keyed off execution failures only; malformed-argument loops never escalated. The `parse:` namespaced key (`kind+invalidItems+first-3-reasons`) won't false-positive against range-spec keys, and the advice text correctly changes from "retry different ranges" to "fix argument shape".
3. **Stale refs across generations** — real, and `currentRefsSnapshot` is the better mechanism than hinting at specific dead refs: min–max span over `byRef` + visible-message count + active-block count, one compact line, always truthful by construction (it reads live state, so it can't go stale itself).
4. **Incoming marker strip** — placement is correct: after panel-strip/normalize, *before* core projection in all three prepare paths (`prepareAnthropic` ~L2003, `prepareOpenai` ~L2169, `prepareResponses` ~L2383). This is also ref-stability-positive: markers were injected into the outgoing stream, so the kernel never hashed them — stripping resends restores exactly the bytes the kernel originally saw, keeping raw-id/ref assignment stable. Leaving them in would churn ids instead. Scope is tight (emoji-prefixed `[ACP]` lines only; user/assistant roles; messages never deleted; emptied text degrades to `" "` so tool-call pairing survives). Both compression modes considered: proxy mode benefits from all three; plugin mode gets the self-contained receipt via the tool result itself, and the incoming strip is a no-op there (native agents' histories carry the tool call, not marker lines).
5. **Per-request marker dedup** — correct scope. I checked the second `visibilityMarkers !== false` site (`src/loop/core.ts:681`): that path pushes loop-*internal* system messages (text-protocol fallback), not client-streamed markers, so dedup belongs only at the stream site. Cross-request duplicates are intentionally left alone — each turn stores one copy, and the incoming strip removes them on the next resend.

## Housekeeping done on the branch

- **CHANGELOG entry added** (`4de8233`, separate docs commit per convention) under `[Unreleased] → Fixes`. The PR omitted it; every merged fix here carries an entry.
- **Overlap with open PR #1008** ("name the live compressible span in failed compress receipts"): its unique remaining content is `compressibleSpanHint(state)` in `src/stream.ts` — same intent as fix ③ here, different mechanism. #1008 is stale (base predates several merges; most of its commits already landed as #743/#739). Recommend **closing #1008 once this merges** — `currentRefsSnapshot` covers its intent more completely (span + counts + generation explanation vs. span only). Owner decides; flagging rather than closing it myself since it has a distinct mechanism you may prefer.

## Auto-merge gate assessment

Touches wire-visible message shape (marker text, incoming content mutation) and spans the compress-loop hot file → **human-merge territory** per §7.4, which is why this lands via your merge button. Nothing load-bearing (no `update.ts`, no persistence format, no kernel pin change).

---

**中文摘要**:修复了 compress 失败回执不自包含(标记只剩一行、拒绝原因丢失导致模型盲目重试)、参数解析失败不触发重复失败升级、会话代际变更后陈旧 m-ref 无可用范围提示三个问题(均已对照内核源码核实根因);已在当前 master 上完成 rebase(Google-wire 冲突按双方意图合并)、补齐 CHANGELOG、本地 typecheck/2051 测试/build 全绿并推回分支,可以合并。

ranxianglei:

[bot] 🏷 **CI follow-up — all four workflows green on the rebased head `4de8233`:**

- `CI` (typecheck + full test suite + build): ✅ success
- `e2e-fake-codex`: ✅ success
- `PR Build Artifact`: ✅ success
- `CI Windows codex overlay (#686)`: ✅ success

PR is `mergeable: true` against current master (`af42145`). Everything verified — ready for your merge whenever: https://github.com/ranxianglei/billion-context/pull/1034

(And the standing suggestion from my review above: close stale PR #1008 once this lands — its span-hint intent is covered by `currentRefsSnapshot` here.)

PR #1053, opened 2026-09-20
fix: omit proxy message ids when replaying Responses input

### LINKED ISSUE #1052: Responses replay rejects a short proxy-generated assistant message ID with invalid_value
## Observed behavior

With bili 0.1.128, a streamed Responses request was rejected upstream with HTTP 400:

```text
Invalid 'input[100].id': 'msg-proxy-2-7d1e132ded554149'. Expected an ID that contains letters, numbers, underscores, or dashes, but this value contained additional characters.
```

The ID is below 64 characters and its visible characters are ASCII letters, numbers, and hyphens. This does not establish that hyphens are invalid or that hidden characters are present. The complete failing payload and a successful live-upstream retry are not available, so the exact upstream cause remains unconfirmed.

## Relevant code path

At master `615a496`, `createResponsesAdapter` emits local `msg-proxy-` message IDs for round-2+ remapping and `emitText()`. Clients can persist these output messages and replay them in later requests. `prepareResponses` normalizes message types and calls `sanitizeResponsesInputIds`, but the sanitizer only rewrites IDs longer than 64 characters; the short ID above survives.

This differs from the over-length case in #242 / #244 and native compaction conversion in #769 / #770.

## Proposed compatibility fix

For a full `type: message`, `role: assistant` input item with string/array content and an ID starting with `msg-proxy-`, omit only its local ID before replay. Preserve all other fields, client-visible stream lifecycle IDs, and existing long-ID cleanup for other items.

Reference-only items and `previous_response_id` behavior are out of scope. Real-upstream acceptance of the resulting message shape, including retained status/phase/output content, still needs validation; this is not a confirmed end-to-end fix.

Candidate commit: https://github.com/liangxu-one/billion-context/commit/8b789db

## Validation

- 7 production lines and 99 test lines added.
- Three new regression tests fail before the patch because the ID is retained.
- After the patch, all 7 focused tests pass serially, covering both ID generation paths, normalization, field preservation, non-target items and existing long-ID behavior.
- Typecheck and build pass.
- Full-suite runs on both baseline and patched code timed out after 120 seconds. Baseline already showed restart/session-routing failures; one parallel focused run also hung/cancelled, while the subsequent serial focused run passed. The full suite is NOT confirmed green.
- Real Codex E2E preflight fails with `spawn codex ENOENT`. Live upstream behavior has not been verified.

A draft PR will link this issue.

liangxu-one:
## Summary

Fixes #1052.

Omit locally generated `msg-proxy-` IDs when replaying self-contained assistant messages through Responses input sanitization. Keep the IDs in streamed output so SSE item/event correlation is unchanged.

The reported short ID contains only visibly valid ASCII characters. This patch avoids replaying a synthetic identity; it does not establish the exact upstream cause of the reported `invalid_value` error.

## Changes

- Remove `id` only for `type: "message"`, `role: "assistant"`, a `msg-proxy-` prefix, and string/array `content`.
- Cover both round-2+ remapped IDs and `emitText()` timestamp IDs.
- Preserve message content and other fields, native IDs, non-message items, reference-only items, and the existing overlong-ID handling.
- Add replay regressions and extend the round-2 lifecycle test to verify that replay sanitization does not mutate the original streamed message.

Three files, 7 production lines and 99 test lines added. No dependency or version changes.

## Validation

- New regression file against the original implementation: 3 failed, 2 passed.
- Focused tests, run serially: 7 passed, 0 failed, 0 skipped.
  `node --import tsx --test --test-concurrency=1 tests/responses-proxy-id-replay.test.ts tests/responses-round2-lifecycle.test.ts`
- `npm run typecheck`: passed.
- `npm run build`: passed.
- `git diff --check`: passed before commit.
- Full baseline and patched suites both exceeded a 120-second limit, with failures/cancellations; neither is a passing full-suite result. One parallel focused run also stalled; the serial run above passed.
- Codex E2E preflight failed with `spawn codex ENOENT`; the real E2E was skipped.

## Draft / remaining verification

Kept as a draft because live-upstream replay and the required Codex E2E are not verified. The exact failing upstream payload was not available. Tests establish local sanitizer and SSE lifecycle behavior, not upstream acceptance of every preserved status/phase/content combination. Reference-only replay and retained `previous_response_id` (`ACP_KEEP_RESPONSE_ID=1`) behavior are not validated by this patch.

This is distinct from the earlier overlong-ID fix (#242/#244) and compaction replay fix (#769/#770).


ranxianglei:

[bot] 🏷 收到,开始审查这个 PR。我会:拉取分支核对代码 → 检查 CI 状态 → 本地跑 typecheck / 测试 / build,有结论后在此回复。

ranxianglei:

[bot] 🏷 Review complete — I pulled branch `2026-09-20_responses-proxy-id-replay` (head `8b789db`, base current master, mergeable/clean), reviewed the diff line-by-line, and re-ran the full pre-flight locally.

**Local verification (on the PR head)**
- `npm run typecheck`: passed
- Focused tests serial (`responses-proxy-id-replay` + `responses-round2-lifecycle`): 7 passed, 0 failed
- `npm run build`: passed
- Full suite: **2082 passed, 0 failed, 2 skipped** (~111s) — the skips are the default-gated Codex E2E. This clears the "full suite not confirmed green" caveat from the PR body in this environment.

**CI on head `8b789db`**: all checks green — test matrix (ubuntu/windows × 22/24), build-artifact, version-guard, and notably the **"real codex vs deterministic fake upstream (zero tokens)" E2E that could not run locally** (`spawn codex ENOENT`).

**Code review findings**
- The fix is surgical: 7 lines in `sanitizeResponsesInputIds` (`src/loop/adapter-responses.ts:60-66`), running input-side only at `src/server.ts:2650` (after `normalizeResponsesMessageItems`, before whitespace-drop). Streamed output is untouched, so SSE item/event correlation keeps its IDs — exactly as the PR claims.
- Predicate is tight and well-tested: `type:"message"` + `role:"assistant"` + string `id` with `msg-proxy-` prefix + string/array content. Native IDs, non-assistant roles, reference-only items (`item_reference`, `reasoning`, `compaction`, `function_call`, `custom_tool_call`), incomplete messages, and numeric IDs all pass through unchanged; cleanup is idempotent.
- Both generation paths are covered: the round-2+ remapped ID (`src/loop/adapter-responses.ts:399`) and the `emitText()` timestamp ID (`:536`) — the new lifecycle test asserts both shapes and that the original streamed message object is not mutated.
- Kernel compatibility (the main thing I checked beyond the diff): `acp-kernel`'s `responsesToCore` derives assistant-message identity from content hashes (`deriveMessageId(role, "text", displayText)`), **not** `item.id`; the two places that do fall back to `item.id` (reasoning with empty text, search/call items) are different item types this patch never touches. So dropping the replay ID shifts no refs, blocks, or compression state — in either compression mode (plugin or proxy), since the sanitizer is mode-agnostic wire hygiene.
- Diff cleanliness: 3 files, +106/−0, single linear commit, every line on-purpose, no version/dependency churn (version-guard agrees). No direct-fixable defects found.

**Residual risks (concur with the author's framing)**
- The exact upstream cause of the #1052 400 is still unconfirmed (no failing payload); this patch avoids replaying a synthetic identity rather than proving a root cause.
- Live-provider acceptance of the resulting ID-less shape (incl. retained `status`/`phase`/content combos) is not validated — the fake-upstream E2E proves wire mechanics, not provider validation policy. Reference-only replay and `previous_response_id` retention remain out of scope.
- Worst case for an upstream that *did* accept the synthetic ID before: losing it only costs prefix-cache continuity — nothing downstream references those IDs.

**Verdict**: mergeable on code and test grounds; keeping it draft until the real failing upstream is observed again is reasonable. Merging stays human-only.

中文摘要：修复了 Responses 重放时把本地合成的 `msg-proxy-*` 助手消息 ID 回传上游导致 400 `invalid_value` 的问题（#1052）——改动只在入站清洗阶段删除合成 ID、保留流式输出中的 ID，本地全量测试与 CI（含 Codex E2E）均绿，代码层面可以合并；是否解除 draft 等真实上游复现验证，由 owner 拍板。

PR #1055, opened 2026-09-20
feat: adopt parent compression blocks on anonymous fork detection (#629)

### LINKED ISSUE #629: Design: tree-shaped prefix affinity — forked branches inherit the shared prefix's compression state
## Background

Anonymous prefix affinity (#309/#316/#499, `src/prefix-affinity.ts`) resolves header-less clients by longest-prefix match over an append-only hash chain (`h_i = sha256(h_{i-1} || msg_i)`). It deliberately mirrors vLLM/SGLang radix prefix caching at the *identity* layer.

Current fork semantics (documented in the file header): when a request's history diverges from every stored chain, it **forks into a brand-new session with zero compression state**. The discarded parent candidates are recorded in `lineage` (`parents[]`, `reason: "forked"|"truncated"`) — but lineage is **UI/debug only, never used for matching or state inheritance** (`MIN_FORK_PREFIX = 3`, comment says so explicitly).

## The gap

A fork (user edits history, rewinds, or two parallel conversations share a common opening) throws away every compression block anchored inside the shared prefix:

- the parent session may have folded 100K tokens of that prefix into blocks;
- the forked branch re-sends the same bytes raw and must re-compress from scratch (nudge starts at ~0% accumulated);
- after compression, the branch's blocks duplicate the parent's coverage over identical content.

This is the classic radix-tree property that prefix-affinity stops short of: identity is content-addressed, but **compression state is not** — it lives per session (`kernel StateStore`, blocks owned by one session id).

## Proposal (discussion)

Make the fold state effectively tree-shaped so a fork **inherits the folds anchored inside its shared prefix**:

1. On fork detection (existing lineage machinery), copy/adopt the parent session's blocks whose `effectiveMessageIds` fall entirely inside the shared prefix (items `0..lcp-1`) into the new session, remapping anchors.
2. Keep per-session ownership for persistence — this is a one-time adoption at fork time, not shared live state (avoids two branches mutating one block).
3. Reuse the existing per-item hash chain as the prefix-addressing mechanism (`MAX_STORED_ITEMS = 128` currently caps it — may need raising or a separate block-anchor index keyed by item hash).
4. Safety unchanged: matching requires byte-identical history possession (#286/#309 argument).

## Open questions

- Cost/benefit: how often do real clients fork (edited history / rewind)? omp rewind, codex `/compact`-adjacent flows, web UIs with regenerate.
- Interaction with `restoreOutputBudget`-style per-session metadata (must NOT inherit).
- Kernel support needed: block adoption/remap API (current kernel owns blocks per session state).

Spawned from the #619 discussion — orthogonal to its remaining decode-fail scope (PR #622).

ranxianglei:
Fixes #629

## What

Anonymous (prefix-affinity) clients — bare `pi`/`opencode` rewind, `dsh` web regenerate, headless shells — have no stable session identity. When such a client forks its history mid-conversation (edit-and-resend / regenerate an earlier turn), the fork minted a brand-new session that started at **zero** compression state: every block the parent had already folded was lost, even blocks whose entire span still lies intact inside the forked request.

This PR adds opt-in **fork block adoption**:

- At the fork detection point in `src/prefix-affinity.ts` (where a lineage `{ reason: "forked" }` is minted), the proxy inspects each parent block.
- A block is adoptable iff its `effectiveMessageIds` are **fully present** in the forked request (no straddling the edit point). Blocks spanning the divergence are left behind — adopting them would misattribute.
- Adopted blocks are **copied** into the child (blocks + `blockContents` summary + seeded `messageRefs.byRaw/byRef` for the block's ids); the parent is never mutated (copy-on-fork).
- The kernel's **id-never-reused** contract is preserved: ids are never renumbered, numbers keep their kernel-assigned meaning; seeding refs only pre-registers the same ids the child's own processing would register.
- Gated behind **`forkAdoption` (default off)** — config flag `forkAdoption: true`, zero behavior change unless enabled.

## Why opt-in

Adoption cross-links two session states, which touches the message-ref/id neighborhood (#629 discussion). Default-off lets it ship dark and be flipped per deployment after soak.

## Files

- `src/fork-adoption.ts` — adoption core (presence check, copy, ref seeding)
- `src/server.ts` — hook at anonymous-fork mint (gated on `forkAdoption`)
- `src/config.ts` — `forkAdoption` option
- `tests/fork-adoption.test.ts` — end-to-end: mock upstream folds a block mid-growth (proxy mode), client forks (middle-message edit), asserts child carries the parent's **block id**, summary, seeded refs; parent untouched; fork keeps reconciling on follow-ups. Second scenario proves `forkAdoption: false` leaves the fork at zero compression state.

## Pre-flight (rebased on master 615a496)

- `npm run typecheck` ✅
- `npm test` ✅ 2079 pass / 0 fail
- `npm run build` ✅

Per AGENTS.md §7.4 this is a `feat:` touching persistence/session seeding — **human merge**, not auto-merge eligible.


ranxianglei:

[bot] 🏷 Reviewing PR #1055 now (branch `2026-09-20_fork-block-adoption` @ `19c433c9`, base master `615a496`). GitHub-side state so far: open, not draft, mergeable clean, all four CI runs green (CI / Build Artifact / Windows codex overlay / e2e-fake-codex).

Plan per AGENTS.md §7: pull the branch, re-run typecheck + full test suite + build against current master myself, audit the diff line-by-line (both compression modes, id-never-reused contract, diff cleanliness), then report findings here with a merge recommendation. Will reply again with the full review.

ranxianglei:

Full review done — I re-ran everything locally on the branch head and walked the mechanism end-to-end. **Verdict: mergeable** (human merge per §7.4, as you noted).

**Verified against the code, not just the description:**

1. **Hook gating (`src/server.ts:1406`).** First-request-only holds: `session.stats.requests` is incremented inside `prepare*` (server.ts:2099 et al.), strictly after the hook. `reason === "forked"` always carries exactly one parent (`src/prefix-affinity.ts:268` mints `{ parents: [forkParent.sessionId], reason: "forked", sharedPrefix }`), so `parents[0]!` is safe; multi-parent `"truncated"` lineages are excluded by the condition itself.
2. **Id fidelity.** `incomingCoreIds` runs `structuredClone` → `stripAcpPanelMessages` → `stripAcpStatusMarkers` → `anthropicToCore`/`openaiToCore` — byte-for-byte the same pipeline `prepare*` uses (anthropic server.ts:2117–2127, openai server.ts:2284–2298), and nothing else transforms `parsed.messages` between JSON.parse and ToCore on those paths. The adoptability predicate therefore judges on exactly the ids `processTurn` will see.
3. **Tier closure.** Confirmed in the pinned kernel (acp-kernel 0.0.80, `dist/index.js` ~L2936): block construction unions each consumed child's `effectiveMessageIds` into the new block's set. So any *active* block's `effectiveMessageIds` transitively covers its whole descendant closure — the active-only predicate plus the `directBlockIds` walk copies exactly the right records, and children ride along as inactive (consumed-by-parent) as designed.
4. **Copy-on-fork safety.** `getStore().loadSync` builds a detached `Session` from the envelope without inserting it into the live pool (only side effect: the #408 negative-token clamp save) — no double-writer. `applyForkAdoption` deep-clones every block/content, seeds `byRaw/byRef/tokenSnapshot` with not-already-present guards, and bumps `nextBlockId`/`nextRunId` to ≥ parent's — verified the kernel's `allocateBlockId`/`allocateRunId` consume exactly those fields, so no `bNNN`/`rNNN` collision between adopted and freshly allocated ids.
5. **id-never-reused.** Seeding only pre-registers raw→ref pairs the child's own `assignRefs` pass would otherwise re-issue fresh; since the cursor is `highestUsedIndex(map)+1`, fresh assignments land above the parent's ref space. No renumbering anywhere.
6. **Both compression modes (§6).** Anonymous prefix-affinity clients are proxy-mode; the predicate's precondition (source content fully present in the wire) is precisely what makes folding valid regardless of carrier. Plugin-mode clients bind stable identities, so an anonymous fork there is a corner case — and even then adoption only pre-registers state the agent's own history already implies.
7. **Known limitation (accepting, not blocking):** `stripImages` shifts content-hash ids once an image message ages out of the recent-N window ("self-healing downstream via orphan-GC" per the kernel). The hook computes ids from the unstripped body; if a folded image-bearing message's id shifts between the parent's fold and the fork request, worst case is a stale adopted block that the existing orphan-GC machinery cleans up — bounded, and no misattribution possible while ids are never reused. Narrow corner (needs stripImages ON + image msg inside a folded block + window-boundary crossing); fine for a default-off v1. Worth a follow-up if anyone actually runs that combo.

**Small fixes I made directly on the branch** (two commits, pushed `19c433c..0da7249`, gate re-run green: typecheck ✅, 2079 pass / 0 fail ✅, build ✅):

- `test:` seeded-ref assertions were missing — the e2e fixture can't catch a broken seed, because with the edit sitting *after* the folded span, order-based ref assignment coincidentally reproduces the parent's numbering. Added explicit asserts: for every adopted block, `child.messageRefs.byRaw[raw] === parent.messageRefs.byRaw[raw]`, plus `nextBlockId`/`nextRunId` ≥ parent's cursors. This pins the contract for the cases where alignment breaks (e.g. a parent with prior archives shifting numbering).
- `docs:` `CONFIGURATION.md` + `CONFIGURATION.zh-CN.md` now document `BILI_FORK_ADOPTION` / `forkAdoption` (env-var mirroring per §7.3 was missing), and I corrected the `FileConfig.forkAdoption` comment in `src/config.ts` — it claimed "the file's explicit false cannot be re-enabled by env", but the code (`config.ts:622`) has env winning both directions, matching the sibling `subagentSplit`/`maskHosts` convention. If asymmetric precedence (file-false as hard off) was the actual intent, that's a one-line code change — say the word and I'll flip it instead.

---

中文摘要：审查确认 fork 块继承的核心机制（完整存在性判定、层级闭包传递性、copy-on-fork 无写者冲突、ref 播种不违反 id 不复用契约、双压缩模式均安全）全部成立，仅识别出一个 stripImages 交互的已知窄角限制（最坏情况由现有 orphan-GC 兜底，不构成合并阻塞）；已在 PR 分支直接补上两个小修 —— 测试中钉死"播种的 ref 必须与父会话映射一致"（原 fixture 会掩盖 ref 播种静默失效）+ 补齐中英文配置文档并修正 config.ts 一条与代码矛盾的注释，修复后 typecheck/test(2079)/build 全绿，**可以合并**。

PR #1065, opened 2026-09-20
chore: bump acp-kernel 0.0.80 -> 0.0.81

ranxianglei:
Keeps billion-context aligned with the kernel release that fixes the compress-gate rework loop (kernel#327 → 0.0.81):

- gate error now names the covering block and gives actionable tier-distillation advice instead of an unreachable block-ID retry hint
- protected-exclusion warning tells the model not to re-target excluded refs
- plugin-mode agents behind bili (dsh etc.) get the same actionable diagnostics pi users got in billion-context-pi#482

0.0.81 verified live on npm before opening. Gates on the branch: typecheck 0 errors; 2129 tests / 2127 pass / 0 fail (2 env-skips); build OK.


ranxianglei (commented):
Independent verification on the branch head (39979c8): exact pin `0.0.81` in package.json + refreshed lockfile (integrity matches npm). `npm ci` clean, node_modules resolves 0.0.81, typecheck 0 errors, full suite 2129/2127 pass / 0 fail (2 env-skips), build OK. Cross-repo sequence respected (0.0.81 verified live before this PR). Diff is version-only — no content changes. Good to merge once CI finishes.

Issue #1066, opened 2026-09-20
bug: non-intercepted Codex compaction triggers are forwarded twice


tarnish233:
A Codex request ending in `compaction_trigger` reaches the upstream twice when `BILI_CODEX_COMPACT=pass`, or when interception rejects the request before preparation (for example, no active compression block).

Reproduction: send a streaming `/v1/responses` request with a Codex user-agent and a final `compaction_trigger` item to a counting mock upstream. Consume the client response, then keep observing upstream requests for one second. The count becomes two; the second request uses the raw client body.

Root cause: the trigger path forwards a normalized body with `prepared: null`; the passthrough tail checks `!prepared` and forwards again. This consumes extra upstream capacity and can replay unnormalized input after the client response has ended.

The existing server-side fix tracks whether forwarding has already happened independently of preparation, with regression coverage for both pass mode and an unmet intercept gate. A scoped PR will follow.

tarnish233:
修复 PR：https://github.com/ranxianglei/billion-context/pull/1068

ranxianglei:
在 master 独立复核,机制属实:

`src/server.ts:1742` trigger 未拦截路径 `return { body: forwardBody, prepared: null }`。调用方 `if (pendingForward) { await forward(...pendingForward.body, null...) }` 完成**第一次**转发;闭包从未给外层 `prepared` 赋值,try/finally 结束后控制流**落穿**到 `src/server.ts:1830` 的 `if (!prepared)` 尾巴 → 用原始 `bodyBuffer` 再发**第二次**(还绕过了 replaceBiliCompactionItems 的归一化,unnormalized replay)。

影响:codex 触发原生压缩时,pass 模式或 intercept gate 未满足 → 上游收到两个请求,第二个是未归一化的 raw body;客户端响应已结束后仍在上游重放。

修复方向(转发与否与 prepared 解耦,单独记账)正确。等 PR。

ranxianglei:
> 🤖 Powered by [ework](https://github.com/ranxianglei/ework) · qwen3.8-27b

Duplicate screening first: no open or closed issue/PR reports this double-forward — this is the first report of it.

**Root cause verified in code** (matches your analysis exactly):

- The non-intercepted trigger branch returns `{ body: forwardBody, prepared: null }` (`src/server.ts:1719`).
- That body is forwarded once at `src/server.ts:1793`.
[...]

PR #1068, opened 2026-09-20
fix: forward non-intercepted Codex compaction requests once

tarnish233:
Fixes #1066

A Codex `compaction_trigger` request that bypasses interception is forwarded with `prepared: null`, then the passthrough tail forwards the raw body again. Track forwarding separately from preparation so the upstream receives the request exactly once.

Includes regression cases for pass mode and an unmet intercept gate, observing upstream requests after the client response ends. The server forwarding guard is shared across plugin/proxy modes and does not change either summary carrier.

Validation: verified on the server by the author. Additional checks on current master (`33cff2b`) passed: typecheck, full suite (2,129 passed, 2 skipped), and build with real dependencies.

Extracted from the existing server-side commit `19e56d1`, preserving its code and tests. The independent tag-filter fix is submitted separately for #1067.

PR #1069, opened 2026-09-20
fix: recognize wrapped tag openings before shared-chunk closes

### LINKED ISSUE #1067: bug: streaming wrapped tag imitation leaks when its close shares the chunk
The streaming prose tag filter leaks a wrapped-turn imitation when its malformed opening and loose closing tag arrive in the same chunk. This is a remaining match-precedence case after #872, not a change to tool-call argument filtering.

Reproduction: push a wrapped ACP opening with an unterminated attribute, followed by imitation markup and a loose close, through `createTagEchoFilter()` in one chunk. Compare its output with `stripAcpTags()` for the same string: streaming emits the imitation, while batch filtering preserves only the surrounding prose.

Root cause: `BROKEN_ATTRS` is checked only when no ordinary tag match exists anywhere in the buffer. A later close therefore prevents recognition of the earlier malformed opening.

The existing fix compares match positions and lets the earlier malformed opening own its span. A regression test checks streaming/batch parity. The filter continues to apply only to model prose; actual tool-call arguments remain byte-exact per #1039. A scoped PR will follow.

tarnish233:
Fixes #1067

When a malformed wrapped-tag opening and its loose close share a stream chunk, the later ordinary tag match suppresses the `BROKEN_ATTRS` check and the imitation leaks into visible prose. Evaluate that malformed opening alongside ordinary matches and swallow its span when it starts first, matching `stripAcpTags()` behavior.

Includes a regression case for a shared-chunk close that checks streaming/batch parity and removal of imitation markup. This only changes the existing prose filter; tool-call arguments remain byte-exact (#1039), and plugin/proxy summary carriers are unchanged.

Validation: verified on the server by the author.

Extracted unchanged from the existing server-side commit `19e56d1` onto current master (`33cff2b`). The independent Codex double-forward fix is submitted separately for #1066.


ranxianglei (commented):
独立验证通过,逻辑正确:

- 把 `BROKEN_ATTRS` 提到候选环之后、`!m` 门之前,用 `broken.index < m.index` 择先——正好对齐批式 `wrappedSpan()` 的位置语义。同位不可能打平:BROKEN 的属性类 `[^<>]` 跨不过 `>`,而 LONE_OPEN/PAIRED 需要 `>` 收尾,矛盾。
- master 上复现:仅打上新测试跑 `tests/tag-echo.test.ts` → 48 pass / 1 fail(fail 即新测试);分支上 2128/2128 全绿,typecheck/build ✓。

**一个必改项(阻塞)**:本 PR 和 #1068 都捎带了 `acp-kernel` 0.0.80→0.0.81 的钉版 bump(package.json + lock)。这是与本修复无关的 scope 外变更,两边同改必然互相冲突,而且跨仓钉版有既定顺序——kernel #339(0.0.82)合并上架后,应由一个独立的 chore PR 直接钉到 0.0.82。请把这两个文件从本 PR 撤掉,保持 fix 纯净。

ranxianglei:

Reviewed PR #1069 independently against current master (`33cff2b`, includes #1065 kernel 0.0.81) — verified, mergeable.

**Diff hygiene.** Single commit `6a39fd6` on top of master, 2 files, +39/−19, every line on-purpose. No whitespace churn, no lockfile/version drift, no unrelated files. Base is the live master head; `mergeable_state: clean`; no other open PR touches `src/loop/tag-echo-filter.ts` or adjacent hot paths.

**Code** (`src/loop/tag-echo-filter.ts:308-340`). `BROKEN_ATTRS.exec(buf)` now runs every loop pass alongside the PAIRED/LONE_OPEN/LONE_CLOSE candidates and wins when `m === null || broken.index < m.index`. The winning branch is byte-identical to the old `!m` branch (same swallow state machine: `IMITATION_SWALLOW_CAP`, `swallowReleases=false`), so behavior outside the race case is unchanged. Two edge cases checked out:
- Tie (`broken.index === m.index`) is unreachable: a properly terminated opening can never match BROKEN_ATTRS at its own index because the attrs class `[^<>]{0,512}` cannot cross the terminating `>` (the comment says exactly this).
- When an ordinary match starts first, normal handling runs and advances the buffer; the broken opening is re-evaluated on the next pass. PARTIAL_TAIL ordering relative to BROKEN_ATTRS is preserved.

This matches the batch path's position-first semantics in `wrappedSpan()`, closing the streaming/batch divergence identified in #1067.

**Regression test** (`tests/tag-echo.test.ts:651-663`). Single-chunk push of `bad ${wrappedTurnEcho()}<\/acp> ${prose} tail`, asserting streaming output equals `stripAcpTags(full)` and no tool-call XML fragment survives. I verified it is a true repro: swapped in master's filter file and the new test fails (`not ok 1`); on the PR branch it passes.

**Validation (rerun here, not taken on faith).** On the PR head: typecheck clean; full suite 2130 tests → 2128 pass / 0 fail / 2 skipped (e2e gate); tsup build success. GitHub CI on `6a39fd6`: 8/8 green (ubuntu 22+24, windows 22+24, build-artifact, version-guard, both codex e2e jobs). The earlier caveat about validation is moot given the rerun.

**Scope/invariants.** Prose filter only — the #1039 byte-exact tool-argument invariant block is untouched, no wire/message-shape change beyond intended stripping, no config schema or persistence change, no acp-kernel pin change. Both compression modes unaffected (the filter applies to model prose fields in both).

Verdict: **mergeable**. Merging stays human-only per repo rules — please merge when ready: https://github.com/ranxianglei/billion-context/pull/1069

中文摘要:复核了 #1069——把流式过滤器里只在"无普通标签命中"时才检查的 BROKEN_ATTRS 提进按位置择先的竞争(与批式 wrappedSpan 对齐),修复了畸形 wrapped-tag 开头与 loose close 同 chunk 时仿冒标记泄漏到正文的 bug(#1067);回归测试已验证在 master 上失败、在 PR 分支通过,typecheck/全量测试/build/CI 全部绿,diff 干净且只动 prose 过滤路径,可以合并。

ranxianglei:
已在 `11c02df`/`d7fce13` 直接撤掉钉版 bump(两个文件恢复为 master 状态),fix 内容零改动。清理后:typecheck 0 错,目标测试文件全绿(tag-echo 49/49,codex-compact 14/14)。阻塞项解除,可合并。

Issue #1078, opened 2026-09-21
[Question] The issue of the order in which multiple bili proxies take effect.


BsoBird:
The user configured a bili proxy for their own workload. The actual request address that the bili proxy forwards to is in fact another bili proxy, which uses config B — but config B is not suitable for the user's workload. So we should make the bili proxy closest to the user take effect globally. It should carry some request parameters, and when the downstream bili proxy detects the config, it should just forward the request directly without processing it.

BsoBird:
example:  user-codex -> local-bili(or not local,but suitable user's workload) -> third part proxy -> bili-proxy-B -> real-api-endpoint

in this case, bili-proxy-B should not work

ranxianglei:
> 🤖 Powered by [ework](https://github.com/ranxianglei/ework) · qwen3.8-27b

The behavior you want already exists — the bili closest to the user governs, and downstream bilis stand down. Here's the mechanism, why your specific chain may not show it, and what I'd propose.

## How it works today (since v0.1.61, 2026-08-28 — #300/#292)

- Every bili instance has a per-process identity. When it forwards a request **it processed**, it stamps `x-bili-hop: <instance-id>` on the outbound request (`src/server.ts:3216`, stamping happens only when the instance actually ran the pipeline).
- A bili that **receives** a request already carrying that header knows an upstream bili already ran the compression pipeline. It skips **all** processing — no session state, no tool/tag injection, no preflight, no nudge — and forwards the request verbatim, logging a prominent warning:
  `[chain] inbound request carries x-bili-hop from another bili instance (…) — bili→bili chain detected. Passing through without processing to avoid double compression` (`src/server.ts:963-969`; pipeline gate at `src/server.ts:1179`).
- The marker keeps propagating down the chain (a passthrough hop never re-stamps it), so longer chains degrade correctly. Covered end-to-end by `tests/e2e-bili-chain.test.ts`: the body reaches the final LLM **byte-identical**, exactly one processing layer, exactly one `[chain]` warn naming the upstream instance's id.

So in your topology `codex → bili-A(config A) → third-party relay → bili-B → real API`, B *should* already stand down and A's config governs globally.

## Why you may still see B's config take effect

[...]

PR #1079, opened 2026-09-21
feat: bili→bili chain integrity — x-bili-sig digest + ACP-artifact fallback

ranxianglei:
Related to #1078 (first of a two-part fix; the metadata-carrier variant is discussed on the issue after this merges).

## Problem

#300's `x-bili-hop` marker stops double processing when bili A forwards directly into bili B — but a third-party relay between them breaks it:

1. Relay **rewrites** the body (re-serialization, injection, auth decoration) → B processes content A produced → double compression with B's config.
2. Relay **strips** unknown headers (`x-bili-hop`) → B can't see the chain at all → same double processing, silently.

## What this does

**S3 — `x-bili-sig` (new header).** Every processing bili stamps the sha256 of the exact body bytes it emits, at every dispatch site that controls the body: initial forward, compat-role retry hop, fake-completion hint retry. The receiving bili verifies it over the raw inbound buffer **before any processing**. Mismatch ⇒ a middlebox rewrote the request ⇒ verbatim passthrough + prominent `[chain]` warn (B must not process content it did not produce).

**S2 — body-artifact fallback.** When both headers are gone, the detector looks for ACP artifacts in the body itself: the full render-tag shape (verified against the kernel renderer's real output — JSON-escaped quotes, K-format token values like `2.1K`, `tool:<name>` types, `mNNNNN` ref + close) or the distinctive tool-name pair `acp_status`+`search_context`. Hit ⇒ same passthrough + `[chain]` warn naming the signal.

**Decision table at the downstream instance:**

| inbound | outcome |
|---|---|
| hop ✓ + sig verified | passthrough (debug log) |
| hop ✓ + sig mismatch | passthrough + rewrite-detect warn |
| hop ✓ + sig absent (legacy old A) | passthrough — today's behavior, unchanged |
| hop ✗ + sig ✓ | passthrough + stripped-header warn |
| hop ✗ + sig ✗ + artifacts ✓ | passthrough + artifact warn |
| none of the above | normal processing |

Invariant: a bili never runs its pipeline on a request whose bytes it did not produce. Plain digest by design — the threat is an indifferent relay, not an adversary (forgery resistance would be a drop-in ed25519 upgrade; no key management needed today).

## Files

- `src/server/chain-sig.ts` (new): stamp/verify/detect module + header constant.
- `src/server.ts`: inbound detection block after the hop-marker check; pipeline gate + route-passthrough log now key off `chainDetected`; digest stamped at all three emission sites; re-export.
- `tests/bili-chain-sig.test.ts` (new): 8 tests — unit round-trip (stamp/verify/detect incl. negative controls), clean A→B chain asserting `sig === sha256(received bytes)` end-to-end, body-rewriting relay (mismatch warn + byte-identical passthrough), header-stripping relay (artifact warn), strip+rewrite relay, legacy marker-only chain, plain-client control (no false positives; B's own emission is marked+signed and self-verifies).
- `README.md` / `README.zh-CN.md`: new "Chaining bili instances" section under Running the proxy.

## Pre-flight (on this head, base = current master 9818c33)

- `npm run typecheck` ✅
- `npm test` ✅ 2170 pass / 0 fail / 2 skipped (ACP_TEST_E2E-gated suite, skipped by design)
- `npm run build` ✅

Note: this PR changes what processed emissions carry (one new header) and adds receiving-side passthrough behavior — wire-shape territory, so it stays human-merge per §7.4.


ranxianglei:

[bot] 🏷 **Review complete — verified, no issues found. Ready for human merge.**

This is a completed-work PR (`ework-agent-pr` lane), so I did not redo anything — I pulled `2026-09-21_bili-chain-sig` @ `feadad9`, audited the code against the spec, and re-ran every gate.

**Base & diff hygiene**
- `feadad9` sits directly on current master `9818c33` (merge-base confirmed) — no stale base, no rebase needed.
- Single commit, exactly the five declared files, +702/−6: no lockfile churn, no generated artifacts, no whitespace noise. README zh/en in sync, section placed where the reader will find it (Running the proxy → Chaining bili instances).

**Code audit (against the decision table)**
- Detection block (`src/server.ts:1005-1025`) implements all six rows exactly; `chainDetected` is a strict superset of the old `hopMarker` gate, so legacy marker-only chains behave byte-for-byte as today.
- All three stamp sites verified in context: initial forward (`src/server.ts:3808`) stamps over the exact `wireBody` after the last mutation (compat-role rewrite) and `init` sends precisely those bytes; the compat-role retry re-stamps over its own `fixed.body`, and for unprocessed requests `retryHeaders === headers === init.headers`, so that hunk is a zero-behavior-change refactor there; the fake-completion retry stamps unconditionally, which is correct because both call sites are provably processed-only (`prepared?.pluginMode` block and the explicit `prepared !== null` guard).
- Multi-hop propagation holds: `buildForwardTarget` copies inbound headers except hop-by-hop, and neither `x-bili-hop` nor `x-bili-sig` is stripped — plus passthrough deliberately does *not* re-stamp, so A→B→C still attributes the bytes to A. That's the subtle part and it's right.
- `verifyChainSig` uses `timingSafeEqual` over fixed-length digests; malformed header values classify as mismatch (conservative). Tag regex requires the full open→ref→close shape, so prose containing partial imitations can't trip it.

**Evidence (re-run on this head, not taken on faith)**
- `npm run typecheck` ✅ · `npm test` ✅ **2170 pass / 0 fail / 2 skipped** (the E2E-gated pair, by design) — matches your pre-flight numbers exactly · `npm run build` ✅
- GitHub checks on `feadad9`: all green, including the real-codex zero-token E2E and the ubuntu/windows × Node 22/24 matrix.
- The full `ACP_TEST_E2E` 4-phase run needs a live loopback upstream, which this sandbox doesn't have — CI's zero-token codex run plus the new mock-e2e relay tests cover the pipeline here.
- Spot-checked the load-bearing test assertions: the clean-chain test asserts `sig === sha256(received bytes)` at the LLM (the actual invariant, not a proxy metric), and the rewriting-relay test asserts the LLM receives the relay's bytes verbatim while the stale sig propagates provably wrong.

**One non-blocking finding (reporting per policy, no separate issue — minor):** the S2 tool-pair seed matches `"acp_status"` + `"search_context"` as quoted substrings anywhere in the body, including message *content*. A plain client whose conversation literally discusses bili/acp-kernel source (both names pasted into the chat) will trip artifact-passthrough, and since those strings then live in the replayed history, compression silently stops for the rest of that session. It's the conservative direction (degrade, never corrupt), the population is small, and the tag-shape alternative is strict enough — acceptable as a documented limitation, but worth knowing if you ever run a plain client *on* this repo's source.

Per §7.4 this stays human-merge regardless: new header on processed emissions + receiving-side passthrough is wire-shape territory affecting both compression modes' neighbors.

---

**中文摘要:** 修复了 bili→bili 链路中第三方中转站改写请求体或剥离 `x-bili-hop` 头导致的静默双重压缩——新增 `x-bili-sig` 体摘要校验(三处发射点全部盖章、透传不重盖,多跳归因正确)加 ACP 产物兜底检测;我在 PR head 上复核了三处盖章点的上下文正确性并全量重跑 typecheck/测试(2170 过)/构建及 CI 状态,全部通过,diff 干净无越界改动,可以合并(按 §7.4 属 wire-shape,需人工合并)。

